PR #27, opened 2019-01-09
Setup dotnet SDK 3.0 preview for Travis

jtattermusch:


jtattermusch:
Got this output, so things are looking good. It might be that some prerequisite libraries might be missing, we will find out once we actually try to build a solution.
```
$ ./build_and_test.sh
+dotnet --version
3.0.100-preview-009812
```

PR #48, opened 2019-01-29
Migrating source changes to grpc/grpc-dotnet

JunTaoLuo:
I'm breaking up this PR into two chunks, src and test. This is the first part. Let me know if it's still too big and you'd like to see samples/benchmarks split out cc @jtattermusch 

@JamesNK I also made some minor infrastructure changes including using a `dependencies.props` to manage our package versions. This is the old-school style we had back in 2.0 before all the tools for `Reference` were added. I don't think the extra complexity with the tools are required for this project and manually updating `dependencies.props` should suffice.


src/Grpc.AspNetCore.Server/Internal/ObjectMethodExecutor/AwaitableInfo.cs
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
JunTaoLuo: cc @Eilon @jtattermusch What license header should these files use? I'm assuming we need to change these to the Grpc headers?

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: csharp
 sudo: required
 dist: xenial
-mono: latest
+mono: none
```
JunTaoLuo: Removing mono dependency to reduce CI build time.

src/Grpc.AspNetCore.Server/Internal/ObjectMethodExecutor/AwaitableInfo.cs
Eilon: This should be whatever it should be from the GRPC perspective. Nothing .NET Foundation about this.

JunTaoLuo: I'll update these license headers

JunTaoLuo:
FYI when we merge, let's keep the commit history and do not squash. Which also makes adding tests back a little easier since I'll only need to revert 7661b99

build/dependencies.props
```diff
@@ -0,0 +1,13 @@
...
+    <GoogleProtobufPackageVersion>3.6.1</GoogleProtobufPackageVersion>
+    <GrpcCorePackageVersion>1.17.0-pre3</GrpcCorePackageVersion>
+    <GrpcToolsPackageVersion>1.17.0-pre3</GrpcToolsPackageVersion>
+    <MicrosoftAspNetCoreTestHostPackageVersion>3.0.0-preview-18579-0056</MicrosoftAspNetCoreTestHostPackageVersion>
+    <MicrosoftNETTestSdkPackageVersion>16.0.0-preview-20181205-02</MicrosoftNETTestSdkPackageVersion>
+    <MoqPackageVersion>4.10.0</MoqPackageVersion>
+    <NunitPackageVersion>3.11.0</NunitPackageVersion>
+    <Nunit3TestAdapterPackageVersion>3.12.0</Nunit3TestAdapterPackageVersion>
+  </PropertyGroup>
+</Project>
```
JamesNK: 👍

doc/packages.md
```diff
@@ -0,0 +1,314 @@
...
+
+## Design considerations
+
+### Grpc.AspNetCore
+
+We envision shipping both a managed client and managed server. Since gRPC services often act as both client and server, we want to make the consumption of both components as frictionless as possible. In addition to the server and client packages, the metapackage will also take dependencies on tooling support (code gen from Grpc.Tools and ServiceReference support from Microsoft.Extensions.ApiDescription.Design) and the default message protocol (Google.Protobuf).
+
+We want to distinguish these from the native clients and server so the package names will be prefixed with Grpc.AspNetCore/Grpc.NetCore. We don't expect most users to take a direct dependency on these packages so we will use a three-part name.
+
+We are committed to delivering the managed server experience Microsoft.AspNetCore.Server functionalities in ASP.NET Core 3.0 timeframe. We will strive to also deliver the mananged client experience in 3.0.
```
jtattermusch: Perhaps the plans and timeline should be described in a separate .md file
(fine to move there later).

src/Grpc.AspNetCore.Server/Internal/StreamUtils.cs
```diff
@@ -0,0 +1,140 @@
...
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    // utilities for parsing gRPC messages from request/response streams
+    // NOTE: implementations are not efficient and are very GC heavy
+    internal class StreamUtils
```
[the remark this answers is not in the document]
JunTaoLuo: Correct, this is temporary until we have the pipeline APIs in place. I wouldn't look at these APIs in particular too closely.

doc/packages.md
JunTaoLuo: I'm not sure there would be enough content for a complete timeline doc so I just included this one line here. cc @shirhatti if you disagree.

sample/Clients/Chatter/Chatter.csproj
```diff
@@ -0,0 +1,26 @@
...
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Requirements (Note that most of these changes will be made to the msbuild targets in Grpc.Tools) -->
+    <!-- 1. Grpc.Tools msbuild files: the Google.Protobuf.Tools.targets file needs to be updated to remove Condition=" '$(DesignTimeBuild)' != 'true' "  -->
+    <!-- 2. Generator metadata: this is the target that's run during the design time build to generate the *.cs files -->
+    <!-- 3. None ItemGroup: Remove the *.proto files from None since the generator metadata for those files are ignored -->
+    <!-- 4. Content ItemGroup: Include the *.proto files in the Content ItemGroup. This isn't required if the file is included in the project directory. I'm not sure about what's going on with this requirement. -->
+    <ProtoBuf Include="..\..\Proto\chat.proto" GrpcServices="Client" Generator="MSBuild:Compile" />
```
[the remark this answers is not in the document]
JunTaoLuo: I was basing my casing on https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Tools/build/_protobuf/Google.Protobuf.Tools.props#L21. Should that be changed? Or should the doc be changed?

benchmarks/Grpc.AspNetCore.Performance/Grpc.AspNetCore.Performance.csproj
```diff
@@ -0,0 +1,25 @@
...
+
+  <ItemGroup>
+    <ProtoBuf Include=".\Proto\*.proto" GrpcServices="Server" Generator="MSBuild:Compile" />
+
+    <None Remove="@(ProtoBuf)" />
+    <Content Include="@(ProtoBuf)" LinkBase="" />
+
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+
+    <PackageReference Include="BenchmarkDotNet" Version="$(BenchmarkDotNetPackageVersion)" />
```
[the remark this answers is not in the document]
JamesNK: It's single threaded only. Its focus is micro-benchmarks.

src/Grpc.AspNetCore.Server/Internal/Constants.cs
```diff
@@ -0,0 +1,29 @@
...
+#endregion
+
+using Grpc.Core;
+using Microsoft.Extensions.Primitives;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal static class Constants
+    {
+        public const string GrpcStatusHeader = "grpc-status";
```
jtattermusch: nit: this is always sent in a trailer, so GrpcStatusTrailer is probably a better name.

src/Grpc.AspNetCore.Server/Internal/Constants.cs
```diff
@@ -0,0 +1,29 @@
...
+        public static readonly StringValues GrpcStatusOk = ((int)StatusCode.OK).ToString();
```
jtattermusch: this one got me confused - I first though it evaluates to the string value "OK" (which would be against https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md), but then realized it's `"0"`. 
Perhaps making it obvious that these are numbers in string form would be good.
Fine to leave as is, but let's come up with something more readable once we start supporting non-ok statuses as well.

src/Grpc.AspNetCore.Server/Internal/Constants.cs
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+#endregion
+
+using Grpc.Core;
+using Microsoft.Extensions.Primitives;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal static class Constants
```
jtattermusch: nit: these are all wire-protocol related constants, so perhaps we could use a more specific name for the class in future.   fine to leave as is.

src/Grpc.AspNetCore.Server/Internal/ObjectMethodExecutor/AwaitableInfo.cs
```diff
@@ -0,0 +1,142 @@
+#region Copyright notice and license
```
jtattermusch: qq: the original file has a .NET foundation license header:
https://github.com/aspnet/AspNetCore/blob/1aa50faa290ecda304507981cd01ed92651d5e34/src/Shared/ObjectMethodExecutor/AwaitableInfo.cs#L1

I am not expert on licensing but my impression was that we need to keep the original header (and add a gRPC license header on top of that?)?
https://github.com/aspnet/AspNetCore/blob/1aa50faa290ecda304507981cd01ed92651d5e34/src/Shared/ObjectMethodExecutor/AwaitableInfo.cs#L1

sample/Clients/Chatter/Chatter.csproj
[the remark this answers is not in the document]
jtattermusch: My understanding was that `<Protobuf>` is more correct. CC @kkm000

sample/Clients/Chatter/Chatter.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
jtattermusch: naming nit: should we change `sample` directory to `example` or `examples`? (we tend to use that more for grpc code). E.g. https://github.com/grpc/grpc/tree/master/examples  or https://github.com/grpc/grpc-java/tree/master/examples

benchmarks/Grpc.AspNetCore.Performance/Grpc.AspNetCore.Performance.csproj
```diff
@@ -0,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
jtattermusch: naming nit:  if the plan for this project is to be mostly about BenchmarkDotNet microbenchmarks, should it be called Grpc.AspNetCore.Microbenchmarks? (it's a more descriptive name and more aligned with existing terminology in grpc codebase).

src/Grpc.AspNetCore.Server/IGrpcServiceActivator.cs
```diff
@@ -0,0 +1,26 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+namespace Grpc.AspNetCore.Server
+{
+    public interface IGrpcServiceActivator<TGrpcService> where TGrpcService : class
```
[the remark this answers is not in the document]
Eilon: I would think the answer is yes, all public APIs must have doc comments. But as far as when they should be put in, that's another question. It might make sense at some point to have an API review, and once the API surface is roughly settled, we can do a doc pass.

src/Grpc.AspNetCore.Server/Internal/ObjectMethodExecutor/AwaitableInfo.cs
Eilon: @jtattermusch in this particular case it's OK because the original author of the code is giving the code to this project (the author is an MSFT person). So, in a sense this isn't taking code from another OSS project, it's the original author contributing the same code twice.

jtattermusch: sounds good, thanks for explaining.

sample/Clients/Chatter/Chatter.csproj
Eilon: BTW `samples` (plural) is a "well known folder" in GitHub. I don't think it does much special stuff - I only recall that it excludes the code in there from statistics such as "what language is this repo written in."

src/Grpc.AspNetCore.Server/Internal/Constants.cs
JamesNK: GrpcServerConstants works for me.

src/Grpc.AspNetCore.Server/Internal/Constants.cs
JamesNK: Lets change it to `public const string GrpcStatusOk = "0";`

benchmarks/Grpc.AspNetCore.Performance/Grpc.AspNetCore.Performance.csproj
JamesNK: It will get renamed to that with a later PR. It's not quite ready https://github.com/aspnet/grpc-dotnet/pull/2/files

src/Grpc.AspNetCore.Server/Internal/GrpcMarkerService.cs
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal class GrpcMarkerService
```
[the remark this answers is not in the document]
JamesNK: > This type is registered in DI by AddGrpc. Its presence when resolved from DI acts as a test of whether GRPC related services are registered.

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+
+                // We need to call Foo.BindService from the declaring type.
+                var declaringType = baseType?.DeclaringType;
+
+                // The method we want to call is public static void BindService(ServiceBinderBase serviceBinder)
+                var bindService = declaringType?.GetMethod("BindService", new[] { typeof(ServiceBinderBase), baseType });
+
+                if (bindService == null)
+                {
+                    throw new InvalidOperationException("Cannot locate BindService(ServiceBinderBase, ServiceBase) method on generated gRPC service type.");
```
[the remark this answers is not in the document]
JamesNK: Lets change it to:

> Cannot locate BindService(ServiceBinderBase, ServiceBase) method for the current service type. The type must be an implementation of a gRPC service.

There is a try/catch that wraps this exception and includes an error message with the name of the service.

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            ValidateServicesRegistered(builder.ServiceProvider);
+
+            var service = typeof(TService);
+
+            try
+            {
+                // Implementation type FooImpl derives from Foo.FooBase (with implicit base type of Object).
```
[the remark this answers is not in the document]
JamesNK: > Note that this is protobuf-specific

Grpc.AspNetCore.Server is only using Grpc.Core APIs. There is no Google.Protobuf dependency.

Is the issue that other generators will have a different way of structuring types? IMO you should encourage a standard.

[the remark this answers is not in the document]
JamesNK: If a standard isn't something you can do, we could add an overload of this method with a Type argument. The type would where the `BindService` for the implementation can be found.

src/Grpc.AspNetCore.Server/Internal/Constants.cs
JunTaoLuo: We lose the reference to StatusCode.OK but I guess we can do that in a comment.

sample/Clients/Chatter/Chatter.csproj
JunTaoLuo: I'm just going to go with the grpc convention since that's where the code is hosted.

benchmarks/Grpc.AspNetCore.Performance/Grpc.AspNetCore.Performance.csproj
jtattermusch: Fine to leave for later.

examples/Clients/Chatter/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+using Grpc.Core;
+using Chat;
+
+namespace Sample.Clients
+{
+    class Program
+    {
+        static async Task Main(string[] args)
+        {
+            var name = args[0];
```
kkm000: nit: I think an error message is a less confusing option than a core dump if the sample is called w/o args.

src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+        where TResponse : class
+        where TService : class
+    {
+        public ClientStreamingServerCallHandler(Method<TRequest, TResponse> method) : base(method)
+        {
+        }
+
+        public override async Task HandleCallAsync(HttpContext httpContext)
+        {
+            httpContext.Response.ContentType = "application/grpc";
```
kkm000: Is anyone on registering this with IANA? Not in [the registry](https://www.iana.org/assignments/media-types/media-types.xml) as of today. Ref. RFC6838.

src/Grpc.AspNetCore.Server/Internal/DefaultGrpcServiceActivator.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Diagnostics;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal class DefaultGrpcServiceActivator<TGrpcService> : IGrpcServiceActivator<TGrpcService> where TGrpcService : class
+    {
+        private static readonly Lazy<ObjectFactory> _objectFactory = new Lazy<ObjectFactory>(() => ActivatorUtilities.CreateFactory(typeof(TGrpcService), Type.EmptyTypes));
+        private readonly IServiceProvider _serviceProvider;
+        private bool? _created;
```
kkm000: Dunno if this code is nearly complete, but the three-state boolean is confusing. To me, a single variable `_createPreviouslyCalled` or such looks much more readable. As is, only its property of having the value is used, not the value itself.

src/Grpc.AspNetCore.Server/Internal/ObjectMethodExecutor/AwaitableInfo.cs
```diff
@@ -0,0 +1,142 @@
...
+#endregion
+
+using System;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal struct AwaitableInfo
```
kkm000: minor suggestion: `readonly struct`. It's immutable as the dodo anyway, and big enough to potentially benefit from by const ref.

src/Grpc.AspNetCore.Server/Internal/StreamUtils.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            if (dest.Length < MessageDelimiterSize)
+            {
+                throw new ArgumentException("Buffer too small to encode message length.");
+            }
+
+            ulong unsignedValue = (ulong)messageLength;
+            for (int i = MessageDelimiterSize - 1; i >= 0; i--)
+            {
+                // msg length stored in big endian
```
kkm000: Does this assume the host is little-endian? `System.Net.IPAddress.HostToNetworkOrder` maybe then?

src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
JunTaoLuo: cc @jtattermusch ?

src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+            httpContext.Response.Headers.Append("grpc-encoding", "identity");
```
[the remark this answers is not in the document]
JunTaoLuo: According to spec, both are valid. However, I think we should follow what the c-core server does in grpc/grpc. I see the [example document](https://github.com/grpc/grpc/blob/2b99c000f1d549f2e184968f5b0b34c40d7e3b7b/doc/PROTOCOL-HTTP2.md#example) shows that these are separate headers. I'll let @jtattermusch make the final call though.

src/Grpc.AspNetCore.Server/Internal/DefaultGrpcServiceActivator.cs
JunTaoLuo: The code is taken from https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/server/Core/src/Internal/DefaultHubActivator.cs. The value is used in the dispose section for cleanup purposes. This brings up a good point on how we should handle TService that implement IDisposable for example.

src/Grpc.AspNetCore.Server/Internal/ObjectMethodExecutor/AwaitableInfo.cs
JunTaoLuo: Could be interesting, might want to benchmark the difference @JamesNK

src/Grpc.AspNetCore.Server/Internal/StreamUtils.cs
JamesNK: As far as I know this is written to work regardless of the endianess of the host.

Writing the message length will be updated to use `BinaryPrimitives.WriteUInt32BigEndian` and `BinaryPrimitives.ReadUInt32BigEndian` in a later PR.

JunTaoLuo: https://github.com/grpc/grpc/blob/2b99c000f1d549f2e184968f5b0b34c40d7e3b7b/doc/PROTOCOL-HTTP2.md#requests It's defined as big endian in the document.

src/Grpc.AspNetCore.Server/Internal/ObjectMethodExecutor/AwaitableInfo.cs
JunTaoLuo: Filed https://github.com/aspnet/AspNetCore/issues/7167

JamesNK:
It would be great to get this merged soon. We still have tests to merge afterwards. A backlog of PRs is building up at https://github.com/aspnet/grpc-dotnet/pulls.

src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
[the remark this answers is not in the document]
jtattermusch: I agree with going with whatever c-core server does and setting content type to "application/grpc".

doc/packages.md
kingpitt1331: @shirhatti

kingpitt1331: @shirhatti

PR #55, opened 2019-02-01
Migrate test changes to grpc/grpc-dotnet

JunTaoLuo:
This is the follow up to https://github.com/grpc/grpc-dotnet/pull/48. All current PRs should be unblocked after this.

### DIFF
--- a/test/FunctionalTests/ClientStreamingMethodTests.cs
+++ b/test/FunctionalTests/ClientStreamingMethodTests.cs
@@ -1,0 +1,81 @@
+#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Count;
+using Grpc.AspNetCore.FunctionalTests.Infrastructure;
+using NUnit.Framework;
+
+namespace Grpc.AspNetCore.FunctionalTests
+{
+    [TestFixture]
+    public class ClientStreamingMethodTests : FunctionalTestBase
+    {
+        [Test]
+        public async Task AccumulateCount_MultipleMessagesThenClose_SuccessResponse()
+        {
+            // Arrange
+            var ms = new MemoryStream();
+            await MessageHelpers.WriteMessageAsync(ms, new CounterRequest
+            {
+                Count = 1
+            }).DefaultTimeout();
+
+            var requestStream = new AwaitableMemoryStream();
+
+            var httpRequest = new HttpRequestMessage(HttpMethod.Post, "Count.Counter/AccumulateCount");
+            httpRequest.Content = new StreamContent(requestStream);
+
+            // Act
+            var responseTask = Fixture.Client.SendAsync(httpRequest, HttpCompletionOption.ResponseHeadersRead);
+
+            // Assert
+            Assert.IsFalse(responseTask.IsCompleted, "Server should wait for client to finish streaming");
+
+            Fixture.Signal.Reset();
+            requestStream.SendData(ms.ToArray());
+            await Fixture.Signal.Task.DefaultTimeout();
+
+            Fixture.Signal.Reset();
+            requestStream.SendData(ms.ToArray());
+            await Fixture.Signal.Task.DefaultTimeout();
+
+            Fixture.Signal.Reset();
+            requestStream.SendData(Array.Empty<byte>());
+            await Fixture.Signal.Task.DefaultTimeout();
+
+            var response = await responseTask.DefaultTimeout();
+
+            Assert.AreEqual(HttpStatusCode.OK, response.StatusCode);
+            Assert.AreEqual("identity", response.Headers.GetValues("grpc-encoding").Single());
+            Assert.AreEqual("application/grpc", response.Content.Headers.ContentType.MediaType);
+
+            var responseStream = await response.Content.ReadAsStreamAsync().DefaultTimeout();
+
+            var replyTask = MessageHelpers.AssertReadMessageAsync<CounterReply>(responseStream);
+            Assert.IsTrue(replyTask.IsCompleted);
+            var reply = await replyTask.DefaultTimeout();
+            Assert.AreEqual(2, reply.Count);
+        }
+    }
+}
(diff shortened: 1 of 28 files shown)


testassets/FunctionalTestsWebsite/Chatter.cs
```diff
@@ -0,0 +1,71 @@
+﻿#region Copyright notice and license
```
jtattermusch: what's the distinction between `test`  and `testassets` directories?

testassets/Proto/chat.proto
```diff
@@ -0,0 +1,29 @@
+// Copyright 2019 The gRPC Authors
```
jtattermusch: doest it make sense to duplicate the .proto files between `examples/` and `testassets/` ?

test/FunctionalTests/ClientStreamingMethodTests.cs
```diff
@@ -0,0 +1,81 @@
...
+using Count;
+using Grpc.AspNetCore.FunctionalTests.Infrastructure;
+using NUnit.Framework;
+
+namespace Grpc.AspNetCore.FunctionalTests
+{
+    [TestFixture]
+    public class ClientStreamingMethodTests : FunctionalTestBase
+    {
+        [Test]
```
jtattermusch: I was under the impression that `[TestCase]` is the recommended attribute in Nunit3?

testassets/FunctionalTestsWebsite/Infrastructure/Signaler.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        public void Reset()
+        {
+            _tcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+        }
+
+        public void Set()
+        {
+            _tcs.TrySetResult(null);
```
jtattermusch: check that TrySetResult returns true?

jtattermusch:
I tried to build locally on debian 9 and got this error. Not sure what's up, just posting here for now:

```
  google/protobuf/Empty.proto: File not found.
  count.proto: Import "google/protobuf/Empty.proto" was not found or had errors.
  count.proto:23:23: "google.protobuf.Empty" is not defined.
/usr/local/google/home/jtattermusch/.nuget/packages/grpc.tools/1.18.0/build/_protobuf/Google.Protobuf.Tools.targets(263,5): error MSB6006: "/usr/local/google/home/jtattermusch/.nuget/packages/grpc.tools/1.18.0/tools/linux_x64/protoc" exited with code 1. [/usr/local/google/home/jtattermusch/github/grpc-dotnet/testassets/FunctionalTestsWebsite/FunctionalTestsWebsite.csproj]
```

testassets/Proto/count.proto
```diff
@@ -0,0 +1,36 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+import "google/protobuf/Empty.proto";
```
jtattermusch: should be `empty.proto`.   `Empty` will fail on unix.

JunTaoLuo: Oops, I only ran the tests on Win.

testassets/FunctionalTestsWebsite/Chatter.cs
JunTaoLuo: tests contains the actual test code. testassets contain the system under test, which could be test services, test clients, or other artifacts.

testassets/Proto/chat.proto
JunTaoLuo: I had the same thought. But as @JamesNK mentioned, we expect the examples and testassets to diverge, see https://github.com/aspnet/grpc-dotnet/pull/6#discussion_r249913689.

test/FunctionalTests/ClientStreamingMethodTests.cs
JunTaoLuo: Hmm seems like `TestCase` is more often used to pass inline arguments to the test method. Although passing no arguments to `TestCase` works, it is functionally the same according to the wiki: https://github.com/nunit/docs/wiki/Attributes. I see both being used in [grpc/grpc](https://github.com/nunit/docs/wiki/Attributes) but we can change the style if you wish.

jtattermusch: Leaving up to you.

JunTaoLuo: Personally I prefer `Test` since it's more concise 😄

test/FunctionalTests/Infrastructure/GrpcTestFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+                services.AddSingleton(Signal);
+            };
+
+            var builder = new WebHostBuilder()
+                .ConfigureServices(configureServices)
+                .UseStartup<TStartup>();
+
+            _server = new TestServer(builder);
+
+            Client = _server.CreateClient();
```
jtattermusch: qq: is this creating an actual server and an actual client?   Btw, I thought there is no currently HttpClient that'd support HTTP/2.

test/FunctionalTests/Infrastructure/AwaitableMemoryStream.cs
```diff
@@ -0,0 +1,110 @@
...
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Grpc.AspNetCore.FunctionalTests.Infrastructure
+{
+    public class AwaitableMemoryStream : Stream
```
jtattermusch: Can we make it more clear what this class is for?
It took me a while to understand that the usage is to fill the stream with calling SendData() which then awakes the ReadAsync() operation - so basically this is a pipe. It sounds we could reuse some existing code for this purpose.

Btw, the logic in this class is nontrivial, so if left as is, shouldn't we have unit tests for this utility class?

test/FunctionalTests/Infrastructure/MessageHelpers.cs
```diff
@@ -0,0 +1,168 @@
...
+                return default;
+            }
+
+            var message = new T();
+            message.MergeFrom(messageData);
+
+            return message;
+        }
+
+        public static async Task<byte[]> ReadMessageAsync(Stream stream)
```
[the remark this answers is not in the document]
JunTaoLuo: I think most of these points are addressed by https://github.com/aspnet/grpc-dotnet/pull/16. Could add more tests though.

test/FunctionalTests/Infrastructure/AwaitableMemoryStream.cs
JunTaoLuo: The implementation will be changed to `SyncPointMemoryStream` in  aspnet#16.

test/FunctionalTests/Infrastructure/MessageHelpers.cs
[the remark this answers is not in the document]
JamesNK: I've added some sanity check tests that the raw bytes are correct - https://github.com/aspnet/grpc-dotnet/pull/16/commits/416d64889934e969fade2777b45665f6aa68e356

As part of the pipelines PR I'll clean up the test helpers to use the core app logic instead of copy pasted code

Issue #45, opened 2019-01-24
Create nightly feed for upstream packages


JunTaoLuo:
To share updates between grpc/grpc and grpc/grpc-dotnet, we should create a nightly NuGet feed where upstream changes are pushed as part of the build. This is one proposal, but we're open to other ideas

Issue #21, opened 2019-01-07
BindService should not require a concrete instance of the GRPC service implementation 


shirhatti:
Currently we need an instance of the service implementation to call BindService and we are creating a dummy instance to bind the service right now: https://github.com/JunTaoLuo/GrpcSandbox/blob/105f2bc65ad1b6b642a62bec31789315936c78b2/Server/Dotnet/GrpcRouteBuilderExtensions.cs#L34-L36. Since we are exploring the use of Scoped services, we may not have an instance until the request is in flight.

JamesNK:
I looked into this and I see two potential solutions:

1. We use the `Descriptor` property that is generated on the service definition. That has all the information that we need: method name, method type, input message, output message.

2. The tools are updated to generate a `BindService` overload that doesn't take an implementation. `ServiceBinderBase` would have a new virtual method that takes a `Method<TRequest, TResponse>` and it is called in the new `BindService` overload.

The nice thing about (2) is it provides a strongly typed hook for handling the method, i.e. you have `TRequest` and `TResponse`, and those generic types can then be used to call the marshaller.

On the other hand, getting into a generic context with (1) isn't that difficult. Combine some Type.MakeGenericType reflection with the input and output messages types and you're done. I lean towards doing (1) for now.

Thoughts @JunTaoLuo?

JunTaoLuo:
I didn't see option (1) before but that sounds intriguing. I'll prototype it out and see what that looks like.

JunTaoLuo:
Implemented option (1)

JunTaoLuo:
@jtattermusch Now that we've changed the implementation, I'd like to propose that we add an option to the grpc code generation to skip generating the `BindService` methods. This would remove `ServiceBinderBase` and `ServerServiceDefinition` from Grpc.Common/Api and keep those types in Grpc.Core.

JunTaoLuo:
@JamesNK I realized that option (1) actually adds a dependency on the Google.Protobuf package since we need access to the `ServiceDescriptor`. Grpc.Core goes around this issue by relying on the generated code for service binding, which isolates the GRPC implementation from the underlying message protocol (ProtoBuf). Though we can influence how to design Grpc.AspNetCore, I'm wondering if the isolation is something we need to keep in mind.

JamesNK:
Good catch. I don't know. As long as we don't expose protobuf types externally we have the flexibility to remove it easily. We can discuss next meeting about whether this is an issue or not.

[...]

PR #57, opened 2019-02-01
Add scripts to setup and update local feeds

JunTaoLuo:
Addresses https://github.com/grpc/grpc-dotnet/issues/45. Add scripts to help ingest nightly gRPC packages. This won't conflict with the test migration PR.

I will update the scripts to download Grpc.Core.Api in a followup PR since I ran into some issues.


JunTaoLuo:
Looking into failure

JamesNK:
One thing that worries me about this is the NuGet package ID of the `Grpc.Core`/`Grpc.Tools` packages doesn't change. If the build server doesn't make completely clean build runs then would the package content will be cached on the build server?

Also what would happen if `Grpc.Core`/`Grpc.Tools` introduce breaking changes (probably unlikely but might happen during dev). Do we immediately opt-in to the latest version?

Issue I created about adding the build number to the Grpc.* packages: https://github.com/grpc/grpc/issues/17886

Grpc.AspNetCore.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28307.136
+# Visual Studio Version 16
```
jtattermusch: is switch to VS2019 necessary for this PR?

build/upgrade-deps.ps1
```diff
@@ -0,0 +1,64 @@
+﻿#!/usr/bin/env powershell
```
jtattermusch: rather than depending on XML parsing and a relatively complex logic being implemented twice, would it be 
easier to just hardcode  the latest Grpc version "1.19.0-dev"  and the prefix for downloading the nuget packages?
(it's always the same prefix for all nugets from the same build).
Besides thing being just simpler that way, I think it's better to rely on a fixed version of the nupkgs (and update the download prefix manually if needed) rather than automatically downloading the latest version.

Grpc.AspNetCore.sln
JunTaoLuo: No but our infrastructure is moving towards requiring VS2019 soon so I updated the version of VS on my machine. I think @JamesNK has already made the switch and I noticed our PRs were ping ponging between the two. There isn't much harm in this though as I was able to open these solutions in VS2017 as well.

build/upgrade-deps.ps1
JunTaoLuo: I think I'm going to change the implementation a little since I wanted to add scripts to remove the manual step. I think I'll split this into two scripts one to update the version and one to download the packages. Most builds will run the download only and will not need the xml dependency. We will run the update version tool manually when we need to ingest new packages.

JunTaoLuo: @JamesNK This would also allow us to control when to ingest new changes.

JunTaoLuo:
@JamesNK That's a valid concern, I know the Travis agents are clean for every build so our checkin tests should be fine. However, I'm not sure if there's additional CI to be set up on kokoro cc @jtattermusch? If so, do we know if the agents will be cleaned (specifically the nuget cache) before every build?

JunTaoLuo:
@jtattermusch @JamesNK can we make the travis tests `Required` for PR merges? It's fast and reliable.

JamesNK:
Fine with me

JunTaoLuo:
Updated to address #21 as a proof of concept. It's such a small change I'll tack it onto this PR instead of opening a separate one that's dependent on this one.

jtattermusch:
> @JamesNK That's a valid concern, I know the Travis agents are clean for every build so our checkin tests should be fine. However, I'm not sure if there's additional CI to be set up on kokoro cc @jtattermusch? If so, do we know if the agents will be cleaned (specifically the nuget cache) before every build?

On Kokoro there will be a freshly re-imaged VM each time a build runs (same behavior like travis).

JunTaoLuo:
Yay we can merge!

PR #58, opened 2019-02-01
Use Lazy for object method executor

JamesNK:
Minor refactor in how the object method executor is resolved.


src/Grpc.AspNetCore.Server/Internal/ServerCallHandlerBase.cs
```diff
@@ -22,42 +22,38 @@
 using Grpc.Core;
 using Grpc.AspNetCore.Server.Internal;
 using Microsoft.AspNetCore.Http;
+using System.Threading;
```
JunTaoLuo: nit:sort

PR #59, opened 2019-02-01
Update to .NET Core 3.0 preview 2

JamesNK:


jtattermusch:
> Rebased but github still says merging is blocked 🤷‍♂️

I also noticed something was odd with the project settings (I was getting the merging blocked message too), so I switched to project merging settings to "create a merge commit". Not sure if that's the change that was needed, but right now, the PR seems mergeable normally.

Merging.

PR #60, opened 2019-02-01
Update ObjectMethodExecutor structs to be readonly

JamesNK:


JamesNK:
Why does this PR require the branch to be up to date before it can be merged? This isn't going to conflict with any other changes.

PR #62, opened 2019-02-04
Use delegates to call through to user code directly

davidfowl:
Today we're using the ObjectMethodExecutor to invoke GRPC methods, this changes it to use open delegates instead since it seems as though signatures for specific types of calls are fixed. Some of the benefits
- We can avoid expression tree compilation.
- We can avoid the object[] allocation for invoking the method as well as potential boxing (though these are all value types)
- We get a bit more type safety

I wasn't sure if some of the method signatures could change for the specific types of calls. If so we can special case a bit more or fallback to the ObjectMethodExecutor (though I removed it in this PR).


src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
```diff
@@ -28,8 +29,15 @@ internal class ClientStreamingServerCallHandler<TRequest, TResponse, TService> :
...
     {
+        private delegate Task<TResponse> ClientStreamingServerCall(TService service, IAsyncStreamReader<TRequest> stream, ServerCallContext serverCallContext);
+
+        private readonly ClientStreamingServerCall _invoker;
+
         public ClientStreamingServerCallHandler(Method<TRequest, TResponse> method) : base(method)
         {
+            var handlerMethod = typeof(TService).GetMethod(Method.Name);
+
+            _invoker = (ClientStreamingServerCall)Delegate.CreateDelegate(typeof(ClientStreamingServerCall), handlerMethod);
```
JamesNK: This signature should match, but if it doesn't do you think it is worth having a try/catch around creating the delegate with a friendly message.

src/Grpc.AspNetCore.Server/Internal/DuplexStreamingServerCallHandler.cs
```diff
@@ -42,14 +49,10 @@ public override async Task HandleCallAsync(HttpContext httpContext)
...
 
-            await ObjectMethodExecutor.ExecuteAsync(
-                service,
-                new object[]
-                {
-                    new HttpContextStreamReader<TRequest>(httpContext, Method.RequestMarshaller.Deserializer),
-                    new HttpContextStreamWriter<TResponse>(httpContext, Method.ResponseMarshaller.Serializer),
-                    null
-                });
+            await _invoker(service,
```
JamesNK: nit: indenting is weird

src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
davidfowl: Yes, we should do this. We should also try to understand if it's reasonable to not match. Did they write the method by hand or is it codegen.

JamesNK: It is codegen.

@jtattermusch is the method signature constant between serializers?

jtattermusch: @JamesNK  not sure I understand the question

JamesNK: Will codegens always output the same signature for the 4 different types of calls. For example, will a unary call always be:

`Task<TResponse> SayHello(TRequest request, ServerCallContext context)`

jtattermusch: The server side method handlers are code generated and their signatures match these delegates:
https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core.Api/ServerMethods.cs

https://github.com/grpc/grpc/blob/ed0a5655833e9d91a361b647db68f34e76e69774/src/csharp/Grpc.Examples/MathGrpc.cs#L284

jtattermusch: @JamesNK  yes, the signature on server side will always be the same and there's no plans to change that.

davidfowl: Ah there are already delegates I can use!

davidfowl: Ah wait, I can't use these because I need to pass in the instance (this) as the first argument. That said, maybe we don't need error handling.

davidfowl:
Anything else required here @jtattermusch ?

davidfowl:
I think it should be assigned in the constructor but a common wouldn't hurt here.

davidfowl:
We could also consider (in another PR) adding these delegates to Grpc.Core.

davidfowl:
@jtattermusch is this good to merge?

JamesNK:
Hrm, I'm still blocked from merging PRs

image

jtattermusch:
I'm looking at the settings and I don't see anything suspicious. The problem is it doesn't say why merging is blocked.

jtattermusch:
I changed the setting: https://help.github.com/articles/enabling-branch-restrictions/   to allow grpc-dotnet-team to push to the master branch. Did that help? If it does, that's odd, because I thought that setting is only for explicit "git push" commands and it doesn't affect merging pull requests.

PR #63, opened 2019-02-04
Add build and test instructions to README.md

jtattermusch:


README.md
```diff
@@ -27,7 +27,19 @@ Instructions coming soon!
 
 ## To develop gRPC for ASP.NET Core
 
-TODO: add development instructions
+Install [.NET Core SDK 3 preview 2](https://dotnet.microsoft.com/download/dotnet-core/3.0)
+or [VS 2019 preview](https://visualstudio.microsoft.com/vs/preview/).
```
JamesNK: I'm pretty sure VS2019 doesn't ship with .NET Core 3 preview 2.

jtattermusch: Ok, so there's currently no way to build with VS? (I haven't tried yet).

JamesNK: VS2017 or VS2019 should work at the moment once you have downloaded preview 2.

I'm not sure about VS Code, I don't use it myself. @davidfowl are you using VS Code?

davidfowl: I'm using VS 2019

davidfowl:
I just noticed the Experimental Only section in the README. Can we change that to say something to the effect that's it's pre-release instead of experimental?

jtattermusch:
@davidfowl  well right now it's really just experimental, you can't really do anything useful with it - so I think that characterization is accurate. But feel free to propose alternative wording - I just wanted to warn potential visitors that this doesn't work just yet.

davidfowl:
I disagree. Experimental means that we don't know if this is going to work not, not that it's going to ship. We use that experimental term for actual experiments that we're trying to figure out if it'll work out. I'd rather say something like this is under active development and there's no release for it as yet. That's much more the reality of the situation rather than calling it an experiment.

jtattermusch:
@davidfowl  makes sense, can you propose the right wording then? Alternatively, feel free to create a PR to update yourself.

JunTaoLuo:
I tried updating the comments over at https://github.com/jtattermusch/grpc-dotnet/pull/1. I don't think I can push to your branch so I opened a PR. @JamesNK @davidfowl @jtattermusch PTAL.

jtattermusch:
Integrated John's suggestions into this PR.

PR #64, opened 2019-02-04
Add pipelines

JamesNK:
++ @davidfowl


src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// Read messages from the pipe reader.
+        /// </summary>
+        /// <param name="input">The request pipe reader.</param>
+        /// <param name="supportMultipleMessages">
+        /// A value that indicates whether multiple messages can be read from the reader.
+        /// When false, a complete message must be read from the reader and the reader is completed with no further data.
+        /// When true, a complete message must be read from the reader, and the reader can be open with additional data.
+        /// </param>
+        /// <returns>Complete message data.</returns>
+        public static async ValueTask<byte[]> ReadMessageAsync(this PipeReader input, bool supportMultipleMessages)
```
davidfowl: I think we should have 2 methods. One that reads a single message and one that reads mulitple (basically `IAsyncEnumerable<byte[]>`)

JamesNK: Do you want the read multiple to use C# 8 async + yield and return an `IAsyncEnumerable<byte[]>`?

I think it is something to try out. Can it wait though? I'd like to avoid requiring VS2019 just yet. Plus it would be good to compare perf with the current implementation.

davidfowl: I hate how complicated this method is 😢

JamesNK: I have split it: `ReadMessageAsync` and `ReadMessageStreamAsync`

test/Grpc.AspNetCore.Server.Tests/PipeExtensionsTests.cs
```diff
@@ -0,0 +1,355 @@
...
+
+            var pipeReader = new StreamPipeReader(ms);
+
+            // Act
+            var ex = Assert.ThrowsAsync<InvalidOperationException>(
+                () => pipeReader.ReadMessageStreamAsync().AsTask());
+
+            // Assert
+            Assert.AreEqual("Incomplete message.", ex.Message);
+        }
```
JunTaoLuo: Add a test for this condition? https://github.com/grpc/grpc-dotnet/pull/64/files#diff-feaffa8467755e7b6df8bf1465d5bed0R152

test/Grpc.AspNetCore.Server.Tests/PipeExtensionsTests.cs
```diff
@@ -0,0 +1,355 @@
...
+
+        [Test]
+        public async Task WriteMessageAsync_NoFlush_WriteNoData()
+        {
+            // Arrange
+            var ms = new MemoryStream();
+            var pipeWriter = new StreamPipeWriter(ms);
+
+            // Act
+            await pipeWriter.WriteMessageAsync(Array.Empty<byte>());
```
JunTaoLuo: To actually test not flushing writes no data, the argument should be a non-empty array.

JamesNK: An empty message would still write the header.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -0,0 +1,262 @@
...
+{
+    internal static class PipeExtensions
+    {
+        private const int MessageDelimiterSize = 4; // how many bytes it takes to encode "Message-Length"
+        private const int HeaderSize = MessageDelimiterSize + 1; // message length + compression flag
+
+        public static Task WriteMessageAsync(this PipeWriter pipeWriter, byte[] messageData, bool flush = false)
+        {
+            WriteHeader(pipeWriter, messageData.Length);
+            pipeWriter.Write(messageData);
```
davidfowl: We've evolved this logic a bit. See https://github.com/aspnet/AspNetCore/pull/7227/files#diff-d2000ab873598422d9bacae788b02a8bR56

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -0,0 +1,262 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Read a message in a stream from the pipe reader. Additional message data is left in the reader.
+        /// </summary>
+        /// <param name="input">The request pipe reader.</param>
+        /// <returns>Complete message data or null if the stream is complete.</returns>
+        public static async ValueTask<byte[]> ReadMessageStreamAsync(this PipeReader input)
```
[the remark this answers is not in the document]
JamesNK: I'd like to see some numbers with IAsyncEnumerable vs the current method before we change it. How would it make it faster? As far as I know it would add overhead of allocating an enumerable. We have an IAsyncStreamReader for that purpose already.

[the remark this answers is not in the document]
davidfowl: - We would write a single loop which means a single async state machine for the entire enumeration, not one per call.
- We would implement the IAsyncStreamReader by getting the enumerator and by manually calling MoveNext() and Current on it.

JamesNK:
🆙 📅

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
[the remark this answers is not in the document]
jtattermusch: Regardless of what ReadMessageStreamAsync() returns,  shouldn't we deduplicate the logic for reading (that's what my original comment was about)?
My suggestion is to make the IAsyncEnumerable switch in a separate PR (where we can collect performance data if needed).

Issue #36, opened 2019-01-18
Accessing HttpContext in a gRPC service call


JamesNK:
gRPC C# offers the `ServerCallContext` with information about the current call. Currently there is no way access the underlying `HttpContext` from service methods.

- [ ] Decide if HttpContext should be accessible
  * Service base class?
  * Service interface?
  * Include it as part of `ServerCallContext`. Get with extension method?
- [ ] Implementation

jtattermusch:
Accessing HttpContext should be more of an exception than a rule. No other gRPC implementation exposes "raw" details about the transport layer (among other reasons, gRPC should be transport layer agnostic). I'd say we should only expose HttpContext if it turns out to be necessary and it should be considered more of a workaround rather than a supported feature.

If we do need to expose HttpContext,  extension method on `ServerCallContext` seems like a good approach.

JamesNK:
It will be useful for accessing values set on the HttpContext by middleware. I will definitely be the exception, mostly for power users.

JamesNK:
Blocked on ServerCallContext

Issue #35, opened 2019-01-18
Pass ServerCallContext to service methods


JamesNK:
Currently `ServerCallContext` passed to service methods is always null.

Tasks?
- [x] Decouple the current `ServerCallContext` from native
- [x] Create an implementation for .NET
- [x] Construct and pass to service methods

Related: https://github.com/grpc/grpc/pull/17733

Issue #47, opened 2019-01-28
Error handling


JamesNK:
Ensure we set the correct headers on error - https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#errors

Consider an error handling API for users to catch and inspect application errors. e.g. MVC's error filtering https://docs.microsoft.com/en-us/aspnet/core/fundamentals/error-handling?view=aspnetcore-2.2#exception-filters

PR #69, opened 2019-02-06
Adding implementation of ServerCallContext and pass to services

JunTaoLuo:
Addresses https://github.com/grpc/grpc-dotnet/issues/35 and https://github.com/grpc/grpc-dotnet/issues/36.

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
@@ -51,18 +51,21 @@ namespace Grpc.AspNetCore.Server.Internal
             // Activate the implementation type via DI.
             var activator = httpContext.RequestServices.GetRequiredService<IGrpcServiceActivator<TService>>();
             var service = activator.Create();
+            var serverCallContext = new HttpContextServerCallContext(httpContext);
 
             var response = await _invoker(
                 service,
                 new HttpContextStreamReader<TRequest>(httpContext, Method.RequestMarshaller.Deserializer),
-                null);
+                serverCallContext);
 
             // TODO(JunTaoLuo, JamesNK): make sure the response is not null
-            var responsePayload = Method.ResponseMarshaller.Serializer(response);
+            var responseBodyPipe = httpContext.Response.BodyPipe;
+            await responseBodyPipe.WriteMessageAsync(response, Method.ResponseMarshaller.Serializer, serverCallContext.WriteOptions);
 
-            await httpContext.Response.BodyPipe.WriteMessageAsync(responsePayload, flush: true);
+            httpContext.Response.ConsolidateTrailers(serverCallContext);
 
-            httpContext.Response.AppendTrailer(GrpcProtocolConstants.StatusTrailer, GrpcProtocolConstants.StatusOk);
+            // Flush any buffered content
+            await httpContext.Response.BodyPipe.FlushAsync();
         }
     }
 }
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
@@ -1,0 +1,134 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System;
+using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+using Grpc.Core;
+using Microsoft.AspNetCore.Http;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal class HttpContextServerCallContext : ServerCallContext
+    {
+        private string _peer;
+        private Metadata _requestHeaders;
+        private Metadata _responseTrailers;
+
+        internal HttpContextServerCallContext(HttpContext httpContext)
+        {
+            HttpContext = httpContext;
+        }
+
+        internal HttpContext HttpContext { get; }
+
+        internal bool HasResponseTrailers => _responseTrailers != null;
+
+        protected override string MethodCore => HttpContext.Request.Path.Value;
+
+        protected override string HostCore => HttpContext.Request.Host.Value;
+
+        protected override string PeerCore
+        {
+            get
+            {
+                if (_peer == null)
+                {
+                    var connection = HttpContext.Connection;
+                    if (connection.RemoteIpAddress != null)
+                    {
+                        _peer = (connection.RemoteIpAddress.AddressFamily == AddressFamily.InterNetwork ? "ipv4:" : "ipv6:") + connection.RemoteIpAddress + ":" + connection.RemotePort;
+                    }
+                }
+
+                return _peer;
+            }
+        }
+
+        // TODO(JunTaoLuo, JamesNK): implement this
+        protected override DateTime DeadlineCore => throw new NotImplementedException();
+
+        protected override Metadata RequestHeadersCore
+        {
+            get
+            {
+                if (_requestHeaders == null)
+                {
+                    _requestHeaders = new Metadata();
+
+                    foreach (var header in HttpContext.Request.Headers)
+                    {
+                        if (!header.Key.StartsWith(':'))
+                        {
+                            _requestHeaders.Add(header.Key, header.Value);
+                        }
+                    }
+                }
+
+                return _requestHeaders;
+            }
+        }
+
+        // TODO(JunTaoLuo, JamesNK): implement this
+        protected override CancellationToken CancellationTokenCore => throw new NotImplementedException();
+
+        protected override Metadata ResponseTrailersCore
+        {
+            get
+            {
+                if (_responseTrailers == null)
+                {
+                    _responseTrailers = new Metadata();
+                }
+
+                return _responseTrailers;
+            }
+        }
+
+        protected override Status StatusCore { get; set; }
+
+        protected override WriteOptions WriteOptionsCore { get; set; }
+
+        // TODO(JunTaoLuo, JamesNK): implement this
+        protected override AuthContext AuthContextCore => throw new NotImplementedException();
+
+        // TODO(JunTaoLuo, JamesNK): implement this
+        protected override ContextPropagationToken CreatePropagationTokenCore(ContextPropagationOptions options)
+        {
+            throw new NotImplementedException();
+        }
+
+        protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
+        {
+            if (responseHeaders != null)
+            {
+                foreach (var entry in responseHeaders)
+                {
+                    if (!entry.IsBinary)
+                    {
+                        // TODO(juntaoluo): what about binary headers?
+                        HttpContext.Response.Headers[entry.Key] = entry.Value;
+                    }
+                }
+            }
+
+            return HttpContext.Response.Body.FlushAsync();
+        }
+    }
+}
--- a/src/Grpc.AspNetCore.Server/Internal/HttpResponseExtensions.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpResponseExtensions.cs
@@ -1,0 +1,44 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using Microsoft.AspNetCore.Http;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal static class HttpResponseExtensions
+    {
+        public static void ConsolidateTrailers(this HttpResponse httpResponse, HttpContextServerCallContext context)
+        {
+            if (context.HasResponseTrailers)
+            {
+                foreach (var trailer in context.ResponseTrailers)
+                {
+                    // TODO(juntaoluo): What about binary trailers
+                    if (!trailer.IsBinary)
+                    {
+                        httpResponse.AppendTrailer(trailer.Key, trailer.Value);
+                    }
+                }
+            }
+
+            // Append status trailers, these overwrite any existing status trailers set via ServerCallContext.ResponseTrailers
+            httpResponse.AppendTrailer(GrpcProtocolConstants.StatusTrailer, context.Status.StatusCode.ToTrailerString());
+            httpResponse.AppendTrailer(GrpcProtocolConstants.MessageTrailer, context.Status.Detail);
+        }
+    }
+}
--- a/src/Grpc.AspNetCore.Server/Internal/ServerStreamingServerCallHandler.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/ServerStreamingServerCallHandler.cs
@@ -55,12 +55,19 @@ namespace Grpc.AspNetCore.Server.Internal
             // Activate the implementation type via DI.
             var activator = httpContext.RequestServices.GetRequiredService<IGrpcServiceActivator<TService>>();
             var service = activator.Create();
+            var serverCallContext = new HttpContextServerCallContext(httpContext);
+            var streamWriter = new HttpContextStreamWriter<TResponse>(serverCallContext, Method.ResponseMarshaller.Serializer);
 
-            await _invoker(service, request, new HttpContextStreamWriter<TResponse>(httpContext, Method.ResponseMarshaller.Serializer), null);
+            await _invoker(
+                service,
+                request,
+                streamWriter,
+                serverCallContext);
 
+            httpContext.Response.ConsolidateTrailers(serverCallContext);
+
+            // Flush any buffered content
             await httpContext.Response.BodyPipe.FlushAsync();
-
-            httpContext.Response.AppendTrailer(GrpcProtocolConstants.StatusTrailer, GrpcProtocolConstants.StatusOk);
         }
     }
 }
--- a/src/Grpc.AspNetCore.Server/Internal/UnaryServerCallHandler.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/UnaryServerCallHandler.cs
@@ -57,15 +57,21 @@ namespace Grpc.AspNetCore.Server.Internal
             // Activate the implementation type via DI.
             var activator = httpContext.RequestServices.GetRequiredService<IGrpcServiceActivator<TService>>();
             var service = activator.Create();
+            var serverCallContext = new HttpContextServerCallContext(httpContext);
 
-            var response = await _invoker(service, request, null);
+            var response = await _invoker(
+                service,
+                request,
+                serverCallContext);
 
             // TODO(JunTaoLuo, JamesNK): make sure the response is not null
-            var responsePayload = Method.ResponseMarshaller.Serializer(response);
+            var responseBodyPipe = httpContext.Response.BodyPipe;
+            await responseBodyPipe.WriteMessageAsync(response, Method.ResponseMarshaller.Serializer, serverCallContext.WriteOptions);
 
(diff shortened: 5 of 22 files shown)


src/Grpc.AspNetCore.Server/Internal/DefaultServerCallContext.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        // TODO(JunTaoLuo, JamesNK): implement this
+        protected override CancellationToken CancellationTokenCore => throw new NotImplementedException();
+
+        // TODO(JunTaoLuo, JamesNK): implement this
+        protected override Metadata ResponseTrailersCore => throw new NotImplementedException();
+
+        protected override Status StatusCore { get; set; }
+
+        protected override WriteOptions WriteOptionsCore { get; set; }
```
JunTaoLuo: I need to hook this up with the HttpContextStreamWriter

JunTaoLuo:
Would be good to see how many more interop tests pass after this change goes in.

src/Grpc.AspNetCore.Server/Internal/DefaultServerCallContext.cs
```diff
@@ -0,0 +1,96 @@
...
+using System;
+using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+using Grpc.Core;
+using Microsoft.AspNetCore.Http;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal class DefaultServerCallContext : ServerCallContext
```
JamesNK: Lets call it HttpContextServerCallContext

src/Grpc.AspNetCore.Server/Internal/DefaultServerCallContext.cs
JamesNK: I've already done it

src/Grpc.AspNetCore.Server/Internal/DefaultServerCallContext.cs
```diff
@@ -0,0 +1,96 @@
...
+        private Lazy<string> _peer;
+
+        internal DefaultServerCallContext(HttpContext httpContext)
+        {
+            HttpContext = httpContext;
+
+            var connection = HttpContext.Connection;
+            if (connection.RemoteIpAddress == null)
+            {
+                _peer = new Lazy<string>();
```
JamesNK: Lazy is overkill. Just make it a string that is default to null and add an if check.

src/Grpc.AspNetCore.Server/ServerCallContextExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace Grpc.Core
+{
+    /// <summary>
+    /// Extension methods for ServerCallContext.
+    /// </summary>
+    public static class ServerCallContextExtensions
+    {
+        public static HttpContext GetHttpContext(this ServerCallContext serverCallContext)
```
JamesNK: XML docs

src/Grpc.AspNetCore.Server/Internal/UnaryServerCallHandler.cs
```diff
@@ -57,13 +57,18 @@ public override async Task HandleCallAsync(HttpContext httpContext)
             // Activate the implementation type via DI.
             var activator = httpContext.RequestServices.GetRequiredService<IGrpcServiceActivator<TService>>();
             var service = activator.Create();
+            var streamWriter = new HttpContextStreamWriter<TResponse>(httpContext, Method.ResponseMarshaller.Serializer);
```
JamesNK: Why did you add this?

src/Grpc.AspNetCore.Server/Internal/DefaultServerCallContext.cs
JunTaoLuo: What do you mean?

src/Grpc.AspNetCore.Server/Internal/UnaryServerCallHandler.cs
JunTaoLuo: Because of WriteOptions. The service can modify the WriteFlags during the call and it can affect how TResponse should be written. For example, the compress flag can be set. Also this standardizes this portion of the logic for all handlers.

src/Grpc.AspNetCore.Server/Internal/DefaultServerCallContext.cs
JamesNK: Yeah, I wrote my own temp version of the context. What you've done is fine. I'll rebase the tests I added on top.

src/Grpc.AspNetCore.Server/Internal/UnaryServerCallHandler.cs
JamesNK: Ok. Then we should have a write helper method that takes the serializer, write options and message. The StreamWriter and this place can use it. That avoids us allocating the stream writer here.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -0,0 +1,124 @@
...
+                _peer = new Lazy<string>($"{(connection.RemoteIpAddress.AddressFamily == AddressFamily.InterNetwork ? "ipv4" : "ipv6")}:{connection.RemoteIpAddress}:{connection.RemotePort}");
+            }
+
+            _requestHeaders = new Lazy<Metadata>(() =>
+            {
+                var metadata = new Metadata();
+
+                foreach (var header in httpContext.Request.Headers)
+                {
+                    if (!header.Key.StartsWith(':'))
```
JunTaoLuo: @jtattermusch looks like there's some discrepancies between how headers are handled and I just want to call them out:
1. Metadata converts header names to lower-case whereas ASP.NET Core does not
2. ASP.NET Core exposes pseudo-headers (:path for example) in the header collection whereas Metadata doesn't allow this.
3. We do not store binary header values. We parse them as UTF-8.

Are we okay with these behavioural differences?

src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
```diff
@@ -51,16 +51,17 @@ public override async Task HandleCallAsync(HttpContext httpContext)
             // Activate the implementation type via DI.
             var activator = httpContext.RequestServices.GetRequiredService<IGrpcServiceActivator<TService>>();
             var service = activator.Create();
+            var streamWriter = new HttpContextStreamWriter<TResponse>(httpContext, Method.ResponseMarshaller.Serializer);
```
JamesNK: Remove here as well

src/Grpc.AspNetCore.Server/Internal/DefaultServerCallContext.cs
JunTaoLuo: What does your implementation look like? I found my way of doing it not quite ideal.

src/Grpc.AspNetCore.Server/Internal/IWriteOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+
+#endregion
+
+using Grpc.Core;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal interface IWriteOptions
+    {
+        WriteOptions WriteOptions { get; set; }
```
JamesNK: Get rid of this. Just pass the ServerCallContext to the stream writer.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -0,0 +1,124 @@
...
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal class HttpContextServerCallContext : ServerCallContext
+    {
+        private readonly IWriteOptions _writeOptions;
+        private readonly Lazy<Metadata> _requestHeaders;
+        private Lazy<string> _peer;
+
+        internal HttpContextServerCallContext(HttpContext httpContext, IWriteOptions writeOptions)
```
JamesNK: Pass the context to the stream writer instead of the stream writer to the context.

src/Grpc.AspNetCore.Server/Internal/DefaultServerCallContext.cs
JunTaoLuo: I was hoping to avoid synthesizing this string unless it's called.

src/Grpc.AspNetCore.Server/Internal/DefaultServerCallContext.cs
JamesNK: https://github.com/JamesNK/grpc-dotnet/commit/ee097a4637812ceb695f38545bac21adee9948a0

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            HttpContext = httpContext;
+
+            var connection = HttpContext.Connection;
+            if (connection.RemoteIpAddress != null)
+            {
+                PeerCore = $"{(connection.RemoteIpAddress.AddressFamily == AddressFamily.InterNetwork ? "ipv4" : "ipv6")}:{connection.RemoteIpAddress}:{connection.RemotePort}";
+            }
+
+            _requestHeaders = new Lazy<Metadata>(() =>
```
JamesNK: Don't use a lazy. Just move this logic to the getter.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -0,0 +1,135 @@
...
+        // TODO(JunTaoLuo, JamesNK): implement this
+        protected override DateTime DeadlineCore => throw new NotImplementedException();
+
+        protected override Metadata RequestHeadersCore
+        {
+            get
+            {
+                if (_requestHeaders == null)
+                {
+                    _requestHeaders = new Metadata();
```
JunTaoLuo: In general I'd be concerned about duplicating collections. I need to verify that the header dictionary is immutable at this point.

JunTaoLuo: Similar issue with ResponseTrailers.

JamesNK: From what I can tell the HttpContext related types aren't designed to be thread safe, e.g. https://github.com/aspnet/AspNetCore/blob/7bd5297cfa672c797e2fcbfffa6dc048699abd32/src/Http/Http/src/Features/ResponseCookiesFeature.cs#L55-L67

src/Grpc.AspNetCore.Server/Internal/HttpResponseExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+                    // TODO(juntaoluo): What about binary trailers
+                    if (!trailer.IsBinary)
+                    {
+                        httpResponse.AppendTrailer(trailer.Key, trailer.Value);
+                    }
+                }
+            }
+
+            // Append status trailers, these overwrite any existing status trailers set via ServerCallContext.ResponseTrailers
+            httpResponse.AppendTrailer(GrpcProtocolConstants.StatusTrailer, context.Status.StatusCode.ToString("D"));
```
JamesNK: I believe ToString("D") will allocation. We could have a method with a switch statement that returns the appropriate string directly. default can fall back to ToString("D").

src/Grpc.AspNetCore.Server/Internal/HttpResponseExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+                {
+                    // TODO(juntaoluo): What about binary trailers
+                    if (!trailer.IsBinary)
+                    {
+                        httpResponse.AppendTrailer(trailer.Key, trailer.Value);
+                    }
+                }
+            }
+
+            // Append status trailers, these overwrite any existing status trailers set via ServerCallContext.ResponseTrailers
```
JunTaoLuo: @jtattermusch I tested the behaviour with the c-core server and this is what I observed. Does that sound correct to you?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: It's not really thread safety, I was more concerned that operations on one collection will not be propagated to the other. I think I can update the implementation to make sure they are in sync though.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -0,0 +1,134 @@
...
+                {
+                    if (!entry.IsBinary)
+                    {
+                        // TODO(juntaoluo): what about binary headers?
+                        HttpContext.Response.Headers[entry.Key] = entry.Value;
+                    }
+                }
+            }
+
+            return HttpContext.Response.Body.FlushAsync();
```
jtattermusch: Is this the right way to force sending the response headers? (we need to support a scenario where server responds with headers, but response messages are sent much later).

Btw, what happens if I've already sent some payload (= a message) in the body?  In that case this method should give an error.

Also, in Grpc.Core, you can only write the response headers once. Second attempt gives an error. Does this behave the same?

src/Grpc.AspNetCore.Server/Internal/HttpResponseExtensions.cs
jtattermusch: I think you're right. setting the status and details "manually" via the Metadata class is actually something that no one really does.  Btw, in c-core there's actually a different API through which we pass the status of the call and I'm pretty sure that overwrites the metadata entries with the same header name.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: Modifying the request headers on the server seems kind of odd. The amount of effort to keep them in sync doesn't seem like it would be worth that scenario.

src/Grpc.AspNetCore.Server/Internal/ServerStreamingServerCallHandler.cs
```diff
@@ -55,12 +55,19 @@ public override async Task HandleCallAsync(HttpContext httpContext)
             // Activate the implementation type via DI.
             var activator = httpContext.RequestServices.GetRequiredService<IGrpcServiceActivator<TService>>();
             var service = activator.Create();
+            var serverCallContext = new HttpContextServerCallContext(httpContext);
+            var streamWriter = new HttpContextStreamWriter<TResponse>(serverCallContext, Method.ResponseMarshaller.Serializer);
 
-            await _invoker(service, request, new HttpContextStreamWriter<TResponse>(httpContext, Method.ResponseMarshaller.Serializer), null);
+            await _invoker(
```
jtattermusch: one thing that's missing here is to handle RpcException thrown by user's code.
That's an alternative way to tell the handling stack to set the status and trailers that are set in the exception.

See https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core/Internal/ServerCallHandler.cs#L296.

Same of other call types.

I think it's fine to address in a followup PR.

src/Grpc.AspNetCore.Server/ServerCallContextExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// Extension methods for ServerCallContext.
+    /// </summary>
+    public static class ServerCallContextExtensions
+    {
+        /// <summary>
+        /// Retrieve the <see cref="HttpContext"/> from a <see cref="ServerCallContext"/> if possible.
+        /// </summary>
+        /// <param name="serverCallContext">The <see cref="ServerCallContext"/> to extract from.</param>
+        /// <returns>The extracted <see cref="HttpContext"/>. If it cannot be extracted, null will be returned.</returns>
+        public static HttpContext GetHttpContext(this ServerCallContext serverCallContext)
```
jtattermusch: nit: add a note that this is for power users (and recommend read-only access)?   trying to change stuff in the httpcontext while the grpc layer potentially accesses the same fields can easily break things.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
jtattermusch: > @jtattermusch looks like there's some discrepancies between how headers are handled and I just want to call them out:
> 
> 1. Metadata converts header names to lower-case whereas ASP.NET Core does not

not sure where that would be a problem. a grpc app would  access the metadata through ServerCallContext where all the entries seems to have the right casing. On the wire everything will be lowercase because it's required by HTTP/2 spec.

> 2. ASP.NET Core exposes pseudo-headers (:path for example) in the header collection whereas Metadata doesn't allow this.

Looks like that's already addressed by hiding the pseudo headers?

> 3. We do not store binary header values. We parse them as UTF-8.

When writing, the "-bin" metadata entries should just be base64 encoded.

when reading the request headers, we can recognize the "-bin" suffix and base64 decode the value when creating Metadata.Entry.

> 
> Are we okay with these behavioural differences?

test/Grpc.AspNetCore.Server.Tests/DefaultServerCallContextTests.cs
```diff
@@ -0,0 +1,166 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
+using Microsoft.Extensions.Primitives;
+using NUnit.Framework;
+
+namespace Grpc.AspNetCore.Server.Tests
+{
+    [TestFixture]
+    public class DefaultServerCallContextTests
```
jtattermusch: nit: rename class to `HttpContextServerCallContextTests`?

src/Grpc.AspNetCore.Server/Internal/ServerStreamingServerCallHandler.cs
JunTaoLuo: Yes I was thinking that work belongs in https://github.com/grpc/grpc-dotnet/issues/47.

src/Grpc.AspNetCore.Server/ServerCallContextExtensions.cs
JamesNK: I'd add a remarks section and say that this method provides access to the underlying HttpContext and that HttpContext is only available when using gRPC hosted on ASP.NET Core.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: I think this is safe albeit not quite efficient.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: Right, I'm trying to replicate the gRPC behaviours here as much as possible but I just want to highlight them so we are aware of these differences. The only remaining point is the binary headers. We don't support these. As in, if we receive binary headers that cannot be parsed as ASCII or UTF-8. Kestrel will throw and we reject the request and sends a 400 response. I'll file a follow up issue to discuss this further as this requires a significant amount of work to support on our side.

JunTaoLuo: Filed https://github.com/grpc/grpc-dotnet/issues/71

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
+        {
+            if (responseHeaders != null)
+            {
+                foreach (var entry in responseHeaders)
+                {
+                    if (!entry.IsBinary)
+                    {
+                        // TODO(juntaoluo): what about binary headers?
```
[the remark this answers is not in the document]
JunTaoLuo: Let's discuss this in the issue.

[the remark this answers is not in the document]
JunTaoLuo: I see, I'll address this in a separate PR.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: There's a few discrepancies in behaviour but this is the closest we can get with the current APIs available.

Yes, this allows the user to send the headers without sending any body since the Body may be empty. The current implementation is that any write to the body is flushed which automatically sends headers if they have not been sent yet. There's no scenario where the body can be sent without the headers. Currently flush is going to no-op if the headers has already been sent. How important is it to throw an error instead? I need to think about how this might be achieved on our side as it's not immediately obvious how this can be achieved.

JunTaoLuo:
Thanks for the comments! I'm going to merge this as is to unblock other changes but I'll address a few more feedback items in followup PRs.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: Actually I lied. there's a super easy way to do this. `HttpContext.Response.HasStarted` I'll still addresss this in a followup PR though.

Issue #46, opened 2019-01-28
Support gRPC timeout


JamesNK:
gRPC supports timeouts

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md - grpc-timeout

https://grpc.io/blog/deadlines

Exposed as `ServerCallContext.Deadline`. Will need to populate from the header.

JunTaoLuo:
I think this will need https://github.com/grpc/grpc-dotnet/issues/35 first right?

JamesNK:
Blocked on ServerCallContext

dhhoang:
#73 allows zero timeout value. The http2 transport docs says that timeout value should be positive: TimeoutValue → {positive integer as ASCII string of at most 8 digits} https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests

JamesNK:
Thanks for the heads up. They also ignore bad values instead of throwing. Updated - https://github.com/grpc/grpc-dotnet/pull/88

PR #73, opened 2019-02-06
 Add deadline

JamesNK:
Addresses #46

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
@@ -53,10 +53,16 @@ namespace Grpc.AspNetCore.Server.Internal
             var service = activator.Create();
             var serverCallContext = new HttpContextServerCallContext(httpContext);
 
-            var response = await _invoker(
-                service,
-                new HttpContextStreamReader<TRequest>(httpContext, Method.RequestMarshaller.Deserializer),
-                serverCallContext);
+            serverCallContext.Initialize();
+
+            TResponse response;
+            using (serverCallContext)
+            {
+                response = await _invoker(
+                    service,
+                    new HttpContextStreamReader<TRequest>(httpContext, Method.RequestMarshaller.Deserializer),
+                    serverCallContext);
+            }
 
             // TODO(JunTaoLuo, JamesNK): make sure the response is not null
             var responseBodyPipe = httpContext.Response.BodyPipe;
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
@@ -17,6 +17,7 @@
 #endregion
 
 using System;
+using System.Globalization;
 using System.Net.Sockets;
 using System.Threading;
 using System.Threading.Tasks;
@@ -25,11 +26,15 @@ using Microsoft.AspNetCore.Http;
 
 namespace Grpc.AspNetCore.Server.Internal
 {
-    internal class HttpContextServerCallContext : ServerCallContext
+    internal sealed class HttpContextServerCallContext : ServerCallContext, IDisposable
     {
+        // Override the current time for unit testing
+        internal ISystemClock Clock = SystemClock.Instance;
         private string _peer;
         private Metadata _requestHeaders;
         private Metadata _responseTrailers;
+        private DateTime _deadline;
+        private CancellationTokenSource _cts;
 
         internal HttpContextServerCallContext(HttpContext httpContext)
         {
@@ -61,8 +66,7 @@ namespace Grpc.AspNetCore.Server.Internal
             }
         }
 
-        // TODO(JunTaoLuo, JamesNK): implement this
-        protected override DateTime DeadlineCore => throw new NotImplementedException();
+        protected override DateTime DeadlineCore => _deadline;
 
         protected override Metadata RequestHeadersCore
         {
@@ -96,8 +100,7 @@ namespace Grpc.AspNetCore.Server.Internal
             }
         }
 
-        // TODO(JunTaoLuo, JamesNK): implement this
-        protected override CancellationToken CancellationTokenCore => throw new NotImplementedException();
+        protected override CancellationToken CancellationTokenCore => _cts.Token;
 
         protected override Metadata ResponseTrailersCore
         {
@@ -149,6 +152,33 @@ namespace Grpc.AspNetCore.Server.Internal
             }
 
             return HttpContext.Response.Body.FlushAsync();
+        }
+		
+        public void Initialize()
+        {
+            var timeout = GrpcProtocolHelpers.GetTimeout(HttpContext);
+
+            if (timeout != TimeSpan.MaxValue)
+            {
+                // CancellationTokenSource does not support greater than int.MaxValue milliseconds
+                if (timeout.TotalMilliseconds > int.MaxValue)
+                {
+                    throw new InvalidOperationException("A timeout greater than 2147483647 milliseconds is not supported.");
+                }
+
+                _deadline = Clock.UtcNow.Add(timeout);
+                _cts = new CancellationTokenSource(timeout);
+            }
+            else
+            {
+                _deadline = DateTime.MaxValue;
+                _cts = new CancellationTokenSource();
+            }
+        }
+
+        public void Dispose()
+        {
+            _cts?.Dispose();
         }
     }
 }
--- a/test/FunctionalTests/DeadlineTests.cs
+++ b/test/FunctionalTests/DeadlineTests.cs
@@ -1,0 +1,89 @@
+#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System.IO;
+using System.IO.Pipelines;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Greet;
+using Grpc.AspNetCore.FunctionalTests.Infrastructure;
+using Grpc.AspNetCore.Server.Internal;
+using NUnit.Framework;
+
+namespace Grpc.AspNetCore.FunctionalTests
+{
+    [TestFixture]
+    public class DeadlineTests : FunctionalTestBase
+    {
+        [TestCase("Greet.Greeter/SayHellosDeadline")]
+        [TestCase("Greet.Greeter/SayHellosDeadlineCancellationToken")]
+        public async Task WriteUntilDeadline_SuccessResponsesStreamed(string requestUri)
+        {
+            // Arrange
+            var requestMessage = new HelloRequest
+            {
+                Name = "World"
+            };
+
+            var requestStream = new MemoryStream();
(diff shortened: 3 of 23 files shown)


src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -61,8 +65,7 @@ protected override string PeerCore
             }
         }
 
-        // TODO(JunTaoLuo, JamesNK): implement this
-        protected override DateTime DeadlineCore => throw new NotImplementedException();
+        protected override DateTime DeadlineCore => _deadline;
```
jtattermusch: Note that the deadline support means two things:
- exposing the deadline to the user code via server call context
- the framework itself should watch the deadline and terminate the http stream if deadline is reached. Also, the  serverCallContext.CancellationToken should be automatically cancelled by the framework when the deadline is reached (another case when the cancellationToken should be triggered is when the client cancels).

testassets/FunctionalTestsWebsite/Greeter.cs
```diff
@@ -37,7 +38,38 @@ public override Task<HelloReply> SayHello(HelloRequest request, ServerCallContex
...
+    {
+        context.WriteResponseHeadersAsync(null);
+
+        return SayHellosCore(request, responseStream);
+    }
+
+    public override async Task SayHellosDeadline(HelloRequest request, IServerStreamWriter<HelloReply> responseStream, ServerCallContext context)
+    {
+        var i = 0;
+        while (DateTime.UtcNow < context.Deadline)
```
jtattermusch: rather than watching the current time, we should probably watch the context.CancellationToken (and framework should cancel it when the deadline is crossed).

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
jtattermusch: I think the right way to terminate a stream on the server side if deadline is reached is to send RST_STREAM (and not to send a status "DeadlineExceeded"), but I will have to confirm that. AFAIK the "DeadlineExceeded" status code is generated on the client side - with gRPC, both client and server are watching the deadline and both parties cancel if it's reached.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -130,5 +133,51 @@ protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
...
+        {
+            const long TicksPerMicrosecond = 10;
+            const long NanosecondsPerTick = 100;
+
+            if (!HttpContext.Request.Headers.TryGetValue(GrpcProtocolConstants.TimeoutHeader, out var values))
+            {
+                return DateTime.MaxValue;
+            }
+
+            if (values.Count == 1)
```
jtattermusch: nit: maybe the logic for parsing the timeout header shouldn't be directly in the ServerCallContext?

AlgorithmsAreCool:
I'm curious if there is any concern about the system clock not being a stable enough time source for timeout logic. 

What if the clock syncs and jumps a minute into the past or future? Is that scenario rare enough to not worry about?

JamesNK:
re: https://github.com/grpc/grpc-dotnet/pull/73#discussion_r254637889

The cancellation token source won't be effected by a changing system clock.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -130,5 +133,51 @@ protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
 
             return HttpContext.Response.Body.FlushAsync();
         }
+
+        private DateTime GetDeadline()
```
JunTaoLuo: Curious where the logic to parse the deadline was specified? Was there a document on this I missed?

test/FunctionalTests/DeadlineTests.cs
```diff
@@ -0,0 +1,88 @@
...
+using Grpc.AspNetCore.Server.Internal;
+using NUnit.Framework;
+
+namespace Grpc.AspNetCore.FunctionalTests
+{
+    [TestFixture]
+    public class DeadlineTests : FunctionalTestBase
+    {
+        [Test]
+        public async Task WriteUntilDeadline_SuccessResponsesStreamed()
```
JunTaoLuo: Add a test for the negative case? where the Deadline is exceeded?

test/FunctionalTests/Infrastructure/PipeReaderFixStream.cs
```diff
@@ -0,0 +1,88 @@
...
+#endregion
+
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Grpc.AspNetCore.FunctionalTests.Infrastructure
+{
+    // Stream to work around StreamPipeReader issue
+    // https://github.com/aspnet/AspNetCore/issues/7329
```
JunTaoLuo: 😢 Add a reminder (TODO or issue) to remove this workaround

test/FunctionalTests/ServerStreamingMethodTests.cs
```diff
@@ -114,12 +117,55 @@ public async Task WriteResponseHeadersAsyncCore_FlushesHeadersToClient()
...
             Assert.False(goodbyeTask.IsCompleted);
             Assert.AreEqual("Goodbye World!", (await goodbyeTask.DefaultTimeout()).Message);
 
-            var finishedTask = MessageHelpers.AssertReadMessageStreamAsync<HelloReply>(responseStream);
+            var finishedTask = MessageHelpers.AssertReadStreamMessageAsync<HelloReply>(pipeReader);
             Assert.IsNull(await finishedTask.DefaultTimeout());
         }
+
+        [Test]
+        public async Task Buffering_SuccessResponsesStreamed()
```
JunTaoLuo: Is this a little redundant? ServerStreaming and DuplexStreaming mechanisms are both using a `IServerStreamWriter`

testassets/FunctionalTestsWebsite/Greeter.cs
JamesNK: I've added a test for that as well.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: I've made the CancellationToken use the deadline. Let me know what cancellation should do to the http stream.

JamesNK:
🆙 📅

test/FunctionalTests/DeadlineTests.cs
```diff
@@ -0,0 +1,88 @@
...
+            Assert.AreEqual("application/grpc", response.Content.Headers.ContentType.MediaType);
+
+            var responseStream = await response.Content.ReadAsStreamAsync().DefaultTimeout();
+            var pipeReader = new StreamPipeReader(new PipeReaderFixStream(responseStream));
+
+            var messageCount = 0;
+
+            var readTask = Task.Run(async() =>
+            {
+                while (true)
```
JunTaoLuo: We should not be relying on this running in a loop. We should explicitly set the clock values for the test and trigger the read/write loops if possible. This may be easier to construct in a DuplexStreaming method.

test/FunctionalTests/ServerStreamingMethodTests.cs
JamesNK: You say that but I found a bug by adding having both tests. One of the call handlers wasn't flushing correctly at the end.

test/FunctionalTests/DeadlineTests.cs
JamesNK: This will run in a loop for 200 milliseconds and then stop. If it doesn't stop then DefaultTimeout will kill it and throw an error. I don't want to be exact about timings in functional tests because it results in flaky tests.

test/FunctionalTests/DeadlineTests.cs
JamesNK: This tests the deadline being exceeded.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

Timeout → "grpc-timeout" TimeoutValue TimeoutUnit
TimeoutValue → {positive integer as ASCII string of at most 8 digits}
TimeoutUnit → Hour / Minute / Second / Millisecond / Microsecond / Nanosecond
Hour → "H"
Minute → "M"
Second → "S"
Millisecond → "m"
Microsecond → "u"
Nanosecond → "n"

src/Grpc.AspNetCore.Server/Internal/ClientStreamingServerCallHandler.cs
```diff
@@ -53,10 +53,16 @@ public override async Task HandleCallAsync(HttpContext httpContext)
...
+            serverCallContext.Initialize();
+
+            TResponse response;
+            using (serverCallContext)
+            {
+                response = await _invoker(
+                    service,
+                    new HttpContextStreamReader<TRequest>(httpContext, Method.RequestMarshaller.Deserializer),
+                    serverCallContext);
+            }
```
JunTaoLuo: This is fine, though I could also imagine allowing cancellation up until the point where we are done serializing the response and are about the write to the body.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -130,5 +133,32 @@ protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
...
+                    throw new InvalidOperationException("A timeout greater than 2147483647 milliseconds is not supported.");
+                }
+
+                _deadline = Clock.UtcNow.Add(timeout);
+                _cts = new CancellationTokenSource(timeout);
+            }
+            else
+            {
+                _deadline = DateTime.MaxValue;
+                _cts = new CancellationTokenSource();
```
JunTaoLuo: in this case, how would the _cts be cancelled?

JamesNK: Client could cancel.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
jtattermusch: the server should send RST_STREAM(CANCEL) if the deadline is exceeded.

The processing on client side should look like this (irrelevant now but will be important for managed client):
When the client receives RST_STREAM, it should double-check its local timer to see if it has expired (it's possible it's expired, but not been processed yet). If it has expired, it should throw away the server "response" (the RST) and generate a DEADLINE_EXCEEDED instead.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolConstants.cs
```diff
@@ -20,6 +20,7 @@ namespace Grpc.AspNetCore.Server.Internal
 {
     internal static class GrpcProtocolConstants
     {
+        internal const string TimeoutHeader = "grpc-timeout";
```
jtattermusch: Something we should probably address at some point:
Not sure if we are excluding these headers from metadata exposed via ServerCallContext, but in general the grpc-specific headers that have special handling in the framework shouldn't be exposed by visible in metadata (Grpc.Core doesn't expose them because the c-core also doesn't expose them).

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: > the server should send RST_STREAM(CANCEL) if the deadline is exceeded.

What about if the client cancels? What does a client cancellation look like?

I've been struggling to find good docs around gRPC cancellation.

jtattermusch: if client cancels (for whatever reason - it can be that the deadline exceeded on the client side or that the user has decided to cancel), it just sends RST_STREAM(cancel) to notify the server. I'm assuming this is the default behavior when HttpClient decides to cancel an inflight request.

PR #74, opened 2019-02-07
Small improvements to PipeExtensions

davidfowl:
- Handle synchronous completion of flushes (the majority case)
- Use AsTask to avoid the state machine completely
- Use in when passing ReadOnlySequence since it's a big struct


davidfowl:
Experience with these API since introducing them in the last year or so. We do this today in several other cases and it’s shown to make a huge difference in hot paths.

Issue #52, opened 2019-02-01
Maximum request and response message size settings


JamesNK:
https://grpc.io/grpc/csharp/api/Grpc.Core.ChannelOptions.html#Grpc_Core_ChannelOptions_MaxReceiveMessageLength

https://grpc.io/grpc/csharp/api/Grpc.Core.ChannelOptions.html#Grpc_Core_ChannelOptions_MaxSendMessageLength

JamesNK:
@jtattermusch Does gRPC have a default maximum message size?

I think we should have global configuration for setting a maximum receive size and maximum send size that applies to all services. We could optionally add configuration on a service by service level.

Receive validation would happen when the message header is read. Too large and an error is thrown.

Send validate would happen before the message is sent. Again, too large and an error is thrown. As far as I can tell the current marshaller API surface doesn't have a way to get the message size so we would have to do this after the message is serialized.

PR #78, opened 2019-02-08
Add options and message size limits

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/52

### DIFF
--- a/src/Grpc.AspNetCore.Server/GrpcServiceExtensions.cs
+++ b/src/Grpc.AspNetCore.Server/GrpcServiceExtensions.cs
@@ -20,6 +20,7 @@ using System;
 using Grpc.AspNetCore.Server;
 using Grpc.AspNetCore.Server.Internal;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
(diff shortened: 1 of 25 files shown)


src/Grpc.AspNetCore.Server/GrpcServiceExtensions.cs
```diff
@@ -29,22 +30,59 @@ namespace Microsoft.Extensions.DependencyInjection
...
+
+        /// <summary>
+        /// Adds gRPC services to the specified <see cref="IServiceCollection" />.
         /// </summary>
         /// <param name="services">The <see cref="IServiceCollection"/> for adding services.</param>
-        /// <returns></returns>
-        // TODO(JunTaoLuo): Options?
-        public static IServiceCollection AddGrpc(this IServiceCollection services)
+        /// <returns>An <see cref="IGrpcServerBuilder"/> that can be used to further configure the gRPC services.</returns>
+        public static IGrpcServerBuilder AddGrpc(this IServiceCollection services)
```
JunTaoLuo: Why do we need a separate Builder type?

JamesNK: https://github.com/grpc/grpc-dotnet/blob/43cd16735b633e6b0d8cdc5837dc48726b87291d/test/Grpc.AspNetCore.Server.Tests/GrpcServicesExtensionsTests.cs#L64-L67

JunTaoLuo: Can this not be done through just a ServiceCollection? It looks like IGrpcServerBuilder is just a wrapper around it anyway. Is there more things you want to add to the interface later?

JamesNK: If it was on a ServiceCollection then you could call AddServiceOptions without AddGrpc, which doesn't make sense. It is there for intelli-sense. MVC and SignalR do the same thing.

JunTaoLuo: Sounds good

src/Grpc.AspNetCore.Server/Internal/ServerCallHandlerBase.cs
```diff
@@ -26,10 +26,12 @@ namespace Grpc.AspNetCore.Server.Internal
     internal abstract class ServerCallHandlerBase<TRequest, TResponse, TService> : IServerCallHandler
     {
         protected Method<TRequest, TResponse> Method { get; }
+        protected GrpcServiceOptions ServiceOptions { get; }
 
-        protected ServerCallHandlerBase(Method<TRequest, TResponse> method)
+        protected ServerCallHandlerBase(Method<TRequest, TResponse> method, GrpcServiceOptions serviceOptions)
         {
-            Method = method ?? throw new ArgumentNullException(nameof(method));
+            Method = method;
```
jtattermusch: nit: can the method be null now?

Issue #71, opened 2019-02-06
Handle binary header and trailer values


JunTaoLuo:
Currently Kestrel will reject headers it cannot parse as ASCII or UTF-8 with a 400 response. We also do not support sending binary trailers. We need to discuss how to support this feature and what changes are required in ASP.NET.

JunTaoLuo:
https://github.com/grpc/grpc-dotnet/pull/69#pullrequestreview-200446456 Actually just noticed that binary headers and trailers are always assumed to be base64 encoded as mentioned in https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests. This makes it easier to handle.

jtattermusch:
Yes this, shouldn't be a problem. There's no such thing as "binary headers" in HTTP/2 AFAIK. gRPC only makes it appear so. One of the popular use cases is let one of the trailers carry a serialized protobuf with e.g. structured error details (example here: https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto#L80).

PR #79, opened 2019-02-08
Handle binary headers and multiple calls to WriteResponseHeadersAsync

JunTaoLuo:
Addresses https://github.com/grpc/grpc-dotnet/issues/71

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
@@ -74,7 +74,18 @@ namespace Grpc.AspNetCore.Server.Internal
 
                     foreach (var header in HttpContext.Request.Headers)
                     {
-                        if (!header.Key.StartsWith(':'))
+                        // ASP.NET Core includes pseudo headers in the set of request headers
+                        // whereas, they are not in gRPC implementations. We will filter them
+                        // out when we construct the list of headers on the context.
+                        if (header.Key.StartsWith(':'))
+                        {
+                            continue;
+                        }
+                        else if (header.Key.EndsWith(Metadata.BinaryHeaderSuffix, StringComparison.OrdinalIgnoreCase))
+                        {
+                            _requestHeaders.Add(header.Key, Convert.FromBase64String(header.Value));
+                        }
+                        else
                         {
                             _requestHeaders.Add(header.Key, header.Value);
                         }
@@ -116,13 +127,22 @@ namespace Grpc.AspNetCore.Server.Internal
 
         protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
         {
+            // Headers can only be written once. Throw on subsequent call to write response header instead of silent no-op.
+            if (HttpContext.Response.HasStarted)
+            {
+                throw new InvalidOperationException("Response headers can only be sent once per call.");
+            }
+
             if (responseHeaders != null)
             {
                 foreach (var entry in responseHeaders)
                 {
-                    if (!entry.IsBinary)
+                    if (entry.IsBinary)
                     {
-                        // TODO(juntaoluo): what about binary headers?
+                        HttpContext.Response.Headers[entry.Key] = Convert.ToBase64String(entry.ValueBytes);
+                    }
+                    else
+                    {
                         HttpContext.Response.Headers[entry.Key] = entry.Value;
                     }
                 }
--- a/test/FunctionalTests/UnaryMethodTests.cs
+++ b/test/FunctionalTests/UnaryMethodTests.cs
@@ -154,7 +154,7 @@ namespace Grpc.AspNetCore.FunctionalTests
             foreach (var b in ms.ToArray())
             {
                 await requestStream.AddDataAndWait(new[] { b }).DefaultTimeout();
-            }            
+            }
             await requestStream.AddDataAndWait(Array.Empty<byte>()).DefaultTimeout();
 
             var response = await responseTask.DefaultTimeout();
@@ -163,6 +163,32 @@ namespace Grpc.AspNetCore.FunctionalTests
 
             var responseMessage = MessageHelpers.AssertReadMessage<HelloReply>(await response.Content.ReadAsByteArrayAsync().DefaultTimeout());
             Assert.AreEqual("Hello World", responseMessage.Message);
+        }
+
+        [Test]
+        public async Task SendHeadersTwice_ThrowsException()
+        {
+            // Arrange
+            var requestMessage = new HelloRequest
+            {
+                Name = "World"
+            };
+
+            var ms = new MemoryStream();
+            MessageHelpers.WriteMessage(ms, requestMessage);
+
+            // Act
+            var response = await Fixture.Client.PostAsync(
+                "Greet.Greeter/SayHelloSendHeadersTwice",
+                new StreamContent(ms)).DefaultTimeout();
+
+            // Assert
+            Assert.AreEqual(HttpStatusCode.OK, response.StatusCode);
+            Assert.AreEqual("identity", response.Headers.GetValues("grpc-encoding").Single());
+            Assert.AreEqual("application/grpc", response.Content.Headers.ContentType.MediaType);
+
+            var responseMessage = MessageHelpers.AssertReadMessage<HelloReply>(await response.Content.ReadAsByteArrayAsync().DefaultTimeout());
+            Assert.AreEqual("Exception validated", responseMessage.Message);
         }
     }
 }
--- a/test/Grpc.AspNetCore.Server.Tests/HttpContextServerCallContextTests.cs
+++ b/test/Grpc.AspNetCore.Server.Tests/HttpContextServerCallContextTests.cs
@@ -16,7 +16,9 @@
 
 #endregion
 
+using System;
 using System.Net;
+using System.Threading.Tasks;
 using Grpc.AspNetCore.Server.Internal;
 using Grpc.Core;
 using Microsoft.AspNetCore.Http;
@@ -46,7 +48,7 @@ namespace Grpc.AspNetCore.Server.Tests
         }
 
         [Test]
-        public void WriteResponseHeadersAsyncCore_AddsMetadataToResponseHeaders()
+        public async Task WriteResponseHeadersAsyncCore_AddsMetadataToResponseHeaders()
         {
             // Arrange
             var httpContext = new DefaultHttpContext();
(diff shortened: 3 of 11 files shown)


src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -74,7 +74,15 @@ protected override Metadata RequestHeadersCore
 
                     foreach (var header in HttpContext.Request.Headers)
                     {
-                        if (!header.Key.StartsWith(':'))
+                        if (header.Key.StartsWith(':'))
+                        {
+                            continue;
```
JamesNK: Comment why

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -74,7 +74,15 @@ protected override Metadata RequestHeadersCore
...
+                        }
+                        else if (header.Key.EndsWith(Metadata.BinaryHeaderSuffix))
```
JamesNK: Ordinal compare. Is it case sensitive? Add tests either way.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -116,13 +124,22 @@ protected override ContextPropagationToken CreatePropagationTokenCore(ContextPro
 
         protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
         {
+            // Headers can only be written once. Throw on subsequent call to write response header instead of silent no-op.
+            if (HttpContext.Response.HasStarted)
+            {
+                throw new InvalidOperationException("Response headers can only be sent once per call.");
```
JamesNK:
```suggestion
                throw new InvalidOperationException("Response headers have already been sent. Response headers can only be sent once per call.");
```

test/Grpc.AspNetCore.Server.Tests/HttpContextServerCallContextTests.cs
```diff
@@ -55,12 +57,29 @@ public void WriteResponseHeadersAsyncCore_AddsMetadataToResponseHeaders()
...
         }
 
+        [Test]
+        public async Task WriteResponseHeadersAsyncCore_Base64EncodesBinaryResponseHeaders()
+        {
+            // Arrange
+            var headerBytes = new byte[] { 0x01, 0x02, 0x03, 0x04 };
+            var httpContext = new DefaultHttpContext();
+            var metadata = new Metadata();
+            metadata.Add("foo-bin", headerBytes);
```
JamesNK: What happens for `FOO-BIN`?

testassets/FunctionalTestsWebsite/Greeter.cs
```diff
@@ -62,4 +63,25 @@ public override Task SayHellosSendHeadersFirst(HelloRequest request, IServerStre
...
+    {
+        context.WriteResponseHeadersAsync(null);
+
+        try
+        {
+            context.WriteResponseHeadersAsync(null);
+        }
+        catch (InvalidOperationException e)
+        {
+            if (e.Message != "Response headers can only be sent once per call.")
```
JamesNK: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when

test/FunctionalTests/UnaryMethodTests.cs
```diff
@@ -164,5 +164,31 @@ public async Task MessageSentInMultipleChunks_SuccessResponse()
...
+                "Greet.Greeter/SayHelloSendHeadersTwice",
+                new StreamContent(ms)).DefaultTimeout();
+
+            // Assert
+            Assert.AreEqual(HttpStatusCode.OK, response.StatusCode);
+            Assert.AreEqual("identity", response.Headers.GetValues("grpc-encoding").Single());
+            Assert.AreEqual("application/grpc", response.Content.Headers.ContentType.MediaType);
+
+            var responseMessage = MessageHelpers.AssertReadMessage<HelloReply>(await response.Content.ReadAsByteArrayAsync().DefaultTimeout());
+            Assert.AreEqual("Exception validated", responseMessage.Message);
```
jtattermusch: In general, I find relying on magic values set in a completely different place in code (the service impl in this case) not very readable.  One approach that I found useful when writing tests from Grpc.Core was to define a "universal" service that allows setting implementations of serverside handlers directly in the test.
Then many tests look like a serverside handler that's provided as a lambda at the start of the test in the Arrange phase.   When reading the test, one can easily see what corresponding serverside handler is doing and the test becomes much more readable.
Another advantage is that you don't need to come up with complicated serverside handler implementations that would be good enough for use in many different test cases.

See example here: 
https://github.com/grpc/grpc/blob/f4079fe185a51967e2773587c7804bd329936a16/src/csharp/Grpc.Core.Tests/ClientServerTest.cs#L73

A side note: for some tests, the test service uses a custom marshaller and the requests and responses are not protobuf messages, but strings. This makes the test code a bit more compact in some cases.

jtattermusch: This is basically just a suggestion - fine to address later, but I think it might be worthwhile to at least consider.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: I'm using the same error message as the Grpc.Core implementation.

JamesNK: Hmm ok. I think this is more clear but we could explain more in logging later.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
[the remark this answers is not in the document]
JamesNK: You were using EndsWith with the current culture.

Passing StringComparison.Ordinal to it ignores the culture and compares the chars
https://docs.microsoft.com/en-us/dotnet/api/system.string.endswith?view=netframework-4.7.2#System_String_EndsWith_System_String_System_StringComparison_

test/FunctionalTests/UnaryMethodTests.cs
JamesNK: Interesting idea. It is a bit more difficult for our functional tests because we simulate building the entire ASP.NET Core application.

I *think* I could emulate something like this. Will look at it later

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: Good point, it's case sensitive in the Grpc.Core implementation. Will add a test.

JamesNK: I recall someone saying that Metadata are always lower case. Is that right?

test/Grpc.AspNetCore.Server.Tests/HttpContextServerCallContextTests.cs
JunTaoLuo: An error is thrown by Metadata.Add(). Since all binary headers must have `-bin` suffix with that exact casing. I don't think it's necessary to add a test for this since it's an implementation detail in Metadata/Metadata.Entry.

test/FunctionalTests/UnaryMethodTests.cs
JunTaoLuo: Filed https://github.com/grpc/grpc-dotnet/issues/80 to track

test/Grpc.AspNetCore.Server.Tests/HttpContextServerCallContextTests.cs
JunTaoLuo: Oops, I was testing something different. The casing is normalized. Will add test case.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -74,7 +74,18 @@ protected override Metadata RequestHeadersCore
...
                     {
-                        if (!header.Key.StartsWith(':'))
+                        // ASP.NET Core includes pseudo headers in the set of request headers
+                        // whereas, they are not in gRPC implementations. We will filter them
+                        // out when we construct the list of headers on the context.
+                        if (header.Key.StartsWith(':'))
+                        {
+                            continue;
+                        }
+                        else if (header.Key.EndsWith(Metadata.BinaryHeaderSuffix, StringComparison.OrdinalIgnoreCase))
```
JamesNK: If headers are always normalized as lowercase then StringComparison.Oridinal is slightly faster.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: Yup, https://github.com/grpc/grpc-dotnet/pull/69#discussion_r254132086, I misremembered and was thinking of something else.

JunTaoLuo:
@jtattermusch would be nice if we can enable the "Squash and Merge" option on PRs so we don't have to manually do them. Also, can we make the Travis checks mandatory for check-ins?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: but they are not lowercased in the Request.Headers. When kestrel surfaces "Foo-Bin" I think we should interpret that as a binary header.

JamesNK: Ah I see, this is going from HttpContext to Metadata. Looks good then.

jtattermusch:
> @jtattermusch would be nice if we can enable the "Squash and Merge" option on PRs so we don't have to manually do them. Also, can we make the Travis checks mandatory for check-ins?

I've done both. Sorry I forgot to do that earlier as I promised.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
jtattermusch: this might be an opportunity for future optimization - in http/2 all header keys are sent in lowercase, and ASP.NET core does seem to adjust casing for some header keys (the well known ones from HTTP1) but I assume it won't change casing for header keys it doesn't recognize - so we can be almost sure that all the received "-bin" suffixes will be lowercase in practice? Perhaps a fastpath can be added later.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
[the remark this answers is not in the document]
jtattermusch: ah, sorry, I somehow missed that you're passing `StringComparison.OrdinalIgnoreCase`

Issue #41, opened 2019-01-23
Use IAsyncEnumerable instead of IAsyncStreamReader


JunTaoLuo:
We should consider using the new C# 8 feature [`IAsyncEnumerable`](https://blogs.msdn.microsoft.com/dotnet/2018/11/12/building-c-8-0/) for streaming RPC calls.

This may need some changes in the Grpc.Core and Grpc.Core.Api packages as well cc @jtattermusch

JamesNK:
Potential idea that wouldn't require breaking the method signature:

```cs
foreach await (var message in requestStream.AsEnumerable())
{
}
```

@davidfowl

davidfowl:
That's not bad. `ChannelReader<T>` added a ReadAllAsync() method that returns the `IAsyncEnumerable<T>`

onovotny:
Hi folks,

As a heads up, we've released Ix Async 4.0 preview, which uses the BCL `IAsyncEnumerable` types and provides the LINQ operators for those. As a result of this, Ix Async 4 is a breaking change over 3.x. That said, we do provide support for using `IAsyncEnumberable` on .NET 4.5 and .NET Standard 2.0 should you need to target those platforms.

davidfowl:
What's the plan here?

JamesNK:
Plan for what?

I created an issue for gRPC to migrate await from Ix Async - https://github.com/grpc/grpc/issues/18592. No progress so far.

We haven't looked at IAsyncEnumerable support yet. Template and gRPC client have been the priority.

davidfowl:
OK here are some more options:
- Support 2 signatures for streaming, one IAsyncEnumerable based (the new one which BTW is available now on ns2.0 (https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/1.0.0-preview6.19259.10)
- Expose an extension method like you describe above and keep the Ix dependency (though I don't want to do that).

JamesNK:
The first option would require a new setting to be added to code generation from proto files and for C Core C# to support the new type. That would be a lot of work and gRPC folks are hesitant to add too many settings to code generation.

I prefer the extension method approach. It doesn't require keeping the Ix dependency. The types in method signatures don't come form Ix, they implement Ix IAsyncEnumerator. Removing that interface will likely not effect the extension method because we'll cast to a known internal type.

JamesNK:
I looked into this. Adding an extension method is extremely trivial.

```cs
[...]

PR #82, opened 2019-02-09
 Only depend on Grpc.Core.Api

jtattermusch:
We forgot to switch to dependency only on the Api package.

- work in progress, we need to figure out if stream extensions should be moved to Grpc.Core.Api, redefined in Grpc.AspNetCore.Server  or if we should just not support them.

### DIFF
--- a/testassets/InteropTestsWebsite/AsyncStreamExtensions.cs
+++ b/testassets/InteropTestsWebsite/AsyncStreamExtensions.cs
@@ -1,0 +1,41 @@
+#region Copyright notice and license
+
+// Copyright 2015 gRPC authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Grpc.Core;
+
+namespace InteropTestsWebsite
+{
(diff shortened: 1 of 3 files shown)


testassets/InteropTestsWebsite/AsyncStreamExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Grpc.Core.Utils
+{
+    /// <summary>
+    /// Extension methods that simplify work with gRPC streaming calls.
+    /// </summary>
+    public static class AsyncStreamExtensions
```
jtattermusch: taken from https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core/Utils/AsyncStreamExtensions.cs

JamesNK: Copying a small piece of code for the tests is fine.

If we added a help method for working with async it would probably be something new in relation to https://github.com/grpc/grpc-dotnet/issues/41

jtattermusch:
@davidfowl  @JamesNK  @JunTaoLuo    I think we still need to determine if we want to:
- option 1)  move AsyncStreamExtensions.cs to Grpc.Core.Api,
- option 2) recreate our own version of AsyncStreamExtensions
- option 3) not support the extensions at all (which would mean that existing serverside implementations might not work as they might rely on these extensions)

I can't decide between Option 1 and 2:
2)  it allows us to do performance optimizations of those extensions that rely on .NET core 3 (in case we need to), it's only a small portion of code to duplicate and it would only require adding a `using` statement to migrate old code. The disadvantage is that if Grpc.Core and Grpc.AspNetCore.Server are both used in one project, the extensions could collide if both extensions classes are imported by using.

RainingNight:
Option 2 good.

JamesNK:
> * option 3) not support the extensions at all (which would mean that existing serverside implementations might not work as they might rely on these extensions)

People still have the choice to reference Grpc.Core if they are porting existing code and want the extension methods.

I like option 3. And we can always add something new later.

davidfowl:
I agree with @JamesNK , lets do 3.

PR #84, opened 2019-02-10
Deadline cancellation

JamesNK:
* Adds a factory for creating call handlers
* Some initial server logging
* Abort on server when deadline is exceeded
* Test trailers in functional tests


src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -167,18 +171,28 @@ public void Initialize()
                 }
 
                 _deadline = Clock.UtcNow.Add(timeout);
-                _cts = new CancellationTokenSource(timeout);
+
+                _timeoutCts = new CancellationTokenSource(timeout);
+                _timeoutCts.Token.Register(DeadlineExceeded, timeout);
+
+                _linkedCts = CancellationTokenSource.CreateLinkedTokenSource(_timeoutCts.Token, HttpContext.RequestAborted);
```
JamesNK: Is it it worth having a cancellation token for the timeout? This could simplified as just a Timer that calls `DeadlineExceeded` which then aborts the HttpContext. `HttpContext.RequestAborted` could always be used.

@davidfowl

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -167,18 +171,28 @@ public void Initialize()
...
             {
                 _deadline = DateTime.MaxValue;
-                _cts = new CancellationTokenSource();
             }
         }
 
+        private void DeadlineExceeded(object timeout)
+        {
+            DeadlineExceeded(_logger, (TimeSpan)timeout);
+            HttpContext.Abort();
```
JamesNK: I noticed some response headers that gRPC sets when the server his the deadline. Still need to set those. Still need to find out what happens when gRPC has started streaming and it hits the deadline.

JamesNK:
@davidfowl

build/sources.props
```diff
@@ -3,6 +3,7 @@
     <RestoreSources>
       $(RestoreSources);
       https://api.nuget.org/v3/index.json;
+      https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json;
```
JamesNK: Microsoft.Extensions.Logging.Testing isn't published on nuget.org

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
davidfowl: You know how I feel about cancellation tokens and this isn't very pay for play. Does this always get allocated?

JamesNK: It is pay for play. Cancellation tokens are only allocated if the client specifies a deadline.

jtattermusch:
@JamesNK  a PR description would be useful to better understand what this is about. It's a big PR and it seems that there are multiple different changes.

build/sources.props
JunTaoLuo: It's intended to be for internal testing purposes only. I see that the only thing you are using in it is TestLogger in which case, I think it's best to copy that code over instead.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -154,48 +159,54 @@ protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
...
-                // CancellationTokenSource does not support greater than int.MaxValue milliseconds
-                if (timeout.TotalMilliseconds > int.MaxValue)
-                {
-                    throw new InvalidOperationException("A timeout greater than 2147483647 milliseconds is not supported.");
-                }
-
                 _deadline = Clock.UtcNow.Add(timeout);
-                _cts = new CancellationTokenSource(timeout);
+
+                _deadlineTimer = new Timer(DeadlineExceeded, timeout, timeout, Timeout.InfiniteTimeSpan);
```
davidfowl: This is a common enough pattern (abort the connection after some timeout, that I think we should make it a first class API).

build/sources.props
JunTaoLuo: I would still prefer copying code instead of adding a reference to the Logging.Testing package from our myget feed.

JamesNK: I'm going to leave this using Logging.Testing for now. If the reference ever becomes an issue then we can replicate TestLogger locally.

PR #85, opened 2019-02-10
Add configuration for nuget building job

jtattermusch:
The job will run on google-internal CI "Kokoro". Another step in the internal pipeline (not visible externally) will be used to sign the nuget and contained `dll`s.

### DIFF
--- a/kokoro/build_nuget.sh
+++ b/kokoro/build_nuget.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Copyright 2019 The gRPC Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


kokoro/build_nuget.sh
```diff
@@ -0,0 +1,35 @@
...
+tar zxf dotnet-sdk.tar.gz -C $PWD/dotnet
+export PATH="$PWD/dotnet:$PATH"
+
+# TODO(jtattermusch): remove this before release, otherwise 
+# references will be broken.
+./build/get-grpc.sh
+
+mkdir -p artifacts
+
+# TODO(jtattermusch): set the package version in csproj files.
```
jtattermusch: we'll need to decide on the versioning scheme.

JunTaoLuo: I think we should follow the rest of the gRPC versioning scheme since eventually we want to have the same version numbers as Grpc.Core package. 

Are you concerned with doing that now? In case there are breaking changes in the first few iterations? Is there a pre-release version scheme you use until we RTM?

jtattermusch: Let's discuss versioning in our syncup.

Issue #81, opened 2019-02-08
Exclude gRPC headers from call context?


JamesNK:
re: https://github.com/grpc/grpc-dotnet/pull/73#discussion_r255155112

Investigate what Grpc.Core does with grpc specific headers and make Grpc.AspNetCore.Server follow.

dhhoang:
Hi, I'm looking to contribute to this repo.
I spent sometime going through `Grpc.Core` and the C library code base. With regard to #73 the C lib does expose grpc-specific metadata (see https://github.com/grpc/grpc/blob/ff51441a29403c50495dc6ee3708e04ff088871d/include/grpc/impl/codegen/grpc_types.h#L497) and the server/client apps can query them. 

How they are handled depends on the metadata itself. For ex with deadline/timeout the client will cancel the request by sending an DEADLINE_EXCEEDED to the application (possibly a RST_STREAM to the server), while the server does not have any default behavior (see comments https://github.com/grpc/grpc/blob/master/src/core/ext/filters/deadline/deadline_filter.cc#L189 ).

Depending on what behavior we want there might be different ways to handle these headers but I think exposing them to applications is a good idea. That might also depend on whether or not `grpc-dotnet
` will have a dependency on `Grpc.Core`.

JamesNK:
Actual cancellation and setting the right status is in this PR - https://github.com/grpc/grpc-dotnet/pull/84

A simple way to find out what headers ServerCallContext exposes gRPC specific headers is to run the native .NET server and test. I do know the client has a specific argument in the call options for setting a deadline that is separate from other headers.

dhhoang:
Turns out that the way `Grpc.Core` deals with headers is a little more complicated. I tried running the native server and it looks like the default behavior is to maintain the headers, except for some.
`grpc-timeout` is not stored in the metadata collection but the deadline is exposed : https://github.com/grpc/grpc/blob/ff51441a29403c50495dc6ee3708e04ff088871d/include/grpc/impl/codegen/grpc_types.h#L506 
Some other headers have special handling by the framework. For example a request with `grpc-status` set to anything other than 0 will be considered an error: https://github.com/grpc/grpc/blob/ff51441a29403c50495dc6ee3708e04ff088871d/src/core/ext/transport/chttp2/transport/parsing.cc#L425 
Is that ok if I send an PR for this?

JamesNK:
[...]

PR #87, opened 2019-02-12
Add grpc-status check on server

dhhoang:
With regards to #81 , there are some grpc-specific headers that need special processing. 

This PR adds checking for `grpc-status` header in the request. The behavior implemented in the native library is that only "0" is a valid value: https://github.com/grpc/grpc/blob/ff51441a29403c50495dc6ee3708e04ff088871d/src/core/ext/transport/chttp2/transport/parsing.cc#L425.

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
@@ -65,5 +65,21 @@ namespace Grpc.AspNetCore.Server.Internal
 
             throw new InvalidOperationException("Error reading grpc-timeout value.");
         }
+
+        public static void CheckStatus(HttpContext httpContext)
+        {
+            //TODO: is status like " 0" valid?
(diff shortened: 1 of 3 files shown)


thelinuxfoundation:

* User @dhhoang isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

dhhoang:
Done signing CLA

jtattermusch:
Why is such check important? It doesn't seem that under normal circumstances request would be sent with the "grpc-status" header. I'd be curious if grpc-java and grpc-go are performing the same check.

dhhoang:
I'm not sure how "important" this is but it's implemented in the core C lib so most certainly other grpc language bindings are doing the same.

jtattermusch:
@yashykt could you provide some insight about the importance of the check mentioned in https://github.com/grpc/grpc-dotnet/pull/87#issue-252174386? I'm a bit puzzled by it.

dhhoang:
Got it. Closing the PR.

PR #90, opened 2019-02-13
Update Grpc dependencies

JamesNK:
Use dependencies with build number to avoid cache issues.


JamesNK:
And delete a leftover file from merging.

JamesNK:
CI is failing and the logs aren't providing any information why 🤷‍♂️

JunTaoLuo:
Hmm looks like the `read` command in the get-grpc.sh script was exiting with an error code. I'm not sure why but adding a new line in the file seems to fix it. I'll check the sh and ps1 scripts and cross-usage between them. But for now this is unblocked.

davidfowl:
I wonder if we can use the maestro bot to update these automagically? @JunTaoLuo something to look into or maybe we don't need it because it won't change that much?

JunTaoLuo:
If it become a frequent enough issue we'll look into better automation.

PR #93, opened 2019-02-14
interop server: support secure connections

jtattermusch:
- import test creds normally used for interop testing
- support `--port` and `--use_tls` cmdline args.


testassets/InteropTestsWebsite/Resources.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.IO;
+
+namespace InteropTestsWebsite
+{
+    public static class Resources
+    {
+        public static string CertDir = Path.Combine(GetSolutionDirectory(), "testassets/InteropTestsWebsite", "Certs");
+        public static string ServerPFXPath = Path.Combine(CertDir, "server1.pfx");
+
+        private static string GetSolutionDirectory()
```
jtattermusch: logic copied from examples for now, but I think a want to use a simple way.

testassets/InteropTestsWebsite/Program.cs
```diff
@@ -29,16 +31,29 @@ public static void Main(string[] args)
...
-        public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
-            WebHost.CreateDefaultBuilder(args)
-                .ConfigureKestrel(options =>
+        public static IWebHostBuilder CreateWebHostBuilder(string[] args)
+        {
+            var builder = WebHost.CreateDefaultBuilder(args);
+
+            // Support --port and --use_tls cmdline arguments normally supported
+            // by gRPC interop servers.
+            int port = int.Parse(builder.GetSetting("port") ?? "50052");
```
jtattermusch: not really proud of this, but it works. Is there a more idiomatic way? :-)

JunTaoLuo: [Command line configuration](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-2.2#command-line-configuration-provider)! This is added by default by the WebHost.CreateDefaultBuilder.

testassets/InteropTestsWebsite/Certs/ca.pem
```diff
@@ -0,0 +1,15 @@
+-----BEGIN CERTIFICATE-----
```
JunTaoLuo: Maybe we can use these certs for the examples as well. Less duplication and infrastructure.

testassets/InteropTestsWebsite/Resources.cs
JunTaoLuo: Yea, duplication is always undesired, not a big deal though.

JamesNK: I'm not a fan of this way of referencing certs. Is there a reason why we have this pattern?

I'd prefer that we include the certs are content in the project, copy them to the output directory on build and reference them there.

Should change this both for the samples and the interop test site.

testassets/InteropTestsWebsite/Program.cs
JunTaoLuo: Something like 
```c#
                .ConfigureKestrel((context, options) =>
                {
                    context.Configuration.GetValue<int>("port");
                    ...
                }
```

testassets/InteropTestsWebsite/Resources.cs
JunTaoLuo: I recall trying this when I set this up but I didn't get it to work. Copying to the output directory means the files will be in the `bin/Debug/netcoreapp3.0` folder. However, when you invoke `dotnet run`, the root directory is set to the project directory, which means you need to include `bin/Debug/netcoreapp3.0` in your path to the certificate file. We found that resolving the full path more robust (against Debug/Release for example). 

Is there any new csproj mechanisms that I'm not aware of that can make this work?

JamesNK: Lets keep going with this for now Because We Know It Works, and clean it up later.

JamesNK: https://github.com/grpc/grpc-dotnet/issues/94

testassets/InteropTestsWebsite/Certs/ca.pem
jtattermusch: sg, let's address in a follow up PR?

testassets/InteropTestsWebsite/Program.cs
jtattermusch: Done. Thanks for help!

jtattermusch:
I think this should be good to merge. PTAL.

testassets/InteropTestsWebsite/Program.cs
```diff
@@ -16,9 +16,12 @@
 
 #endregion
 
+using System.Runtime.InteropServices;
+
```
JamesNK: nit: remove extra line

Issue #80, opened 2019-02-08
Improve functional test readability


JunTaoLuo:
Comment: https://github.com/grpc/grpc-dotnet/pull/79#discussion_r255013018.

This looks like a worthwhile suggestion but we'll need to add a bit of infrastructure to support.

dhhoang:
I think this can be done by setting up a `TestServer` that maps grpc methods to lambda function as the arrange phase in the test method rather than having one static server. 
This is probably less efficient but that shouldn't matter too much?
Is this up-for-grabs? I'd like to send a PR.

jtattermusch:
Looks like #105 has established a nice way to define the handlers used in testing directly in the test case method, but we haven't really converted the existing tests to use this approach.
(we have a bunch of "overloads" of SayHello defined here: https://github.com/grpc/grpc-dotnet/blob/2c2f93ac6773085466baef79e8f8dde0e8536c48/testassets/Proto/greet.proto#L24 and most of them seem to be only used once). Are we planning to do some test refactoring to make the tests more readable?

PR #95, opened 2019-02-15
Improve functional test

dhhoang:
Hi, in this PR I'd like to introduce an approach to address #80 .
The idea is to expand `MapGrpcService` to allow services to be registered using custom service creation and dispose functions: 
```csharp
public static IEndpointConventionBuilder MapGrpcService<TService>(
            this IEndpointRouteBuilder builder,
            Func<IServiceProvider, TService> createService,
            Action<TService> releaseService = null) where TService : class
```
Services registered with this method would bypass the lifecycle managed by `IGrpcServiceActivator` and instead invoke these functions on creation/release. Functional test methods can then create an ASP.NET core `TestServer` as part of the arrange phase and providing custom service implementation (either some test implementation or mocked objects) 
```csharp
var gMock = new Mock<Greeter.GreeterBase>();
gMock
    .Setup(g => g.SayHello(It.IsAny<HelloRequest>(), It.IsAny<ServerCallContext>()))
    .Returns((HelloRequest r, ServerCallContext ctx) =>
    {
        return Task.FromResult(new HelloReply { Message = message });
    });
var obj = gMock.Object;
TestServer server = UniversalTestServerBuilder.WithServices(r =>
{
    r.Register(() => obj);
}).Build();
using (server)
using (var client = server.CreateClient())
{
...
```
I believe with this approach the static test website would no longer be necessary.
This PR is not meant to be complete (some implementations are not done yet and unit tests are missing) but rather a POC to (hopefully) get some feedback.

### DIFF
--- a/src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
+++ b/src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
@@ -76,6 +76,60 @@ namespace Microsoft.AspNetCore.Builder
             }
         }
 
+        public static IEndpointConventionBuilder MapGrpcService<TService>(
+            this IEndpointRouteBuilder builder,
+            Func<IServiceProvider, TService> createService,
+            Action<TService> releaseService = null) where TService : class
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            //the service creation function must exist
+            if (createService == null)
+            {
+                throw new ArgumentNullException(nameof(createService));
+            }
+
+            ValidateServicesRegistered(builder.ServiceProvider);
+
+            // Unlike the mapping for a concrete server implementation, TService could be inferred as the TServiceBase
+            // type itself, so we trace up the tree from there.
+            var baseType = typeof(TService);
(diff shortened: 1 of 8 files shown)


JamesNK:
Hi

I have a different solution for improving functional tests that I prefer. Check out https://github.com/grpc/grpc-dotnet/pull/105 and let me know if there are any scenarios that this PR enables.

dhhoang:
> Hi
> 
> I have a different solution for improving functional tests that I prefer. Check out #105 and let me know if there are any scenarios that this PR enables.

Unless I'm mistaken the main point of the functional tests is to test the server end-to-end. In #105 it seems like the "default" `InvokerFactory` is `ReflectionMethodInvokerFactory`, while in the functional tests it is a mock object. So if `ReflectionMethodInvokerFactory` later changes (or we invent a new `IGrpcMethodInvokerFactory` somehow), then we don't get to test _that_. Would it be a problem?

JamesNK:
I don't think so because there are going to be a mix of services: some defined inline, some defined traditionally in Startup.cs. Services defined in Startup.cs will test invoker factory.

dhhoang:
Understood :) . I was just a bit concerned by the test logic being written in different places. Closing the PR.

PR #96, opened 2019-02-15
Handle errors in service methods

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/47

Handle user code errors in service methods.
Error is logged on the server.
Setting EnableDetailedErrors to true also returns the exception message with grpc-message.
Fix an issue in calculating options for services without explicit configuration.
Handle errors when serializing and deserializing content.
Handle unary and client streaming service methods returning null message.
Return RpcException trailers.


src/Grpc.AspNetCore.Server/Internal/HttpContextStreamWriter.cs
```diff
@@ -39,6 +39,14 @@ public WriteOptions WriteOptions
             set => _context.WriteOptions = value;
         }
 
-        public Task WriteAsync(TResponse message) => _context.HttpContext.Response.BodyPipe.WriteMessageAsync(message, _context, _serializer);
+        public Task WriteAsync(TResponse message)
+        {
+            if (message == null)
```
davidfowl: Why null check a internal type? That just bloats the method. Just assert.

perf/Grpc.AspNetCore.Microbenchmarks/Internal/MessageHelpers.cs
```diff
@@ -21,20 +21,22 @@
...
+        private static readonly HttpContextServerCallContext TestServerCallContext = new HttpContextServerCallContext(new DefaultHttpContext(), new GrpcServiceOptions(), NullLogger.Instance);
 
         public static void WriteMessage<T>(Stream stream, T message) where T : IMessage
         {
             var messageData = message.ToByteArray();
 
             var pipeWriter = new StreamPipeWriter(stream);
 
-            PipeExtensions.WriteMessageAsync(pipeWriter, messageData, TestServiceOptions, flush: true).GetAwaiter().GetResult();
+            PipeExtensions.WriteMessageAsync(pipeWriter, messageData, TestServerCallContext, flush: true).GetAwaiter().GetResult();
```
JamesNK: I had an initial approach that needed the context which is why the pipe extensions now take it, but it isn't required right now. I suspect at some point we'll need info off it so I have left it this way for now.

src/Grpc.AspNetCore.Server/Internal/HttpContextStreamWriter.cs
JamesNK: The instance is internal, the interface is public. This method will be called by user code in service methods.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -32,21 +32,27 @@ internal static class PipeExtensions
         private const int MessageDelimiterSize = 4; // how many bytes it takes to encode "Message-Length"
         private const int HeaderSize = MessageDelimiterSize + 1; // message length + compression flag
 
-        public static Task WriteMessageAsync<TResponse>(this PipeWriter pipeWriter, TResponse response, GrpcServiceOptions serviceOptions, Func<TResponse, byte[]> serializer, WriteOptions writeOptions)
+        private static readonly Status MessageCancelledStatus = new Status(StatusCode.Internal, "Incoming message cancelled.");
+        private static readonly Status AdditionalDataStatus = new Status(StatusCode.Internal, "Additional data after the message received.");
+        private static readonly Status IncompleteMessageStatus = new Status(StatusCode.Internal, "Incomplete message.");
+        private static readonly Status SendingMessageExceedsLimitStatus = new Status(StatusCode.ResourceExhausted, "Sending message exceeds the maximum configured message size.");
+        private static readonly Status ReceivedMessageExceedsLimitStatus = new Status(StatusCode.ResourceExhausted, "Received message exceeds the maximum configured message size.");
```
jtattermusch: FTR https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
```diff
@@ -49,35 +49,44 @@ public override async Task HandleCallAsync(HttpContext httpContext)
...
+                    new HttpContextStreamReader<TRequest>(httpContext, serverCallContext, Method.RequestMarshaller.Deserializer),
+                    serverCallContext);
+
+                if (response == null)
                 {
-                    response = await _invoker(
-                        service,
-                        new HttpContextStreamReader<TRequest>(httpContext, ServiceOptions, Method.RequestMarshaller.Deserializer),
-                        serverCallContext);
+                    throw new RpcException(new Status(StatusCode.Cancelled, "Cancelled"));
```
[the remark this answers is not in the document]
JamesNK: I tested what Grpc.Core did when null was returned from a unary and it returned a Cancelled status.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
JamesNK: Yeah I'm aware of that page and used it as a guide. This is the correct status, right?

Issue #97, opened 2019-02-15
several interop tests failing against grpc-java client


jtattermusch:
Context: Most interop test pass when grpc-go or c-based library is used, but when grpc-java is used, there seems to be some problem.

E.g.:
```
~/github/grpc-java$ ./run-test-client.sh --server_host=localhost --server_port=50052 --test_case=large_unary --use_tls=false
```

Results in
```
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HLKJ9F9LRHFI", Request id "0HLKJ9F9LRHFI:00000003": An unhandled exception was thrown by the application.
System.FormatException: The input is not a valid Base-64 string as it contains a non-base 64 character, more than two padding characters, or an illegal character among the padding characters.
   at System.Convert.FromBase64CharPtr(Char* inputPtr, Int32 inputLength)
   at System.Convert.FromBase64String(String s)
   at Grpc.AspNetCore.Server.Internal.HttpContextServerCallContext.get_RequestHeadersCore() in /usr/local/google/home/jtattermusch/github/grpc-dotnet/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs:line 95
   at Grpc.Core.ServerCallContext.get_RequestHeaders()
   at Grpc.Testing.TestServiceImpl.EnsureEchoMetadataAsync(ServerCallContext context) in /usr/local/google/home/jtattermusch/github/grpc-dotnet/testassets/InteropTestsWebsite/TestServiceImpl.cs:line 99
   at Grpc.Testing.TestServiceImpl.UnaryCall(SimpleRequest request, ServerCallContext context) in /usr/local/google/home/jtattermusch/github/grpc-dotnet/testassets/InteropTestsWebsite/TestServiceImpl.cs:line 40
   at Grpc.AspNetCore.Server.Internal.UnaryServerCallHandler`3.HandleCallAsync(HttpContext httpContext) in /usr/local/google/home/jtattermusch/github/grpc-dotnet/src/Grpc.AspNetCore.Server/Internal/UnaryServerCallHandler.cs:line 70
   at Microsoft.AspNetCore.Routing.EndpointRoutingMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
[...]

PR #98, opened 2019-02-15
Parse un-padded value in binary header

dhhoang:
Addresses #97 .
Could be done more efficiently (for the rare case of un-padded headers) but enough to make grpc-java interop test pass.

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
@@ -92,7 +92,7 @@ namespace Grpc.AspNetCore.Server.Internal
                         }
                         else if (header.Key.EndsWith(Metadata.BinaryHeaderSuffix, StringComparison.OrdinalIgnoreCase))
                         {
-                            _requestHeaders.Add(header.Key, Convert.FromBase64String(header.Value));
+                            _requestHeaders.Add(header.Key, ParseBinaryHeader(header.Value));
                         }
                         else
                         {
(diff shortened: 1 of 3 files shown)


src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -208,5 +208,29 @@ public void Dispose()
...
             _deadlineTimer?.Dispose();
         }
+
+        private static byte[] ParseBinaryHeader(string base64)
+        {
+            string decodable;
+            switch (base64.Length % 4)
+            {
+                case 0:
+                    // no padding
```
JamesNK: Would "either has required padding or no padding needed" be more accurate?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -208,5 +208,29 @@ public void Dispose()
         {
             _deadlineTimer?.Dispose();
         }
+
+        private static byte[] ParseBinaryHeader(string base64)
```
JamesNK: Move to GrpcProtocolHelpers

dhhoang:
Updated to address comments.

jtattermusch:
LGTM, except the unpadded headers aren't the "rare" case, because implementations should send unpadded according to the spec. We should probably add a TODO for more efficient implementation.

I'm actually a bit surprised that `Convert.FromBase64String()` doesn't have an option to accept unpadded base64.

PR #102, opened 2019-02-16
Add basic nuget package properties

jtattermusch:


src/Grpc.AspNetCore.Server/Grpc.AspNetCore.Server.csproj
```diff
@@ -1,5 +1,16 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
+  <PropertyGroup>
+    <Authors>The gRPC Authors</Authors>
+    <Copyright>Copyright 2019 The gRPC Authors</Copyright>
+    <Description>gRPC support for ASP.NET Core</Description>
+    <PackageIconUrl>https://github.com/grpc/grpc.github.io/raw/master/img/grpc_square_reverse_4x.png</PackageIconUrl>
+    <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
+    <PackageProjectUrl>https://github.com/grpc/grpc-dotnet</PackageProjectUrl>
+    <PackageTags>gRPC RPC Protocol HTTP/2 ASP.NET</PackageTags>
```
JunTaoLuo: let's add the `aspnetcore` tag. This is what we use in our packages: https://github.com/aspnet/AspNetCore/blob/release/2.1/src/Hosting/Hosting/src/Microsoft.AspNetCore.Hosting.csproj#L8. Also, NuGet interprets this field with space delimitation so I'm not sure if the `Protocol` tag is necessary since it's too generic.

PR #103, opened 2019-02-18
WIP Dynamic service methods in functional tests

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/80

Also adds a `MapGrpcService` that allows explicit binding. This is for code gen services that place BindService in a different class structure.

### DIFF
--- a/src/Grpc.AspNetCore.Server/GrpcServiceExtensions.cs
+++ b/src/Grpc.AspNetCore.Server/GrpcServiceExtensions.cs
@@ -62,6 +62,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
             services.AddRouting();
             services.AddOptions();
+            services.AddLogging();
             services.TryAddSingleton<GrpcMarkerService>();
             services.TryAddSingleton(typeof(ServerCallHandlerFactory<>));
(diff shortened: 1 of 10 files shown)


src/Grpc.AspNetCore.Server/GrpcServiceExtensions.cs
```diff
@@ -62,6 +62,7 @@ public static IGrpcServerBuilder AddGrpc(this IServiceCollection services)
 
             services.AddRouting();
             services.AddOptions();
+            services.AddLogging();
```
davidfowl: For easy testing? Nothing else in our stack does this.

JamesNK: I noticed AddGrpc by itself errors. Is that fine?

davidfowl: Logging is added by the host, why does it error?

JamesNK: It errored by itself in a test. If every other framework leaves it to the host then I'll remove it.

JamesNK:
Feel free to review but I might change this up a bit. I'm going to try out a couple more things to remove the hard dependency services have on `typeof(Service).GetMethod(Method.Name)`.

PR #105, opened 2019-02-18
Dynamic service methods in functional tests - with invoker factory

JamesNK:
Addresses #80

Builds on top of https://github.com/grpc/grpc-dotnet/pull/103 and adds an invoker factory.

The invoker factory provides a way for developers to inject custom logic to control how the service method is invoked. The default logic is still to use the method name to get a method via `typeof(Service).GetMethod(Method.Name)` and wrap it in a delegate.

Update: Made invoker factory internal. It can be made public in the future if there is user demand for it.

### DIFF
--- a/src/Grpc.AspNetCore.Server/GrpcBindingOptions.cs
+++ b/src/Grpc.AspNetCore.Server/GrpcBindingOptions.cs
@@ -16,24 +16,23 @@
 
 #endregion
 
+using System;
+using Grpc.AspNetCore.Server.Internal;
+using Grpc.Core;
+
 namespace Grpc.AspNetCore.Server
 {
     /// <summary>
-    /// A <see cref="TGrpcService"/> activator abstraction.
+    /// Options used to configure the binding of a gRPC service.
     /// </summary>
-    /// <typeparam name="TGrpcService">The service type.</typeparam>
-    public interface IGrpcServiceActivator<TGrpcService> where TGrpcService : class
+    public class GrpcBindingOptions<TService>
     {
         /// <summary>
-        /// Creates a service.
+        /// The action invoked to get service metadata via <see cref="ServiceBinderBase"/>.
         /// </summary>
-        /// <returns>The created service.</returns>
-        TGrpcService Create();
+        public Action<ServiceBinderBase, TService> BindAction { get; set; }
 
-        /// <summary>
-        /// Releases the specified service.
-        /// </summary>
-        /// <param name="hub">The service to release.</param>
-        void Release(TGrpcService service);
+        // Currently internal. It is set in tests via InternalVisibleTo. Can be made public if there is demand for it
+        internal IGrpcMethodInvokerFactory<TService> InvokerFactory { get; set; }
     }
 }
--- a/src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
+++ b/src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
@@ -17,19 +17,44 @@
 #endregion
 
 using System;
+using System.Reflection;
 using Grpc.AspNetCore.Server;
 using Grpc.AspNetCore.Server.Internal;
 using Grpc.Core;
-using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Builder
 {
+    /// <summary>
+    /// Provides extension methods for <see cref="IEndpointRouteBuilder"/> to add gRPC service endpoints.
+    /// </summary>
     public static class GrpcEndpointRouteBuilderExtensions
     {
+        /// <summary>
+        /// Maps incoming requests to the specified <see cref="TService"/> type.
+        /// </summary>
+        /// <typeparam name="TService">The service type to map requests to.</typeparam>
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder"/> to add the route to.</param>
+        /// <returns>An <see cref="IEndpointConventionBuilder"/> for endpoints associated with the service.</returns>
         public static IEndpointConventionBuilder MapGrpcService<TService>(this IEndpointRouteBuilder builder) where TService : class
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            return builder.MapGrpcService<TService>(configureOptions: null);
+        }
+
+        /// <summary>
+        /// Maps incoming requests to the specified <see cref="TService"/> type.
+        /// </summary>
+        /// <typeparam name="TService">The service type to map requests to.</typeparam>
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder"/> to add the route to.</param>
+        /// <param name="configureOptions">A callback to configure binding options.</param>
+        /// <returns>An <see cref="IEndpointConventionBuilder"/> for endpoints associated with the service.</returns>
+        public static IEndpointConventionBuilder MapGrpcService<TService>(this IEndpointRouteBuilder builder, Action<GrpcBindingOptions<TService>> configureOptions) where TService : class
         {
             if (builder == null)
             {
@@ -38,42 +63,47 @@ namespace Microsoft.AspNetCore.Builder
 
             ValidateServicesRegistered(builder.ServiceProvider);
 
-            var service = typeof(TService);
+            var options = new GrpcBindingOptions<TService>();
+            options.BindAction = ReflectionBind<TService>;
+            options.InvokerFactory = new ReflectionMethodInvokerFactory<TService>();
+
+            configureOptions?.Invoke(options);
+
+            var callHandlerFactory = builder.ServiceProvider.GetRequiredService<ServerCallHandlerFactory<TService>>();
+            var serviceBinder = new GrpcServiceBinder<TService>(builder, callHandlerFactory, options.InvokerFactory);
 
             try
             {
-                // TService is an implementation of the gRPC service. It ultimately derives from Foo.TServiceBase base class.
-                // We need to access the static BindService method on Foo which implicitly derives from Object.
-                var baseType = service.BaseType;
-                while (baseType?.BaseType?.BaseType != null)
-                {
-                    baseType = baseType.BaseType;
-                }
-
-                // We need to call Foo.BindService from the declaring type.
-                var declaringType = baseType?.DeclaringType;
-
-                // The method we want to call is public static void BindService(ServiceBinderBase, BaseType)
-                var bindService = declaringType?.GetMethod("BindService", new[] { typeof(ServiceBinderBase), baseType });
-
-                if (bindService == null)
-                {
-                    throw new InvalidOperationException($"Cannot locate BindService(ServiceBinderBase, ServiceBase) method for the current service type: {service.FullName}. The type must be an implementation of a gRPC service.");
-                }
-
-                var callHandlerFactory = builder.ServiceProvider.GetRequiredService<ServerCallHandlerFactory<TService>>();
-
-                var serviceBinder = new GrpcServiceBinder<TService>(builder, callHandlerFactory);
-
-                // Invoke
-                bindService.Invoke(null, new object[] { serviceBinder, null });
-
-                return new CompositeEndpointConventionBuilder(serviceBinder.EndpointConventionBuilders);
+                options.BindAction(serviceBinder, null);
             }
             catch (Exception ex)
             {
-                throw new InvalidOperationException($"Unable to map gRPC service '{service.Name}'.", ex);
+                throw new InvalidOperationException($"Error binding gRPC service '{typeof(TService).Name}'.", ex);
             }
+
+            return new CompositeEndpointConventionBuilder(serviceBinder.EndpointConventionBuilders);
+        }
+
+        private static void ReflectionBind<TService>(ServiceBinderBase binder, TService service)
+        {
+            var serviceType = typeof(TService);
+
+            // TService is an implementation of the gRPC service. It ultimately derives from Foo.TServiceBase base class.
+            // We need to access the static BindService method on Foo which implicitly derives from Object.
+            var baseType = serviceType.BaseType;
+
+            // We need to call Foo.BindService from the declaring type.
+            var declaringType = baseType?.DeclaringType;
+
+            // The method we want to call is public static void BindService(ServiceBinderBase, BaseType)
+            var bindService = declaringType?.GetMethod("BindService", BindingFlags.Public | BindingFlags.Static, null, new[] { typeof(ServiceBinderBase), baseType }, Array.Empty<ParameterModifier>());
+
+            if (bindService == null)
+            {
+                throw new InvalidOperationException($"Cannot locate BindService(ServiceBinderBase, ServiceBase) method for the current service type: {serviceType.FullName}.");
+            }
+
+            bindService.Invoke(null, new object[] { binder, service });
         }
 
         private static void ValidateServicesRegistered(IServiceProvider serviceProvider)
--- a/src/Grpc.AspNetCore.Server/Internal/ServerMethods.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/ServerMethods.cs
@@ -1,0 +1,34 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System.Threading.Tasks;
+using Grpc.Core;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    // Open delegate (the first argument is the TService instance) versions of the service call types.
+    // Needed because methods are executed with a new service instance each request.
+
+    internal delegate Task<TResponse> UnaryServerMethod<TService, TRequest, TResponse>(TService service, TRequest request, ServerCallContext serverCallContext);
+
+    internal delegate Task<TResponse> ClientStreamingServerMethod<TService, TRequest, TResponse>(TService service, IAsyncStreamReader<TRequest> stream, ServerCallContext serverCallContext);
+
+    internal delegate Task ServerStreamingServerMethod<TService, TRequest, TResponse>(TService service, TRequest request, IServerStreamWriter<TResponse> stream, ServerCallContext serverCallContext);
+
+    internal delegate Task DuplexStreamingServerMethod<TService, TRequest, TResponse>(TService service, IAsyncStreamReader<TRequest> input, IServerStreamWriter<TResponse> output, ServerCallContext serverCallContext);
+}
--- a/src/Grpc.AspNetCore.Server/Properties/AssemblyInfo.cs
+++ b/src/Grpc.AspNetCore.Server/Properties/AssemblyInfo.cs
@@ -35,3 +35,10 @@ using System.Runtime.CompilerServices;
     "0442bb8e12768722de0b0cb1b15e955b32a11352740ee59f2c94c48edc8e177d1052536b8ac651bce11ce5da3a" +
     "27fc95aff3dc604a6971417453f9483c7b5e836756d5b271bf8f2403fe186e31956148c03d804487cf642f8cc0" +
     "71394ee9672dfe5b55ea0f95dfd5a7f77d22c962ccf51320d3")]
+
+// For Moq. This assembly needs access to internal types via InternalVisibleTo to be able to mock them
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602" +
+    "000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02ba" +
+    "a56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667" +
+    "bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6" +
+    "a7d3113e92484cf7045cc7")] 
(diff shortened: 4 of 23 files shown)


test/FunctionalTests/Infrastructure/DynamicGrpcService.cs
```diff
@@ -0,0 +1,116 @@
...
+        private readonly DynamicEndpointDataSource _endpointDataSource;
+        private readonly IServiceProvider _serviceProvider;
+
+        public DynamicGrpcServiceRegister(DynamicEndpointDataSource endpointDataSource, IServiceProvider serviceProvider)
+        {
+            _endpointDataSource = endpointDataSource;
+            _serviceProvider = serviceProvider;
+        }
+
+        public string AddUnaryMethod<TRequest, TResponse, TService>(UnaryServerMethod<TRequest, TResponse> callHandler)
```
JamesNK: Can add other service types if/when we're happy with this.

src/Grpc.AspNetCore.Server/ServerMethods.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    // We're using an open delegate (the first argument is the TService instance) to represent calls here since the instance is create per request.
+
+    public delegate Task<TResponse> UnaryServerMethod<TRequest, TResponse, TService>(TService service, TRequest request, ServerCallContext serverCallContext);
+
+    public delegate Task<TResponse> ClientStreamingServerMethod<TRequest, TResponse, TService>(TService service, IAsyncStreamReader<TRequest> stream, ServerCallContext serverCallContext);
+
+    public delegate Task ServerStreamingServerMethod<TRequest, TResponse, TService>(TService service, TRequest request, IServerStreamWriter<TResponse> stream, ServerCallContext serverCallContext);
+
+    public delegate Task DuplexStreamingServerMethod<TRequest, TResponse, TService>(TService service, IAsyncStreamReader<TRequest> input, IServerStreamWriter<TResponse> output, ServerCallContext serverCallContext);
```
JamesNK: The open delegates are now public because they are created by the invoker factory. Thoughts?

davidfowl: If we make them public I'd prefer for them to live in Grpc.Core

JamesNK: Maybe. Nothing in Grpc.Core.Api would use them unless codegen and ServiceBinderBase was updated.

Grpc.Core.* is locked down for 1.19 so they have to live here for the first preview.

test/FunctionalTests/UnaryMethodTests.cs
```diff
@@ -196,5 +215,49 @@ public async Task SendHeadersTwice_ThrowsException()
             var responseMessage = MessageHelpers.AssertReadMessage<HelloReply>(await response.Content.ReadAsByteArrayAsync().DefaultTimeout());
             Assert.AreEqual("Exception validated", responseMessage.Message);
         }
+
+        [Test]
+        public async Task ValidRequest_ReturnContextInfoInTrailers()
+        {
+            static Task<Empty> ReturnContextInfoInTrailers(Empty request, ServerCallContext context)
```
JamesNK: spicy

src/Grpc.AspNetCore.Server/GrpcBindingOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+#endregion
+
+using System;
+using Grpc.Core;
+
+namespace Grpc.AspNetCore.Server
+{
+    /// <summary>
+    /// Options used to configure the binding of HTTP/2 to a .NET service.
```
davidfowl: >     /// Options used to configure the binding of HTTP/2 to a .NET service.

    /// Options used to configure the binding a GRPC service.

src/Grpc.AspNetCore.Server/ServerMethods.cs
rynowak: Why does the invoker factory need to be public?

test/FunctionalTests/UnaryMethodTests.cs
rynowak: That is real spicy.

src/Grpc.AspNetCore.Server/ServerMethods.cs
JamesNK: I could be internal and be accessed via InternalVisibleTo. It could be useful for other devs who don't want to be died to reflection to get the call handler method.

davidfowl: We should only make things like that public when people ask. It’s looking pretty niche right now

src/Grpc.AspNetCore.Server/Properties/AssemblyInfo.cs
```diff
@@ -35,3 +35,9 @@
     "0442bb8e12768722de0b0cb1b15e955b32a11352740ee59f2c94c48edc8e177d1052536b8ac651bce11ce5da3a" +
     "27fc95aff3dc604a6971417453f9483c7b5e836756d5b271bf8f2403fe186e31956148c03d804487cf642f8cc0" +
     "71394ee9672dfe5b55ea0f95dfd5a7f77d22c962ccf51320d3")]
+
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602" +
```
JamesNK: So Moq will work with InternalVisibleTo types.

src/Grpc.AspNetCore.Server/GrpcBindingOptions.cs
```diff
@@ -31,9 +32,7 @@ public class GrpcBindingOptions<TService>
         /// </summary>
         public Action<ServiceBinderBase, TService> BindAction { get; set; }
```
davidfowl: What’s this for? When would you override it?

JamesNK: gRPC will code gen a method that we use to discover methods on the service:

```cs
    /// <summary>Register service method with a service binder with or without implementation. Useful when customizing the  service binding logic.
    /// Note: this method is part of an experimental API that can change or be removed without any prior notice.</summary>
    /// <param name="serviceBinder">Service methods will be bound by calling <c>AddMethod</c> on this object.</param>
    /// <param name="serviceImpl">An object implementing the server-side handling logic.</param>
    public static void BindService(grpc::ServiceBinderBase serviceBinder, GreeterBase serviceImpl)
    {
      serviceBinder.AddMethod(__Method_SayHello, serviceImpl == null ? null : new grpc::UnaryServerMethod<global::Greet.HelloRequest, global::Greet.HelloReply>(serviceImpl.SayHello));
      serviceBinder.AddMethod(__Method_SayHellos, serviceImpl == null ? null : new grpc::ServerStreamingServerMethod<global::Greet.HelloRequest, global::Greet.HelloReply>(serviceImpl.SayHellos));
    }
```

Gprc.Tools always places this in the same location, but if another tool is used to generate the services - perhaps you want to use something other than Google.Protobuf for serialization - then the method could be in a different location. This delegate allows you to specify the method explicitly.

```cs
    builder.MapGrpcService<GreeterService>(options => options.BindAction = Greet.Greeter.BindService);
```

davidfowl: Oh no I’m aware I’m mostly asking if we think this is useful to expose in this first class way. Do we think people should be doing this? Is this something we’re going to generate in the future?

JamesNK: I see explicitly setting the bind action as an escape hatch for when  `MapGrpcService<T>()` doesn't work. If we don't provide an escape hatch then any code gen-ed types that don't have the right structure (and the structure is pretty weird tbh) will be locked out from being used with `Grpc.AspNetCore`.

src/Grpc.AspNetCore.Server/Internal/ServerMethods.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using System.Threading.Tasks;
+using Grpc.Core;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    // Open delegate (the first argument is the TService instance) versions of the service call types
+    // Useful because methods are executed with an instance create per request.
+
+    internal delegate Task<TResponse> UnaryServerMethod<TRequest, TResponse, TService>(TService service, TRequest request, ServerCallContext serverCallContext);
```
jtattermusch: possible refactor: generic parameter order is different `<TRequest, TResponse, TService>` from the argument order. Should we align these?

src/Grpc.AspNetCore.Server/Internal/IGrpcServiceInvokerFactory.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    /// <summary>
+    /// An interface for creating delegates used to invoke service methods on .NET types.
+    /// </summary>
+    /// <typeparam name="TService">The service type.</typeparam>
+    internal interface IGrpcServiceInvokerFactory<TService>
+    {
+        UnaryServerMethod<TRequest, TResponse, TService> CreateUnaryInvoker<TRequest, TResponse>(Method<TRequest, TResponse> method);
```
jtattermusch: While in the default MapGrpcService() use case the TService is actually the type of the generated service class, 
if one uses the MapGrpcService() overload with binder options and custom IGrpcServiceInvokerFactory,  it is no longer clear that TService denotes a service class. It could be something else, depending on given IGrpcServiceInvokerFactory implementation?

src/Grpc.AspNetCore.Server/ServerMethods.cs
rynowak: I'd strongly advise against making implementation details like this public. If users ask for it, consider making a *different thing* for their scenario and making that public instead.

test/FunctionalTests/Grpc.AspNetCore.FunctionalTests.csproj
```diff
@@ -5,7 +5,7 @@
 
     <IsPackable>false</IsPackable>
 
-    <LangVersion>latest</LangVersion>
+    <LangVersion>8.0</LangVersion>
```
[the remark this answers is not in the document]
JamesNK: For some reason "latest" is defaulting to 7.3 (?) of csharp and not 8.0. This might be a preview thing. I think this is required for now and can change to latest in the future.

8.0 is required for static local methods.

test/FunctionalTests/Infrastructure/DynamicGrpcService.cs
```diff
@@ -0,0 +1,116 @@
...
+            where TService : class
+            where TRequest : class, IMessage, new()
+            where TResponse : class, IMessage, new()
+        {
+            var method = CreateMethod<TRequest, TResponse, TService>(MethodType.Unary, Guid.NewGuid().ToString());
```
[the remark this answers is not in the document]
JamesNK: I didn't like having a set of 4 "dynamic" methods that were changed on a test-by-test basis because it runs into trouble when tests execute in parallel.

If test A sets the unary method to Foo, and test B sets the unary method to Bar, and they run at the same time then a race condition occurs.

Also this pattern lets different message types for different tests.

src/Grpc.AspNetCore.Server/Internal/IGrpcServiceInvokerFactory.cs
JamesNK: We need TService because that is part of the delegate signature, and it is the type that will be passed to the delegate when it is executed.

But TService doesn't necessarily have to be used by the delegate. In the inline tests it is ignored.

Does that make sense?

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
```diff
@@ -20,60 +20,94 @@
...
+        }
+
+        /// <summary>
+        /// Maps incoming requests to the specified <see cref="TService"/> type.
+        /// </summary>
+        /// <typeparam name="TService">The service type to map requests to.</typeparam>
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder"/> to add the route to.</param>
+        /// <param name="configureOptions">A callback to configure binding options.</param>
+        /// <returns>An <see cref="IEndpointConventionBuilder"/> for endpoints associated with the service.</returns>
+        public static IEndpointConventionBuilder MapGrpcService<TService>(this IEndpointRouteBuilder builder, Action<GrpcBindingOptions<TService>> configureOptions) where TService : class
```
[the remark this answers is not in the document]
JamesNK: Note that invoker factory is internal for now.

I think these are scenarios to think about but not solve in this PR. I'll create issues for them to discuss in more detail.

> * define a custom handler for given gRPC method (without necessarily using a generated class as a handler).

This is an interesting thought. I created an issue for it - https://github.com/grpc/grpc-dotnet/issues/111

> * define a fallback gRPC method handler that would enable implementing e.g. "Unimplemented method" handler (one of the interop test cases).

I've created an issue to cover the basics of setting the status - https://github.com/grpc/grpc-dotnet/issues/109

Are fallback handlers a thing in other frameworks? We could solve that scenaro using an [adhoc handler](https://github.com/grpc/grpc-dotnet/issues/111) with a wildcard route: `routes.MapGrpcUnaryMethod<HelloRequest, HelloReply>("Greet.Greeter/{any*}", handler);`

Another option could be an attribute that a user could place on a method in the service implementation that says that this method should be used as a fallback.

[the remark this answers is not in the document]
JamesNK: > * allow defining a "generic" grpc service that allows handling RPCs without access to raw message payloads, without necessarily parsing them.

Something like:

```cs
routes.MapGrpcUnaryMethod("Greet.Greeter/{any*}", (payload, context) =>
{
    var messageData = new byte { 0, 1 };
    return messageData;
});
```

???

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
```diff
@@ -20,60 +20,94 @@
...
-                }
+        private static void ReflectionBind<TService>(ServiceBinderBase binder, TService _)
+        {
+            var service = typeof(TService);
 
-                var callHandlerFactory = builder.ServiceProvider.GetRequiredService<ServerCallHandlerFactory<TService>>();
+            // TService is an implementation of the gRPC service. It ultimately derives from Foo.TServiceBase base class.
+            // We need to access the static BindService method on Foo which implicitly derives from Object.
+            var baseType = service.BaseType;
+            while (baseType?.BaseType?.BaseType != null)
```
halter73: I know this logic isn't new in this PR, but it cannot be correct to jump over two base types at a time. It looks like this code is trying to get the type that derives from System.Object (e.g. Greet.Greeter+GreeterBase) instead of the object type itself. The problem is that this only works if there is an even number of base types. So it works for:

```
GreeterService -> Greet.Greeter+GreeterBase -> System.Object -> null
```

But it breaks for:

```
GreeterService -> IntermediateDerivedType -> Greet.Greeter+GreeterBase -> System.Object -> null
```

In the second case, baseType would be System.Object instead of Greet.Greeter+GreeterBase as expected when the loop ends.

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
```diff
@@ -20,60 +20,94 @@
...
+            // The method we want to call is public static void BindService(ServiceBinderBase, BaseType)
+            var bindService = declaringType?.GetMethod("BindService", new[] { typeof(ServiceBinderBase), baseType });
 
-                return new CompositeEndpointConventionBuilder(serviceBinder.EndpointConventionBuilders);
-            }
-            catch (Exception ex)
+            if (bindService == null)
             {
-                throw new InvalidOperationException($"Unable to map gRPC service '{service.Name}'.", ex);
+                throw new InvalidOperationException($"Cannot locate BindService(ServiceBinderBase, ServiceBase) method for the current service type: {service.FullName}.");
```
halter73: Why remove "The type must be an implementation of a gRPC service."? If ReflectionBind is being used, TService does need to implement a generated gRPC service type, right?

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
```diff
@@ -20,60 +20,94 @@
...
-                // We need to access the static BindService method on Foo which implicitly derives from Object.
-                var baseType = service.BaseType;
-                while (baseType?.BaseType?.BaseType != null)
-                {
-                    baseType = baseType.BaseType;
-                }
-
-                // We need to call Foo.BindService from the declaring type.
-                var declaringType = baseType?.DeclaringType;
+                options.BindAction(serviceBinder, null);
```
halter73: Is BindAction ever called with a non-null TService? If not, why have the second argument?

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
```diff
@@ -20,60 +20,94 @@
...
-            }
-            catch (Exception ex)
+            if (bindService == null)
             {
-                throw new InvalidOperationException($"Unable to map gRPC service '{service.Name}'.", ex);
+                throw new InvalidOperationException($"Cannot locate BindService(ServiceBinderBase, ServiceBase) method for the current service type: {service.FullName}.");
             }
+
+            // Invoke
+            bindService.Invoke(null, new object[] { binder, null });
```
halter73: You might as well pass the `TService _` argument through here instead of null, right? In practice it seems it will always be null anyway.

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
```diff
@@ -20,60 +20,94 @@
...
+            }
 
-                var serviceBinder = new GrpcServiceBinder<TService>(builder, callHandlerFactory);
+            // We need to call Foo.BindService from the declaring type.
+            var declaringType = baseType?.DeclaringType;
 
-                // Invoke
-                bindService.Invoke(null, new object[] { serviceBinder, null });
+            // The method we want to call is public static void BindService(ServiceBinderBase, BaseType)
+            var bindService = declaringType?.GetMethod("BindService", new[] { typeof(ServiceBinderBase), baseType });
```
halter73: Should we use custom BindingFlags to exclude non-static methods?

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
JamesNK: Because of different code gen output, what we support as a gRPC service is different from what your tool output as a gRPC service. This error message is more accurate to what the problem is.

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
JamesNK: Because code gen outputs a second argument that we don't use. We could make BindAction only take ServiceBinderBase but user code would go from:

```cs
builder.MapGrpcService<GreeterService>(options => options.BindAction = Greet.Greeter.BindService);
```

To:

```cs
builder.MapGrpcService<GreeterService>(options => options.BindAction = (binder) => Greet.Greeter.BindService(binder, null));
```

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
JamesNK: I guess, although we'll pass in the ServiceBinderBase implementation and we'll never do anything with the service.

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
JamesNK: 🤷‍♂️ 

I think it has been copy and pasted through the ages. I'll check it out

JamesNK: As far as I can tell it never did anything.

src/Grpc.AspNetCore.Server/Internal/IGrpcServiceInvokerFactory.cs
jtattermusch: Sure, I was just wondering if it should really be called TService when it can in fact represent other things too.

test/FunctionalTests/Infrastructure/DynamicGrpcService.cs
```diff
@@ -0,0 +1,116 @@
...
+using Microsoft.AspNetCore.Builder.Internal;
+using Microsoft.AspNetCore.Routing;
+using Moq;
+
+namespace Grpc.AspNetCore.FunctionalTests.Infrastructure
+{
+    /// <summary>
+    /// Used by tests to add new service methods.
+    /// </summary>
+    public class DynamicGrpcServiceRegister
```
jtattermusch: nit: should this be DynamicGrpcServiceRegistry?

test/FunctionalTests/Infrastructure/DynamicGrpcService.cs
```diff
@@ -0,0 +1,116 @@
+﻿#region Copyright notice and license
```
jtattermusch: nit: make the file name match the class name?

src/Grpc.AspNetCore.Server/GrpcBindingOptions.cs
```diff
@@ -0,0 +1,38 @@
...
+using System;
+using Grpc.AspNetCore.Server.Internal;
+using Grpc.Core;
+
+namespace Grpc.AspNetCore.Server
+{
+    /// <summary>
+    /// Options used to configure the binding of a gRPC service.
+    /// </summary>
+    public class GrpcBindingOptions<TService>
```
jtattermusch: I think I'm happy with the GrpcBindingOptions for now, if the assumption is that we will revisit the API for mapping/binding grpc services and methods in the future before we treat this API as stable (I think we should do an API review at some point as proposed in https://github.com/grpc/grpc-dotnet/issues/111#issuecomment-465504107).

src/Grpc.AspNetCore.Server/Internal/IGrpcServiceInvokerFactory.cs
JamesNK: Maybe. IMO we shouldn't think too hard about it because it is internal, and only in unit tests could it mean something else.

test/FunctionalTests/Infrastructure/DynamicGrpcServiceRegistry.cs
```diff
@@ -0,0 +1,116 @@
...
+        private readonly DynamicEndpointDataSource _endpointDataSource;
+        private readonly IServiceProvider _serviceProvider;
+
+        public DynamicGrpcServiceRegistry(DynamicEndpointDataSource endpointDataSource, IServiceProvider serviceProvider)
+        {
+            _endpointDataSource = endpointDataSource;
+            _serviceProvider = serviceProvider;
+        }
+
+        public string AddUnaryMethod<TService, TRequest, TResponse>(UnaryServerMethod<TRequest, TResponse> callHandler)
```
JunTaoLuo: Also add support for streaming methods?

JamesNK: Ah yes. Will do in a follow up.

PR #106, opened 2019-02-19
Add functional tests for using HttpContext

JamesNK:


testassets/FunctionalTestsWebsite/Greeter.cs
```diff
@@ -39,6 +42,22 @@ public override Task<HelloReply> SayHello(HelloRequest request, ServerCallContex
         return Task.FromResult(new HelloReply { Message = "Hello " + request.Name });
     }
 
+    public override Task<HelloReply> SayHelloWithHttpContextAccessor(HelloRequest request, ServerCallContext context)
+    {
+        var httpContext = _httpContextAccessor.HttpContext;
```
davidfowl: I hope this isn't the only way to get the HttpContext

JamesNK: https://github.com/grpc/grpc-dotnet/pull/106/files#diff-e2b95dbc2cf449787b017fdd16ca0d35R55 ???

davidfowl: Phew

davidfowl: The SignalR similarities are uncanny

PR #112, opened 2019-02-20
Required changes for 3.0.0 preview3

JunTaoLuo:
I noticed a few changes in preview3 and updated our code accordingly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,5 @@ branches:
   only:
     - master
 script:
-  - ./build/get-dotnet.sh
+  - source build/get-dotnet.sh
   - ./build_and_test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ branches:
   only:
     - master
 install:
-  - curl -o dotnet-sdk.tar.gz -sSL https://download.visualstudio.microsoft.com/download/pr/efa6dde9-a5ee-4322-b13c-a2a02d3980f0/dad445eba341c1d806bae5c8afb47015/dotnet-sdk-3.0.100-preview-010184-linux-x64.tar.gz
+  - curl -o dotnet-sdk.tar.gz -sSL https://dotnetcli.blob.core.windows.net/dotnet/Sdk/release/3.0.1xx/dotnet-sdk-latest-linux-x64.tar.gz
   - mkdir -p $PWD/dotnet
   - tar zxf dotnet-sdk.tar.gz -C $PWD/dotnet
   - export PATH="$PWD/dotnet:$PATH"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,6 @@ mono: none
 branches:
   only:
     - master
-install:
-  - curl -o dotnet-sdk.tar.gz -sSL https://dotnetcli.blob.core.windows.net/dotnet/Sdk/release/3.0.1xx/dotnet-sdk-latest-linux-x64.tar.gz
-  - mkdir -p $PWD/dotnet
-  - tar zxf dotnet-sdk.tar.gz -C $PWD/dotnet
-  - export PATH="$PWD/dotnet:$PATH"
 script:
+  - ./build/get-dotnet.sh
   - ./build_and_test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,6 @@ branches:
   only:
     - master
 script:
-  - source build/get-dotnet.sh
+  - ./build/get-dotnet.sh
+  - export PATH="~/.dotnet/:$PATH"
   - ./build_and_test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ mono: none
 branches:
   only:
     - master
-script:
+install:
   - ./build/get-dotnet.sh
   - export PATH="~/.dotnet/:$PATH"
+script:
   - ./build_and_test.sh
--- a/examples/Server/Startup.cs
+++ b/examples/Server/Startup.cs
@@ -16,9 +16,7 @@
 
 #endregion
 
-using Grpc.AspNetCore;
 using Microsoft.AspNetCore.Builder;
-using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace GRPCServer
@@ -34,7 +32,7 @@ namespace GRPCServer
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
+        public void Configure(IApplicationBuilder app)
         {
             app.UseRouting(routes =>
             {
--- a/global.json
+++ b/global.json
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "3.0.100-preview-010184"
+    "version": "3.0.100-preview3-010313"
   }
 }
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
@@ -199,7 +199,7 @@ namespace Grpc.AspNetCore.Server.Internal
                 }
             }
 
-            return HttpContext.Response.Body.FlushAsync();
+            return HttpContext.Response.StartAsync();
         }
 
         public void Initialize()
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
@@ -199,7 +199,7 @@ namespace Grpc.AspNetCore.Server.Internal
                 }
             }
 
-            return HttpContext.Response.StartAsync();
+            return HttpContext.Response.Body.FlushAsync();
         }
 
         public void Initialize()
--- a/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
@@ -56,10 +56,14 @@ namespace Grpc.AspNetCore.Server.Internal
             }
 
             // Must call StartAsync before the first pipeWriter.GetSpan() in WriteHeader
-            var startAsyncTask = serverCallContext.HttpContext.Response.StartAsync();
-            if (!startAsyncTask.IsCompletedSuccessfully)
+            var response = serverCallContext.HttpContext.Response;
+            if (!response.HasStarted)
             {
-                return pipeWriter.WriteMessageCoreAsyncAwaited(messageData, serverCallContext, flush, startAsyncTask);
+                var startAsyncTask = response.StartAsync();
+                if (!startAsyncTask.IsCompletedSuccessfully)
+                {
+                    return pipeWriter.WriteMessageCoreAsyncAwaited(messageData, serverCallContext, flush, startAsyncTask);
+                }
             }
 
             return pipeWriter.WriteMessageCoreAsync(messageData, serverCallContext, flush);
--- a/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
@@ -55,19 +55,23 @@ namespace Grpc.AspNetCore.Server.Internal
                 return Task.FromException(new RpcException(SendingMessageExceedsLimitStatus));
             }
 
-            if (!serverCallContext.HttpContext.Response.HasStarted)
+            // Must call StartAsync before the first pipeWriter.GetSpan() in WriteHeader
+            var response = serverCallContext.HttpContext.Response;
+            if (!response.HasStarted)
             {
-                return pipeWriter.WriteMessageCoreInitialAsync(messageData, serverCallContext, flush);
+                var startAsyncTask = response.StartAsync();
+                if (!startAsyncTask.IsCompletedSuccessfully)
+                {
+                    return pipeWriter.WriteMessageCoreAsyncAwaited(messageData, serverCallContext, flush, startAsyncTask);
+                }
             }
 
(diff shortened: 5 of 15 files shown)


examples/Server/Startup.cs
```diff
@@ -34,7 +32,7 @@ public void ConfigureServices(IServiceCollection services)
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
```
JunTaoLuo: `IHostingEnvironment` will be deprecated in favour of `IWebHostEnvironment` but https://github.com/aspnet/AspNetCore/pull/7725 isn't merged yet. I didn't want to wait until the changes go in and a new SDK build so I opted to just remove the argument since we weren't using it anyway.

global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "3.0.100-preview-010184"
+    "version": "3.0.100-preview3-010313"
```
JunTaoLuo: Ingesting daily aspnetcore builds is going to be a pain. I need to figure out a better way to do this but for now, download from https://github.com/dotnet/core-sdk#installers-and-binaries. I need to find a way to get a permanent link to a specific version since https://dotnetcli.blob.core.windows.net/dotnet/Sdk/release/3.0.1xx/dotnet-sdk-latest-win-x64.zip will change frequently.

test/Grpc.AspNetCore.Server.Tests/GrpcEndpointRouteBuilderExtensionsTests.cs
```diff
@@ -78,11 +78,11 @@ public void MapGrpcService_CanBind_CreatesEndpoints()
 
             var routeEndpoint1 = (RouteEndpoint)endpoints[0];
             Assert.AreEqual("/Greet.Greeter/SayHello", routeEndpoint1.RoutePattern.RawText);
-            Assert.AreEqual("POST", ((IHttpMethodMetadata)routeEndpoint1.Metadata.Single()).HttpMethods.Single());
+            Assert.AreEqual("POST", ((IHttpMethodMetadata)routeEndpoint1.Metadata.Single(m => m is HttpMethodMetadata)).HttpMethods.Single());
```
JunTaoLuo: @JamesNK Looks like this is due to some changes in endpoint routing. https://github.com/aspnet/AspNetCore/pull/6911 maybe?

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -55,6 +55,24 @@ public static Task WriteMessageAsync(this PipeWriter pipeWriter, byte[] messageD
...
+                return pipeWriter.WriteMessageCoreInitialAsync(messageData, serverCallContext, flush);
+            }
+
+            return pipeWriter.WriteMessageCoreAsync(messageData, serverCallContext, flush);
+        }
+
+        private static async Task WriteMessageCoreInitialAsync(this PipeWriter pipeWriter, byte[] messageData, HttpContextServerCallContext serverCallContext, bool flush)
+        {
+            // Must call StartAsync before the first pipeWriter.GetSpan() in WriteHeader
+            await serverCallContext.HttpContext.Response.StartAsync();
```
JunTaoLuo: This is a new requirement in preview3 in order to use PipeWriter.GetSpan/GetMemory.

global.json
JunTaoLuo: @natemcmaster I don't have access to the dotnetcli blob store in the storage explorer so I can't find a permanent link. Do you have any other ideas on how to ingest daily builds more easily?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -199,7 +199,7 @@ protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
                 }
             }
 
-            return HttpContext.Response.Body.FlushAsync();
+            return HttpContext.Response.StartAsync();
```
davidfowl: This isn't right. You want to keep calling FlushAsync

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
davidfowl: This should be optimized to avoid the state machine in the common case of completing synchronously

JunTaoLuo: That's what the HasStarted check is for. IN the common case, we call WriteMessageCoreAsync instead of WriteMessageCoreInitialAsync

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: Hmm does StartAsync not flush the headers?

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
davidfowl: Except, that's not the common case. The common case is that HasStarted isn't true and StartAsync completes synchronously. See https://github.com/aspnet/AspNetCore/blob/master/src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs#L64 for a more complicated example

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
davidfowl: No, it does not.

JamesNK: Is Body.FlushAsync right tho? Everywhere else we're all about BodyPipe.

`BodyPipe.FlushAsync()`? (or whatever its pipe equivalent is)

test/Grpc.AspNetCore.Server.Tests/GrpcEndpointRouteBuilderExtensionsTests.cs
JamesNK: This is better:

```suggestion
            Assert.AreEqual("POST", routeEndpoint1.Metadata.GetMetadata<IHttpMethodMetadata>().HttpMethods.Single());
```

Out of interest, what is the other metadata?

test/Grpc.AspNetCore.Server.Tests/GrpcEndpointRouteBuilderExtensionsTests.cs
```diff
@@ -78,11 +78,11 @@ public void MapGrpcService_CanBind_CreatesEndpoints()
...
 
             var routeEndpoint2 = (RouteEndpoint)endpoints[1];
             Assert.AreEqual("/Greet.Greeter/SayHellos", routeEndpoint2.RoutePattern.RawText);
-            Assert.AreEqual("POST", ((IHttpMethodMetadata)routeEndpoint2.Metadata.Single()).HttpMethods.Single());
+            Assert.AreEqual("POST", ((IHttpMethodMetadata)routeEndpoint2.Metadata.Single(m => m is HttpMethodMetadata)).HttpMethods.Single());
```
JamesNK:
```suggestion
            Assert.AreEqual("POST", routeEndpoint2.Metadata.GetMetadata<IHttpMethodMetadata>().HttpMethods.Single());
```

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
davidfowl: I dunno if we implemented this. @jkotalik if you do a flush without writing any body, do we flush the headers? Maybe we should add something first class on the HttpResponse...

global.json
natemcmaster: Don't download from dotnetcli.blob.core.windows.net directly. This location is an implementation detail. I would recommend using `dotnet-install.sh`/`.ps1` instead. See https://docs.microsoft.com/dotnet/core/tools/dotnet-install-script for more details.

Example:
```ps1
$globalJson = Get-Content -Raw ./global.json | ConvertFrom-Json
Invoke-WebRequest https://dot.net/v1/dotnet-install.ps1 -out dotnet-install.ps1
./dotnet-install.ps1 -version $globalJson.sdk.version
```

.travis.yml
```diff
@@ -9,7 +9,7 @@ branches:
   only:
     - master
 install:
-  - curl -o dotnet-sdk.tar.gz -sSL https://download.visualstudio.microsoft.com/download/pr/efa6dde9-a5ee-4322-b13c-a2a02d3980f0/dad445eba341c1d806bae5c8afb47015/dotnet-sdk-3.0.100-preview-010184-linux-x64.tar.gz
+  - curl -o dotnet-sdk.tar.gz -sSL https://dotnetcli.blob.core.windows.net/dotnet/Sdk/release/3.0.1xx/dotnet-sdk-latest-linux-x64.tar.gz
```
jtattermusch: we should use a fixed version here, not "latest"  to prevent sudden breakages. Or is this just temporary because the actual preview3 archive doesn't exist yet? (if so, please add a TODO).

davidfowl: Agreed

test/Grpc.AspNetCore.Server.Tests/GrpcEndpointRouteBuilderExtensionsTests.cs
JunTaoLuo:
```
System.Runtime.CompilerServices.AsyncStateMachineAttribute
System.Diagnostics.DebuggerStepThroughAttribute
```

JunTaoLuo:
cc @halter73

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: At one point I thought StartAsync would actually write the headers not just commit them. I could be mis-remembering though. Regardless the FlushAsync API should still work? cc @halter73

halter73: Flushing the response pipe should flush headers. StartAsync just commits them.

JunTaoLuo: I think I'll leave the implementation as is for now as that should still work. I see some test failures when I try to switch to BodyPipe.FlushAsync including HasStarted but I think it's a test. I'll investigate it later since I need this merged ASAP.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
JunTaoLuo: Ah StartAsync can complete synchronously.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -55,6 +55,28 @@ public static Task WriteMessageAsync(this PipeWriter pipeWriter, byte[] messageD
                 return Task.FromException(new RpcException(SendingMessageExceedsLimitStatus));
             }
 
+            // Must call StartAsync before the first pipeWriter.GetSpan() in WriteHeader
+            var response = serverCallContext.HttpContext.Response;
+            if (!response.HasStarted)
+            {
```
halter73: It might be cleaner to remove the above condition since that same condition is at the start of Response.StartAsync(). When the response has already started, StartAsync will immediately return a static completed task.

.travis.yml
JunTaoLuo: Using global.json to lock the version of the SDK and added scripts for installation.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -55,6 +55,24 @@ public static Task WriteMessageAsync(this PipeWriter pipeWriter, byte[] messageD
                 return Task.FromException(new RpcException(SendingMessageExceedsLimitStatus));
             }
 
+            // Must call StartAsync before the first pipeWriter.GetSpan() in WriteHeader
+            var startAsyncTask = serverCallContext.HttpContext.Response.StartAsync();
```
JamesNK: `WriteMessageAsync` can be called multiple times for a request when streaming. I would prefer that we explicitly check for `Response.HasStarted` before calling `Response.StartAsync()`. First, I think calling StartAsync on the response multiple times feels wrong. Second, it would slightly improve performance by not calling into the feature and having it no-op, and then checking the task.

JamesNK: Also when https://github.com/aspnet/AspNetCore/issues/7779 goes into Kestrel then we should only attempt to call `StartAsync` if we're not buffering.

If we're buffering then not calling `StartAsync` would allow the method to modify the response headers with `WriteResponseHeadersAsync` after calling `WriteMessageAsync`.

test/FunctionalTests/Infrastructure/GrpcTestFixture.cs
```diff
@@ -45,6 +46,8 @@ public GrpcTestFixture()
                 .UseStartup<TStartup>();
 
             _server = new TestServer(builder);
+            // Temporary workaround, this should be added to the TestServer by default
+            _server.Features.Set<IHttpResponseStartFeature>(new TestHttpResponseStartFeature());
```
JamesNK: I didn't know you could do this. I've been setting workaround features in the functional website's Startup.cs via middleware.

JunTaoLuo: I was wondering if this worked but seems like it didn't.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
jkotalik: To clarify, when you are buffering, are you replacing Kestrel's PipeWriter with a buffering one?

Issue #109, opened 2019-02-19
Unimplemented service/method


JamesNK:
## [unimplemented_method](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#unimplemented_method)

Imagine we have a service located at "Count.Counter". It has a method AccumulateCount method at "Count.Counter/AccumulateCount".

Attempting to call a method that doesn't exist on the service, e.g. "Count.Counter/DoNotExist" would return a UNIMPLEMENTED status.

I think we could do this via a policy matcher.

Count.Counter/AccumulateCount -> explicit endpoint
Count.Counter/* -> wildcard endpoint that returns UNIMPLEMENTED status for method

---

## [unimplemented_service](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#unimplemented_service)

Imagine we have a service located at "Count.Counter". Someone typos and tries to call "Count.CounterTypo". It should return an UNIMPLEMENTED status.

Doing this is more tricky because Grpc.AspNetCore.Server isn't necessarily the only framework in the web application. We don't want to completely take over request handling and assume that requests that don't match a service should always return a gRPC UNIMPLEMENTED status.

Would returning this status when the URL starts with the package name be enough to meet this requirement? e.g.

Count.Counter/AccumulateCount -> explicit endpoint
Count.Counter/* -> wildcard endpoint that returns UNIMPLEMENTED status for method
Count.* -> wildcard endpoint that returns UNIMPLEMENTED status for service

JamesNK:
@jtattermusch

jtattermusch:
Related to unimplemented_method and unimplemented_service in this issue https://github.com/grpc/grpc-dotnet/issues/32. Interop test descriptions https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md

jtattermusch:
**unimplemented_method** case seems to be relatively straightforward:
whatever URL starts with "/SERVICE_NAME/" where service_name is a gRPC service registered through MapGrpcService, we should have a fallback method handler that returns StatusCode.Unimplemented if a method doesn't exist. I think it's safe to do because registering a gRPC service is equivalent with reserving a whole "directory"  on the http server and it matches with that I'd expect as a user.

[...]

PR #113, opened 2019-02-20
Return unimplemented status for missing services and methods

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/109

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
@@ -71,6 +71,11 @@ namespace Grpc.AspNetCore.Server.Internal
 
         public static bool IsGrpcContentType(string contentType)
         {
+            if (contentType == null)
+            {
+                return false;
+            }
+
             if (!contentType.StartsWith(GrpcProtocolConstants.GrpcContentType, StringComparison.OrdinalIgnoreCase))
             {
                 return false;
@@ -145,6 +150,18 @@ namespace Grpc.AspNetCore.Server.Internal
             }
 
             return Convert.FromBase64String(decodable);
+        }
+
+        public static void AddProtocolHeaders(HttpResponse response)
+        {
+            response.ContentType = "application/grpc";
+            response.Headers.Append("grpc-encoding", "identity");
+        }
+
+        public static void AppendStatusTrailers(HttpResponse response, Status status)
+        {
+            response.AppendTrailer(GrpcProtocolConstants.StatusTrailer, status.StatusCode.ToTrailerString());
+            response.AppendTrailer(GrpcProtocolConstants.MessageTrailer, status.Detail);
         }
     }
 }
--- a/src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
@@ -77,50 +77,52 @@ namespace Grpc.AspNetCore.Server.Internal
         private void AddMethodCore(IMethod method, RequestDelegate requestDelegate)
         {
             ServiceMethods.Add(method);
-            EndpointConventionBuilders.Add(_builder.MapPost(method.FullName, requestDelegate, new object[] { method }));
+
+            // IMethod is added as metadata for the endpoint
+            var methodata = new object[] { method };
+
+            EndpointConventionBuilders.Add(_builder.MapPost(method.FullName, $"gRPC - {method.FullName}", requestDelegate, methodata));
         }
 
         internal void CreateUnimplementedEndpoints()
         {
-            // Return UNIMPLEMENTED status for requests with a URL that has two segments and a grpc/application content-type
-
             // Return UNIMPLEMENTED status for missing service:
-            // - /{service}/{method}
+            // - /{service}/{method} + content-type header = grpc/application
             if (_serviceMethodsRegistry.Methods.Count == 0)
             {
-                EndpointConventionBuilders.Add(CreateUnimplementedEndpoint("{unimplementedService}/{unimplementedMethod}", _serverCallHandlerFactory.CreateUnimplementedService()));
+                // Only one unimplemented service endpoint is needed for the application
+                EndpointConventionBuilders.Add(CreateUnimplementedEndpoint("{unimplementedService}/{unimplementedMethod}", "gRPC - Unimplemented service", _serverCallHandlerFactory.CreateUnimplementedService()));
             }
 
             // Return UNIMPLEMENTED status for missing method:
-            // - /Package.Service/{method}
+            // - /Package.Service/{method} + content-type header = grpc/application
+            var serviceNames = ServiceMethods.Select(m => m.ServiceName).Distinct().ToList();
 
             // Typically there should be one service name for a type
             // In case the bind method sets up multiple services in one call we'll loop over them
-            var serviceNames = ServiceMethods.Select(m => m.ServiceName).Distinct().ToList();
-
             foreach (var serviceName in serviceNames)
             {
                 if (_serviceMethodsRegistry.Methods.Any(m => string.Equals(m.ServiceName, serviceName, StringComparison.Ordinal)))
                 {
-                    // Unimplemented endpoint has already been added for this service
+                    // Only one unimplemented method endpoint is need for the service
                     continue;
                 }
 
-                EndpointConventionBuilders.Add(CreateUnimplementedEndpoint(serviceName + "/{unimplementedMethod}", _serverCallHandlerFactory.CreateUnimplementedMethod()));
+                EndpointConventionBuilders.Add(CreateUnimplementedEndpoint(serviceName + "/{unimplementedMethod}", $"gRPC - Unimplemented method for {serviceName}", _serverCallHandlerFactory.CreateUnimplementedMethod()));
             }
 
             _serviceMethodsRegistry.Methods.AddRange(ServiceMethods);
         }
 
-        private IEndpointConventionBuilder CreateUnimplementedEndpoint(string pattern, RequestDelegate requestDelegate)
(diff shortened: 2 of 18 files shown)


src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
```diff
@@ -18,47 +18,98 @@
...
+            ServiceMethods.Add(method);
+            EndpointConventionBuilders.Add(_builder.MapPost(method.FullName, requestDelegate, new object[] { method }));
+        }
+
+        internal void CreateUnimplementedEndpoints()
+        {
+            // Create endpoint that will return UNIMPLEMENTED status for missing method:
+            // - /Package.Service/{method}
+            // Create endpoint that will return UNIMPLEMENTED status for missing service:
+            // - /Package.{service}/{method}
```
jtattermusch: The case for unimplemented method make sense.

The case for unimplemented service is more tricky and needs some work, see https://github.com/grpc/grpc-dotnet/issues/109#issuecomment-465487229 (perhaps we could leave it for a followup PR?).

src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
```diff
@@ -18,47 +18,98 @@
...
         public override void AddMethod<TRequest, TResponse>(Method<TRequest, TResponse> method, UnaryServerMethod<TRequest, TResponse> handler)
         {
             var callHandler = _serverCallHandlerFactory.CreateUnary(method);
-            EndpointConventionBuilders.Add(_builder.MapPost(method.FullName, callHandler.HandleCallAsync));
+            AddMethodCore(method, callHandler.HandleCallAsync);
+        }
+
+        private void AddMethodCore(IMethod method, RequestDelegate requestDelegate)
+        {
+            ServiceMethods.Add(method);
```
jtattermusch: qq: what happens when we try to add a handler for the same method twice? (two "different" generated services that have the same URL, or registering the same service twice by mistake).
In gRPC C# registering the same method url twice is an error: https://github.com/grpc/grpc/blob/68381892d3fc52594e963a12b077e176021b36c4/src/csharp/Grpc.Core/Server.cs#L262

JamesNK: At application startup nothing will happen. When someone attempts to go to a URL registered twice they will get an error that multiple endpoints match that URL.

src/Grpc.AspNetCore.Server/Internal/ServerCallHandlerFactory.cs
```diff
@@ -72,5 +76,56 @@ public ServerCallHandlerFactory(ILoggerFactory loggerFactory, IOptions<GrpcServi
...
+                var unimplementedService = httpContext.Request.RouteValues["unimplementedService"]?.ToString();
+                ServiceUnimplemented(logger, unimplementedService);
+
+                GrpcProtocolHelpers.AppendStatusTrailers(httpContext.Response, new Status(StatusCode.Unimplemented, "Service is unimplemented."));
+                return Task.CompletedTask;
+            };
+        }
+    }
+
+    internal partial class ServerCallHandlerFactory<TService>
```
JunTaoLuo: nit: Move this to a separate file. ServerCallHandlerFactory.Log

src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
```diff
@@ -18,47 +18,98 @@
...
+                {
+                    var packageName = serviceName.Substring(0, packageDelimiterIndex + 1);
+
+                    // Unimplemented endpoint has already been added for this package
+                    if (_serviceMethodsRegistry.Methods.Any(m => m.ServiceName.StartsWith(packageName, StringComparison.Ordinal)))
+                    {
+                        continue;
+                    }
+
+                    EndpointConventionBuilders.Add(_builder.MapPost(packageName + "{unimplementedService}/{unimplementedMethod}", _serverCallHandlerFactory.CreateUnimplementedService()));
```
JunTaoLuo: Hmm now that I think of it, is the prefix sufficient to avoid conflicts with other non-gRPC endpoints? I'm not familiar with endpoints routing but what happens if multiple patterns are matched?

JamesNK: There is route precedence.

"Greet.Greeter/SayHello" is matched by both of these routes:

Route 1: "Greet.Greeter/SayHello"
Route 2: "{service}/{method}"

But the first one has a higher precedence because it matches literals instead of parameters.

JamesNK: > I'm not familiar with endpoints routing but what happens if multiple patterns are matched for example?

If two routes are matched and they have the same precedence then an error is thrown.

JunTaoLuo: Sorry, I was thinking of another scenario. See the updated comment.

JunTaoLuo: So a more concrete example:
Endpoint 1:
- Mapping a gRPC service called Namespace.GrpcService
- Creates a mapping of Namespace.*/* => unimplemented service request delegate
Endpoint 2:
- Mapping a static file server at Namespace.StaticServer
- Creates a mapping of Namespace.StaticServer/* => files

A request comes in with /Namespace.StaticServer/file.txt. This matches both routes and both would be considered parameters right? Then an error will be thrown?

JamesNK: If someone had `/Namespace.StaticServer/file.txt` as a route then that would match because it is a literal. It would only be a problem if it had the same literal prefix followed by two parameters.

Note: static files don't (go through routing.

JunTaoLuo: Doesn't have to be static files, any other non-gRPC request would suffice. I suppose it's fine to add this now and if we ever run into mapping issues, we'll look into alternatives.

src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
jtattermusch: I'm still not convinced that this way of handling unimplemented services is the correct approach. It's impossible to predict which exact routes will users use in their servers, so it's hard to reason about whether the "/Package.{service}/{method}" route will take precedence over the routes they have and if this will lead to unwanted behavior.

JamesNK: Sure. I commented on #109 a few days ago with a solution to also require content type = `application/grpc`. Would adding that be acceptable?

https://github.com/grpc/grpc-dotnet/issues/109#issuecomment-465717408

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
```diff
@@ -90,5 +92,17 @@ public static byte[] ParseBinaryHeader(string base64)
 
             return Convert.FromBase64String(decodable);
         }
+
+        public static void AddProtocolHeaders(HttpResponse response)
```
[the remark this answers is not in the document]
JamesNK: Isn't that what you asked for earlier?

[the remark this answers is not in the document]
jtattermusch: Sorry for not being clear. I was only proposing possible names for both methods - either Add* and Append* prefixes are fine, but I think both methods should be called consistently as they are basically doing the same thing. This is really a nit so feel free to ignore if you don't feel like it's important enough.

src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
jtattermusch: Yes, I think checking the content type is a good solution (and looks like we're planning to do that anyway) See https://github.com/grpc/grpc-dotnet/issues/109#issuecomment-467469913 for details.
Are you planning to update this PR with that approach?

JamesNK:
🆙 📅

src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
```diff
@@ -31,40 +35,106 @@ namespace Grpc.AspNetCore.Server.Internal
...
+            _serviceMethodsRegistry.Methods.AddRange(ServiceMethods);
+        }
+
+        private IEndpointConventionBuilder CreateUnimplementedEndpoint(string pattern, RequestDelegate requestDelegate)
+        {
+            var routePattern = RoutePatternFactory.Parse(pattern, null, new { contentType = GrpcContentTypeContraint.Instance });
+            return _builder.Map(routePattern, requestDelegate, new HttpMethodMetadata(new[] { "POST" }));
+        }
+
+        private class GrpcContentTypeContraint : IRouteConstraint
```
jtattermusch: typo: Contraint

src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
```diff
@@ -31,40 +35,106 @@ namespace Grpc.AspNetCore.Server.Internal
...
+            // Return UNIMPLEMENTED status for requests with a URL that has two segments and a grpc/application content-type
+
+            // Return UNIMPLEMENTED status for missing service:
+            // - /{service}/{method}
+            if (_serviceMethodsRegistry.Methods.Count == 0)
+            {
+                EndpointConventionBuilders.Add(CreateUnimplementedEndpoint("{unimplementedService}/{unimplementedMethod}", _serverCallHandlerFactory.CreateUnimplementedService()));
+            }
+
+            // Return UNIMPLEMENTED status for missing method:
```
jtattermusch: comment accuracy: looks like the content type constraint is being applied for unimplemented methods too but from the comment it doesn't look it.

PR #121, opened 2019-02-23
Support binding services that subclass implementation

JamesNK:


src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
```diff
@@ -92,6 +92,12 @@ private static void ReflectionBind<TService>(ServiceBinderBase binder, TService
             // We need to access the static BindService method on Foo which implicitly derives from Object.
             var baseType = serviceType.BaseType;
 
+            // Handle services that have multiple levels of inheritence
+            while (baseType?.BaseType?.BaseType != null)
```
JunTaoLuo: This is the same logic as https://github.com/grpc/grpc-dotnet/blob/5521bb3d1348f90be9335885a324691015453232/src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs#L41-L47. When was that lost?

JamesNK: It was pointed out that it didn't do anything in a code review so I removed it and no tests broke. It later occurred to me that it was probably to handle inheritance. This re-adds it with unit tests.

Issue #110, opened 2019-02-19
Customizing gRPC URLs?


JamesNK:
How strict is the gRPC standard with respect to URLs?

Today we generate URLs that always place them at the root of the web app.

e.g. /Count.Counter/AccumulateCount

Is there flexibility to have them prefixed with a URL segment in a web app?

e.g. /api/Count.Counter/AccumulateCount

If this is allowed then we should provide a way to customize service URLs in the startup experience.

dhhoang:
The way I understand it, URL prefixes don't work well with generated clients (see grpc/grpc/issues/14900).

JamesNK:
https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

> Some gRPC implementations may allow the Path format shown above to be overridden, but this functionality is strongly discouraged. gRPC does not go out of its way to break users that are using this kind of override, but we do not actively support it, and some functionality (e.g., service config support) will not work when the path is not of the form shown above.

dhhoang:
Question related to #43 
Will the managed client be generated as well (based on `.proto` file)? There doesn't seem to be a way to specify a custom url prefix in the `.proto` file .

JamesNK:
I could be configuration in the client. You'll already need to specify a base address. A prefix could be part of that, e.g. https://localhost:1111/prefix

dhhoang:
Right, but other generated clients might now work well with it. 
Let's say I have some `.proto` files and and build a Aspnet core server, and then generate my C#/Ruby/Java client with the Grpc core tool. As per https://github.com/grpc/grpc/issues/14900#issuecomment-378071244 the client would set the HTTP2 `:path` header based on the default format. The prefix wouldn't be compatible then.

JamesNK:
We'll be writing a managed client at some point. If we want to support a prefix then we'd include it in request's `:path`

kinglionsoft:
My grpc server is hosted by Service Fabric as a guest executable service. The accessing url of this service via the reverse proxy from outside the cluster becomes:
```
http://servicefabric:19081/GuestExeSample/GrpcHello
```
In order to using current grpc client to access this service, I have to transfer the url to another port without path through **nginx**.

```
server {
    listen 9001 http2;
    location / {
         grpc_pass grpc://servicefabric:19081/GuestExeSample/GrpcHello;
    }
}	
```

It's a good idea that a managed clien supportsa prefix. @JamesNK

[...]

PR #122, opened 2019-02-23
Validate request content type

JamesNK:
Accept application/grpc and application/grpc+proto. Other message formats, and requests with no content type get a 415 response. Matches grpc-java behaviour.

### DIFF
--- a/perf/Grpc.AspNetCore.Microbenchmarks/UnaryServerCallHandlerBenchmark.cs
+++ b/perf/Grpc.AspNetCore.Microbenchmarks/UnaryServerCallHandlerBenchmark.cs
@@ -23,7 +23,7 @@ using Chat;
 using Google.Protobuf;
 using Grpc.AspNetCore.Microbenchmarks.Internal;
 using Grpc.AspNetCore.Server;
-using Grpc.AspNetCore.Server.Internal;
+using Grpc.AspNetCore.Server.Internal.CallHandlers;
 using Grpc.Core;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
--- a/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
@@ -41,7 +41,7 @@ namespace Grpc.AspNetCore.Server.Internal.CallHandlers
             _invoker = invoker;
         }
 
-        public override async Task HandleCallAsync(HttpContext httpContext)
+        protected override async Task HandleCallAsyncCore(HttpContext httpContext)
         {
             httpContext.Response.ContentType = "application/grpc";
             httpContext.Response.Headers.Append("grpc-encoding", "identity");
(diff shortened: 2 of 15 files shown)


dhhoang:
Question: If I understand #109 correctly, the `application/grpc` header should be the way to recognize whether the client request is a Grpc call so that ASP.NET core routes the request to the service handler or let the rest of the stack takes care of it. 
Would having a middleware that recognizes this header be a more appropriate way to handle this? Something similar to  
```csharp
public async Task InvokeAsync(HttpContext context)
{
    if (!IsGrpcContentType(context.Request.Headers))
    {
        await _next(context);
        return;
    }

    // do Grpc stuffs
}
```

Or maybe another way is
```csharp
app.MapWhen(ctx => IsGrpcContentType(ctx.Request.Headers), app =>
{
    app.UseRouting(routes =>
    {
        routes.MapGrpcService<TService>();
    });
});
```

JamesNK:
You're thinking about something like the `[Consumes]` feature in MVC where you can have different actions response to a request depending on the request content-type.

It is definitely possible to setup routing using a route constraint or a matcher policy so that:

* `/Service/Method` + `application/grpc` calls a gRPC method
* `/Service/Method` + `application/json` calls something else entirely (MVC action?)

Would people want to do that? I don't think so. I think it is better experience to assume that the `/Service/Method` URL is owned by gRPC and invalid requests to it are an error.

And if someone did want to route to different endpoints based on content-type that could still do that themselves.

dhhoang:
What would happen for requests with `application/grpc` but do not match any Grpc routes or maybe match another mapping ? I thought the plan was to make them return one of the `unimplemented` statuses. https://github.com/grpc/grpc-dotnet/issues/109#issuecomment-465717408

> We could return an unimplemented status for any unmatching requests that have a `application/grpc (+proto/+json)` content-type that don't match a service.

From a design standpoint, including an RPC service with MVC or something else in one server seems rather bizarre. But if that is what people do, distinguishing the requests based on the header looks like a more clear and straightforward solution. Since the actual Grpc routing (`/Service/Method` format) is hidden from the users (at least for now), I can see some users accidentally set up routing that might duplicate these URLs. (especially with MVC's very familiar `Controller/Action` format.

JamesNK:
> What would happen for requests with `application/grpc` but do not match any Grpc routes or maybe match another mapping ?

To be decided.

> Since the actual Grpc routing (`/Service/Method` format) is hidden from the users (at least for now), I can see some users accidentally set up routing that might duplicate these URLs. (especially with MVC's very similar `Controller/Action` format.

The URL largely being invisible is a good point. If there are conflicting routes that can't be resolved then there will be an error explaining why.

I don't think it is a big worry but it can be mostly mitigated by allowing customization via a base path - https://github.com/grpc/grpc-dotnet/issues/110 - and documentation.

jtattermusch:
Actually, I have one comment:
The grpc-java server only checks that the  content type is `application/grpc` or `application/grpc+whatever`
https://github.com/grpc/grpc-java/blob/041cf2abd445f90d0bb41e48b0c226a5d09c13b8/core/src/main/java/io/grpc/internal/GrpcUtil.java#L425
https://github.com/grpc/grpc-java/blob/81da3eb95be37fa0647ce8da2e19de96ab84c600/core/src/test/java/io/grpc/internal/GrpcUtilTest.java#L138

if we only accept `application/grpc+proto`, we will shut the door to using other serialization formats  (which is doable even now I think, by using a custom marshaller in grpc service binder).
What grpc-java is doing seems more reasonable.

PR #125, opened 2019-02-24
Add compression

JamesNK:


src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
```diff
@@ -90,5 +93,71 @@ public static byte[] ParseBinaryHeader(string base64)
 
             return Convert.FromBase64String(decodable);
         }
+
+        internal static bool TryDecompressMessage(string compressionEncoding, byte[] messageData, out byte[] result)
+        {
+            switch (compressionEncoding)
+            {
+                case "gzip":
```
JamesNK: Hardcoded. Add an API for new compression encodings? For example, https://docs.microsoft.com/en-us/aspnet/core/performance/response-compression?view=aspnetcore-2.2#custom-providers

test/FunctionalTests/CompressionTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            httpRequest.Content = new StreamContent(requestStream);
+
+            // Act
+            var response = await Fixture.Client.SendAsync(httpRequest).DefaultTimeout();
+
+            // Assert
+            Assert.AreEqual(HttpStatusCode.OK, response.StatusCode);
+
+            // The overall encoding is gzip but the actual response does not use compression
+            Assert.AreEqual("gzip", response.Headers.GetValues("grpc-encoding").Single());
```
JamesNK: Correct?

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
```diff
@@ -90,5 +93,71 @@ public static byte[] ParseBinaryHeader(string base64)
...
+                    {
+                        var output = new MemoryStream();
```
davidfowl: The allocations in here are just.... Not great

JamesNK: Yeah. All the compression APIs in .NET are stream based at the moment which makes it difficult to do efficiently 😐 

Compression off by default and is something you opt-in to. People can turn it on if they have large payloads and wire-size is more important that server allocations.

Also this will be revisited when we can get protobuf to use span.

davidfowl: We still need to measure this allocates like crazy and I’d bet is extremely inefficient. I think we can change this code a bit to improve that but I don’t think we should merge this until we get perf numbers.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -77,7 +79,28 @@ private static async Task WriteMessageCoreAsyncAwaited(this PipeWriter pipeWrite
...
+            //   2. Use compression algorithm of request
+            var compressionEncoding = canCompress
+                ? serverCallContext.GetGrpcEncoding()
+                : null;
+
+            if (compressionEncoding != null)
+            {
+                if (!GrpcProtocolHelpers.TryCompressMessage(compressionEncoding, messageData, out var compressedData))
+                {
+                    // It is not an error if the server doesn't support the incoming compression algorithm
```
JunTaoLuo: cc @jtattermusch ?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -250,5 +250,24 @@ public void Dispose()
...
+            // In order of precedence:
+            // 1. Compression algorithm configured on server
+            // 2. Compression algorithm of request
+
+            if (ServiceOptions.CompressionAlgorithm != null)
+            {
+                return ServiceOptions.CompressionAlgorithm;
+            }
+
+            if (!HttpContext.Request.Headers.TryGetValue(GrpcProtocolConstants.EncodingHeader, out var values))
```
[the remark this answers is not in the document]
JunTaoLuo: Nevermind, I misunderstood the spec. This is correct.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
JamesNK: Add providers

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -250,5 +250,24 @@ public void Dispose()
         {
             _deadlineTimer?.Dispose();
         }
+
+        internal string GetGrpcEncoding()
+        {
+            // In order of precedence:
```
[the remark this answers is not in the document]
JamesNK: That line in the spec refers to how the client calculates the request compression.

This code is for the server's response compression.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
```diff
@@ -90,5 +93,71 @@ public static byte[] ParseBinaryHeader(string base64)
...
+        }
+
+        internal static bool TryCompressMessage(string compressionEncoding, byte[] messageData, out byte[] result)
+        {
+            switch (compressionEncoding)
+            {
+                case "gzip":
+                    {
+                        var output = new MemoryStream();
+                        using (var compressionStream = new GZipStream(output, CompressionMode.Compress))
```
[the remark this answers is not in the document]
JamesNK: The GzipCompressionProvider accepts an argument for the compression level.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -313,6 +333,23 @@ private static bool TryReadMessage(ref ReadOnlySequence<byte> buffer, HttpContex
...
+                string encoding = context.GetGrpcEncoding();
+                if (encoding == null)
+                {
+                    throw new RpcException(NoMessageEncodingMessageStatus);
+                }
+
+                // Performance improvement would be to decompress without converting to an intermediary byte array
+                if (!GrpcProtocolHelpers.TryDecompressMessage(encoding, message, out var decompressedMessage))
+                {
+                    throw new RpcException(UnknownMessageEncodingMessageStatus);
```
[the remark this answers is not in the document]
JamesNK: Done

src/Grpc.AspNetCore.Server/Compression/ICompressionProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <summary>
+        /// Create a new compression stream.
+        /// </summary>
+        /// <param name="stream">
+        /// The stream where the compressed data is written when <paramref name="compressionMode"/> is <c>Compress</c>,
+        /// and where compressed data is copied from when <paramref name="compressionMode"/> is <c>Decompress</c>.
+        /// </param>
+        /// <param name="compressionMode">The compression mode.</param>
+        /// <returns>A stream used to compress or decompress data.</returns>
+        Stream CreateStream(Stream stream, CompressionMode compressionMode);
```
JunTaoLuo: Hmm we are currently working with byte[] which means we need to wrap them in streams for this API. How do we plan to adapt this in the world of Pipes/Span that we plan to move to?

JamesNK: I don't know. But all .NET compression is stream based at the moment.

There might be opportunities to use StreamPipeReader and StreamPipeReader with this. To be discovered as we optimize further.

JunTaoLuo: Nevermind, I see Gzip is implemented as a stream, similar to other popular providers. This makes sense.

JamesNK:
All feedback resolved. Looking for those sweet sweet likes and subscribes.

🆙 📅

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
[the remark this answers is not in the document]
JamesNK: Ah, I didn't see this doc. Changes related to this feedback - https://github.com/grpc/grpc-dotnet/pull/125/commits/8289c3c6ca4979bec3d98b46c71da4fdf5cc23d2

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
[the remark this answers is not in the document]
JamesNK: Changes here - https://github.com/grpc/grpc-dotnet/pull/125/commits/45dc0f6b410597f2c7265b28de84531b4625c04a

src/Grpc.AspNetCore.Server/Internal/GrpcServiceOptionsSetup.cs
```diff
@@ -33,6 +35,10 @@ public void Configure(GrpcServiceOptions options)
             {
                 options.ReceiveMaxMessageSize = DefaultReceiveMaxMessageSize;
             }
+            if (options._compressionProviders == null || options._compressionProviders.Count == 0)
```
JunTaoLuo: if we are going to add a default, is there any reason to optimize for the case where _compressionProvider is null? I think _compressionProviders should be set to private and just initialize it as an empty list.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -218,6 +219,13 @@ public void Initialize()
             {
                 _deadline = DateTime.MaxValue;
             }
+
+            if (HttpContext.Request.Headers.TryGetValue(GrpcProtocolConstants.MessageEncodingHeader, out var values))
+            {
+                _grpcEncoding = values;
+            }
+
+            HttpContext.Response.Headers.Append(GrpcProtocolConstants.MessageEncodingHeader, GetGrpcEncoding() ?? GrpcProtocolConstants.IdentityGrpcEncoding);
```
JunTaoLuo: So by default the response will be encoded in the same way as the request? Hmm though esoteric, I think we should check if an grpc-accept-encoding is available on the request for us to consider. If not, we can fallback to using the same encoding as the request.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -317,6 +344,40 @@ private static bool TryReadMessage(ref ReadOnlySequence<byte> buffer, HttpContex
             var messageBuffer = buffer.Slice(HeaderSize, messageLength);
             message = messageBuffer.ToArray();
 
+            if (compressed)
+            {
+                string encoding = context.GetGrpcEncoding();
+                if (encoding == null)
+                {
+                    throw new RpcException(NoMessageEncodingMessageStatus);
+                }
```
JunTaoLuo: nit: new line after brace

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -317,6 +344,40 @@ private static bool TryReadMessage(ref ReadOnlySequence<byte> buffer, HttpContex
...
+
+                    if (!context.HttpContext.Response.HasStarted)
+                    {
+                        context.HttpContext.Response.Headers[GrpcProtocolConstants.MessageAcceptEncodingHeader] = string.Join(",", supportedEncodings);
+                    }
+
+                    throw new RpcException(CreateUnknownMessageEncodingMessageStatus(encoding, supportedEncodings));
+                }
+
+                context.ValidateGrpcAcceptEncoding();
```
JunTaoLuo: Wait a minute, why do we need to validate the encoding here? As long as the server can decode the request it's fine.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -78,7 +86,29 @@ private static async Task WriteMessageCoreAsyncAwaited(this PipeWriter pipeWrite
...
         private static Task WriteMessageCoreAsync(this PipeWriter pipeWriter, byte[] messageData, HttpContextServerCallContext serverCallContext, bool flush)
         {
-            WriteHeader(pipeWriter, messageData.Length);
+            var canCompress = serverCallContext.CanWriteCompressed();
+
+            // If compression has not been disabled by WriteFlags.NoCompress then
+            //   1. Use compression algorithm of request
+            //   2. Use default compression algorithm configured on server
+            var compressionEncoding = canCompress
+                ? serverCallContext.GetGrpcEncoding()
```
JunTaoLuo: This doesn't address https://github.com/grpc/grpc-dotnet/pull/125#discussion_r268081897. Why is the encoding of the request used to compress the response?

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
JamesNK: https://user-images.githubusercontent.com/2030323/54811447-ef22cb00-4c45-11e9-8b73-23bb296f19be.png

"Is the algorithm present in the incoming grpc-accept-encoding?"

src/Grpc.AspNetCore.Server/Compression/GzipCompressionProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// </summary>
+    public class GzipCompressionProvider : ICompressionProvider
+    {
+        private readonly CompressionLevel _compressionLevel;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="GzipCompressionProvider"/> class with the specified <see cref="CompressionLevel"/>.
+        /// </summary>
+        /// <param name="compressionLevel">The compression level to use when compressing data.</param>
+        public GzipCompressionProvider(CompressionLevel compressionLevel)
```
JunTaoLuo: What about per call compression level settings? https://github.com/grpc/grpc/blob/master/doc/compression_cookbook.md#compression-level-in-call-responses

JunTaoLuo: Actually I think it's sufficient that we provide functionality through WriteFlags.NoCompress.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
JunTaoLuo: Ah, I see. Then I'm really confused as to what the grpc-accept-encoding means then. Let's talk about this during standup.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
JunTaoLuo: Actually I may be misunderstanding the flow chart a little.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
jtattermusch: I don't think the server should respond with the same compression as that client used. Server should not use compression unless it has been configured to do so (see e-mail for more discussion).

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -252,5 +260,65 @@ public void Dispose()
         {
             _deadlineTimer?.Dispose();
         }
+
+        internal string GetGrpcEncoding()
+        {
+            // In order of precedence:
+            // 1. Compression algorithm of request
+            // 2. Compression algorithm configured on server
+            return _grpcEncoding ?? ServiceOptions.DefaultCompressionAlgorithm;
```
jtattermusch: I suspect this is incorrect. See e-mail discussion.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -252,5 +260,65 @@ public void Dispose()
...
+        }
+
+        internal bool GrpcAcceptEncodingMatch(string encoding)
```
jtattermusch: nit: it's immediately not obvious what this method checks (=that the grpc-encoding is contained in grpc-accept-encoding). Let's add a comment or rename.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
jtattermusch: This one just logs a message to make sure the implementation is compliant with a the spec.
perhaps rename the ValidateGrpcAcceptEncoding() to make it clearer what it does?

src/Grpc.AspNetCore.Server/GrpcServiceOptions.cs
```diff
@@ -38,6 +44,28 @@ public class GrpcServiceOptions
         /// Detailed error messages include details from exceptions thrown on the server.
         /// </summary>
         public bool? EnableDetailedErrors { get; set; } = null;
+
+        /// <summary>
+        /// Gets or sets the default compression algorithm used to compress messages sent from the server.
+        /// The grpc-encoding header value in a request will override this setting.
+        /// </summary>
+        public string DefaultCompressionAlgorithm { get; set; }
```
jtattermusch: Exposing DefaultCompressionAlgorithm makes sense as it's also exposed by C++ server, 
but I think the recommendation is to use DefaultCompressionLevel instead, because it allows the server to choose the compression algorithm intelligently based on what the client supports (and announces in  grpc-accept-encoding). 
https://github.com/grpc/grpc/blob/bf7ed7a053dc799419f8ff7b278f3af347c309fe/include/grpcpp/server_builder.h#L184

I think its fine to add that functionality later though, exposing DefaultCompressionAlgorithm seems good enough for now and it is not incorrect.

Besides that, the C++ server also allows configuring the compression level at the per-rpc level (in ServerContext) and per-message. I think for now we don't need to support those.
The important parts is to be able to read compressed request, comply with the spec when responding and to be able to disable compression on per-message basis (using WriteFlags.NoCompress, as that's important for security reasons) - and this PR mostly has those.

test/FunctionalTests/CompressionTests.cs
```diff
@@ -0,0 +1,311 @@
...
+using Grpc.Core;
+using NUnit.Framework;
+
+namespace Grpc.AspNetCore.FunctionalTests
+{
+    [TestFixture]
+    public class CompressionTests : FunctionalTestBase
+    {
+        [TestCase("gzip")]
+        public async Task SendCompressedMessage_CompressedResponseOfSameEncoding(string encoding)
```
jtattermusch: this is probably not the behavior we want, as explained above.

test/FunctionalTests/CompressionTests.cs
```diff
@@ -0,0 +1,311 @@
...
+using Grpc.AspNetCore.Server.Compression;
+using Grpc.AspNetCore.Server.Internal;
+using Grpc.AspNetCore.Server.Tests;
+using Grpc.Core;
+using NUnit.Framework;
+
+namespace Grpc.AspNetCore.FunctionalTests
+{
+    [TestFixture]
+    public class CompressionTests : FunctionalTestBase
```
jtattermusch: it might be a good idea to try to add test scenarios that match the behavior described in the compression interop tests:
https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#server_compressed_unary
https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#server_compressed_streaming

test/FunctionalTests/CompressionTests.cs
```diff
@@ -0,0 +1,311 @@
...
+            Assert.AreEqual("gzip", response.Headers.GetValues("grpc-encoding").Single());
+
+            var responseMessage = MessageHelpers.AssertReadMessage<HelloReply>(await response.Content.ReadAsByteArrayAsync().DefaultTimeout(), "gzip");
+            Assert.AreEqual("Hello World", responseMessage.Message);
+
+            Assert.AreEqual(StatusCode.OK.ToTrailerString(), Fixture.TrailersContainer.Trailers[GrpcProtocolConstants.StatusTrailer].ToString());
+        }
+
+        [Test]
+        public async Task SendIdentityGrpcEncodingToServiceWithCompression_ResponseUncompressed()
```
jtattermusch: the name of the test case seems a bit off, as the reason for responding with "identity" is the grpc-accept-encoding header only containing "identity", not related to what the request compression was.

test/FunctionalTests/CompressionTests.cs
JamesNK: That is covered with this functional test: SendCompressedMessageAndReturnResultWithNoCompressFlag_ResponseNotCompressed

test/FunctionalTests/CompressionTests.cs
JamesNK: Renamed to SendIdentityGrpcAcceptEncodingToServiceWithCompression_ResponseUncompressed

src/Grpc.AspNetCore.Server/GrpcServiceOptions.cs
JamesNK: The compression level can be specified when you setup the compression provider for the algorithm. For example, you would configure a gzip compression provider and set the compression level for it then.

src/Grpc.AspNetCore.Server/Internal/GrpcServiceOptionsSetup.cs
JamesNK: A null value is needed for inheritance from the global settings to the per-service settings.

src/Grpc.AspNetCore.Server/GrpcServiceOptions.cs
JamesNK: On this name, DefaultCompressionAlgorithm no longer seems the right name. Now the server will only compress messages when the request grpc-accept-endoing and DefaultCompressionAlgorithm match. Should this be renamed to ResponseCompressionAlgorithm?

JamesNK:
🆙 📅

src/Grpc.AspNetCore.Server/GrpcServiceOptions.cs
JamesNK: I have:

* Renamed DefaultCompressionAlgorithm to ResponseCompressionAlgorithm
* Added ResponseCompressionLevel

src/Grpc.AspNetCore.Server/Compression/GzipCompressionProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+
+using System.IO;
+using System.IO.Compression;
+
+namespace Grpc.AspNetCore.Server.Compression
+{
+    /// <summary>
+    /// GZIP compression provider.
+    /// </summary>
+    public class GzipCompressionProvider : ICompressionProvider
```
jtattermusch: nit: API stability: once we declare the API as stable, we won't be able to add new methods to ICompressionProvider (and we might need to add some for e.g. pipepine-compatible compression support one day). Not a concern now, just FYI.

src/Grpc.AspNetCore.Server/GrpcServiceOptions.cs
```diff
@@ -38,6 +44,35 @@ public class GrpcServiceOptions
         /// Detailed error messages include details from exceptions thrown on the server.
         /// </summary>
         public bool? EnableDetailedErrors { get; set; } = null;
+
+        /// <summary>
+        /// Gets or sets the compression algorithm used to compress messages sent from the server.
+        /// The request grpc-accept-encoding header value must contain this algorithm for it to
+        /// be used.
+        /// </summary>
+        public string ResponseCompressionAlgorithm { get; set; }
```
jtattermusch: I think the original names `DefaultCompressionAlgorithm` and `DefaultCompressionLevel` were ok (I think internally in the impl, calling some of the methods/fields "Response*" makes sense to distinguish between response and request logic, but there's not reason for the public API to reflect that). For C++ server, these are also called DefaultCompressionAlgorithm and DefaultCompressionLevel.  In grpc-java the "reponse" prefix is also not being used for setting the compression: e.g.
https://github.com/grpc/grpc-java/blob/master/examples/src/main/java/io/grpc/examples/experimental/CompressingHelloWorldServerPerMethod.java#L89

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
jtattermusch: The comment needs to be updated and logic revisited?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -252,5 +267,67 @@ public void Dispose()
...
+            {
+                return values;
+            }
+
+            return null;
+        }
+
+        internal bool IsEncodingInRequestAcceptEncoding(string encoding)
+        {
+            if (HttpContext.Request.Headers.TryGetValue(GrpcProtocolConstants.MessageAcceptEncodingHeader, out var values))
```
JunTaoLuo: > Currently however, compression levels are only supported at the server side, which is aware of the client's capabilities through the incoming Message-Accept-Encoding header.

I'm reading the specs and the emails again and I'd like to confirm that "Message-Accept-Encoding" refers to `grpc-accept-encoding` instead of the `accept-encoding` defined by HTTP

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
JunTaoLuo: So this happens when the compression encoding is set on the server for which a provider cannot be found. This would be an user error so we should just log a warning and send the uncompressed message.

For example, this would occur if the server has `gzip` provider added but the ResponseCompressionAlgorithm on the GrpcServiceOption is set to `deflate`.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -317,6 +344,40 @@ private static bool TryReadMessage(ref ReadOnlySequence<byte> buffer, HttpContex
...
+
+                    if (!context.HttpContext.Response.HasStarted)
+                    {
+                        context.HttpContext.Response.Headers[GrpcProtocolConstants.MessageAcceptEncodingHeader] = string.Join(",", supportedEncodings);
+                    }
+
+                    throw new RpcException(CreateUnknownMessageEncodingMessageStatus(encoding, supportedEncodings));
+                }
+
+                context.ValidateAcceptEncodingContainsResponseEncoding();
```
JunTaoLuo: Curious why this check is needed?

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
JunTaoLuo: Actually I think we have this check in the ServerCallHandlerFactory, so this case can't happen.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -252,5 +267,67 @@ public void Dispose()
...
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        internal void ValidateAcceptEncodingContainsResponseEncoding()
+        {
+            if (ResponseGrpcEncoding != null)
```
JunTaoLuo: Looking at the initialize code path, this property can never be null can it?

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -78,7 +86,29 @@ private static async Task WriteMessageCoreAsyncAwaited(this PipeWriter pipeWrite
 
         private static Task WriteMessageCoreAsync(this PipeWriter pipeWriter, byte[] messageData, HttpContextServerCallContext serverCallContext, bool flush)
         {
-            WriteHeader(pipeWriter, messageData.Length);
+            var isCompressed =
+                serverCallContext.CanWriteCompressed() &&
+                serverCallContext.ResponseGrpcEncoding != null &&
```
JunTaoLuo: Same here, can this ever be null?

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
JamesNK: Correct. This can be simplified.

src/Grpc.AspNetCore.Server/Compression/GzipCompressionProvider.cs
JamesNK: I know. We'll do an API review at some point. ICompressionProvider might turn into an abstract base class so we can extend it.

src/Grpc.AspNetCore.Server/GrpcServiceOptions.cs
JamesNK: In the Java example it is just called compression, but it is on the response stream.

I don't like DefaultCompressionAlgorithm because nothing ever overrides it.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: Yes. MessageAcceptEncoding is what it is called in the gRPC HTTP2 spec.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
JamesNK: That's the second time you've asked that in this PR 😋

src/Grpc.AspNetCore.Server/Compression/GzipCompressionProvider.cs
```diff
@@ -0,0 +1,64 @@
...
```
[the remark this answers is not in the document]
JamesNK: Lets do it as a follow up

JamesNK:
🆙 📅 

Looking for likes and subscribes 🙏

jtattermusch:
> :shipit: I think we've iterated enough on the implementation. The interop tests run for each PR isn't the exhaustive list right? I see `cloud_to_cloud:csharp:aspnetcore_server:...` but not `cloud_to_cloud:java:aspnetcore_server:` for example.

Yes, currently this only does C# Grpc.Core client against Grpc.AspNetCore.Server
https://github.com/grpc/grpc-dotnet/blob/a064946e614f412da59cfe4d6a9aa1d51e127e45/kokoro/interop.sh#L30
(we could add more but that also will results in the test taking longer time and it will increase risk of test flakes unrelated to this repo).

src/Grpc.AspNetCore.Server/GrpcServiceOptions.cs
jtattermusch: Fair enough.

JamesNK:
Hurrah! Thanks for everyone's feedback.

PR #126, opened 2019-02-25
Verify server logging

JamesNK:
Update functional tests to inject a logger factory into the test website and inspects log messages at the end of each test. Verifies no unexpected exceptions are thrown.

Also explicitly throw an error when attempting to write a message after the deadline has passed.

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/ServerCallHandlerBase.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/ServerCallHandlerBase.cs
@@ -34,7 +34,7 @@ namespace Grpc.AspNetCore.Server.Internal
         {
             Method = method;
             ServiceOptions = serviceOptions;
-            Logger = loggerFactory.CreateLogger(GetType());
+            Logger = loggerFactory.CreateLogger(typeof(TService));
         }
 
         public abstract Task HandleCallAsync(HttpContext httpContext);
--- a/test/FunctionalTests/DeadlineTests.cs
+++ b/test/FunctionalTests/DeadlineTests.cs
@@ -64,7 +64,7 @@ namespace Grpc.AspNetCore.FunctionalTests
 
             var messageCount = 0;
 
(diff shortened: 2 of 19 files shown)


src/Grpc.AspNetCore.Server/Internal/ServerCallHandlerBase.cs
```diff
@@ -34,7 +34,7 @@ protected ServerCallHandlerBase(Method<TRequest, TResponse> method, GrpcServiceO
         {
             Method = method;
             ServiceOptions = serviceOptions;
-            Logger = loggerFactory.CreateLogger(GetType());
```
JunTaoLuo: Is this an optimization?

test/FunctionalTests/DeadlineTests.cs
```diff
@@ -90,5 +90,84 @@ public async Task WriteUntilDeadline_SuccessResponsesStreamed(string requestUri)
...
+            static async Task WriteUntilError(HelloRequest request, IServerStreamWriter<HelloReply> responseStream, ServerCallContext context)
+            {
+                var i = 0;
+                while (true)
+                {
+                    var message = $"How are you {request.Name}? {i}";
+                    await responseStream.WriteAsync(new HelloReply { Message = message });
+                    i++;
+
+                    await Task.Delay(10);
```
JunTaoLuo: I'm wary of task delays in tests. Looks like in this case, it's just to reduce the iteration speed since the timeout is set to 200ms?

JamesNK: Yeah. No need to blast through messages as quickly as possible.

src/Grpc.AspNetCore.Server/Internal/ServerCallHandlerBase.cs
JamesNK: It is so the error is related to the user's service implementation type.

Issue #128, opened 2019-02-26
Add doc comments (XML IntelliSense) for public methods in Grpc.AspNetCore.Server


DamianEdwards:
😢 

![image](https://user-images.githubusercontent.com/249088/53438498-e1ff1d00-39b4-11e9-9190-1d08962da44c.png)

@JamesNK @shirhatti

PR #130, opened 2019-02-26
Generate xml documentation on build

JamesNK:
Addresses #128

@DamianEdwards


JamesNK:
@jtattermusch Are any build changes required to include the XML docs? I'm going to assume that with `dotnet pack` it will Just Work.

Issue #134, opened 2019-02-27
Investigate using HttpClientFactory for gRPC client


JamesNK:
HttpClientFactory background - https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests

Investigate using HttpClientFactory to create gRPC clients inside the context of Grpc.AspNetCore.Server.

Standard HttpClientFactory benefits:
* DI driven
* Inject cross-cutting concerns
* Smart HttpClient pooling

Possible gRPC specific benefits:
* Hide setting up the CallInvoker inside the factory
* Automatically propagate deadline, cancellation context, correlation and tracing identifier

// @rynowak

PR #142, opened 2019-03-06
Create gPRC HttpClient project and HttpClientFactory integration

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/134

* Adds John's gRPC HttpClient implementation as `Grpc.NetCore.HttpClient` project
* Adds HttpClientFactory integration as `Grpc.AspNetCore.Server.GrpcClient` project. @rynowak it would be great if you lend your expert eyes to this area

This PR contains experimental and prototype code but does not touch Grpc.AspNetCore.Server.

There are many many issues to decide on and clean up in future PRs:

* Tests. There is one integration test. Need more
* Split client types out from Grpc.Core to Grpc.Core.Api and change `Grpc.NetCore.HttpClient`'s package reference
* `HttpClientCallInvoker` has some issues: exceptions are not thrown back to the caller, I noticed duplex content was not returned until the caller was closed
* Update HttpClientCallInvoker to use trailer headers. Blocked on HttpClient updates
* c-core gRPC clients are thin wrappers over a channel, and the channel is disposed of. Need to figure out what the life-cycle of the HttpClient will be when used with a gRPC client.
* Figure out what server values can be automatically propagated to a client in the HttpClientFactory. Deadline, cancellation token, propagation token. Are there others?

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -9,9 +9,10 @@
     <MicrosoftAspNetCoreTestHostPackageVersion>3.0.0-preview-18579-0056</MicrosoftAspNetCoreTestHostPackageVersion>
     <MicrosoftExtensionsLoggingTestingPackageVersion>2.0.0-*</MicrosoftExtensionsLoggingTestingPackageVersion>
     <MicrosoftNETTestSdkPackageVersion>16.0.0-preview-20181205-02</MicrosoftNETTestSdkPackageVersion>
+    <MicrosoftSourceLinkGitHubPackageVersion>1.0.0-beta2-18618-05</MicrosoftSourceLinkGitHubPackageVersion>
     <MoqPackageVersion>4.10.0</MoqPackageVersion>
     <NunitPackageVersion>3.11.0</NunitPackageVersion>
     <Nunit3TestAdapterPackageVersion>3.12.0</Nunit3TestAdapterPackageVersion>
-    <MicrosoftSourceLinkGitHubPackageVersion>1.0.0-beta2-18618-05</MicrosoftSourceLinkGitHubPackageVersion>
+    <SystemIOPipelinesPackageVersion>4.6.0-preview.19073.11</SystemIOPipelinesPackageVersion>
   </PropertyGroup>
 </Project>
--- a/examples/Clients/Chatter/Program.cs
+++ b/examples/Clients/Chatter/Program.cs
@@ -17,12 +17,16 @@
 #endregion
 
 using System;
+using System.IO;
+using System.Net.Http;
 using System.Runtime.InteropServices;
+using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
 using Chat;
 using Common;
 using Grpc.Core;
+using Grpc.NetCore.HttpClient;
 
 namespace Sample.Clients
 {
@@ -39,10 +43,8 @@ namespace Sample.Clients
             var name = args[0];
 
             // Server will only support Https on Windows and Linux
-            var credentials = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? ChannelCredentials.Insecure : ClientResources.SslCredentials;
-
-            var channel = new Channel("localhost:50051", credentials);
-            var client = new Chatter.ChatterClient(channel);
+            var certificate = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? null : new X509Certificate2(Path.Combine(Resources.CertDir, "client.crt"));
+            var client = GrpcClientFactory.Create<Chatter.ChatterClient>(@"https://localhost:50051", certificate);
 
             using (var chat = client.Chat())
             {
@@ -67,7 +69,7 @@ namespace Sample.Clients
             }
 
             Console.WriteLine("Shutting down");
-            channel.ShutdownAsync().Wait();
+            //client.Dispose();
             Console.WriteLine("Press any key to exit...");
             Console.ReadKey();
 
--- a/examples/Clients/Chatter/Program.cs
+++ b/examples/Clients/Chatter/Program.cs
@@ -17,16 +17,12 @@
 #endregion
 
 using System;
-using System.IO;
-using System.Net.Http;
 using System.Runtime.InteropServices;
-using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
 using Chat;
 using Common;
 using Grpc.Core;
-using Grpc.NetCore.HttpClient;
 
 namespace Sample.Clients
 {
@@ -43,8 +39,10 @@ namespace Sample.Clients
             var name = args[0];
 
             // Server will only support Https on Windows and Linux
-            var certificate = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? null : new X509Certificate2(Path.Combine(Resources.CertDir, "client.crt"));
-            var client = GrpcClientFactory.Create<Chatter.ChatterClient>(@"https://localhost:50051", certificate);
+            var credentials = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? ChannelCredentials.Insecure : ClientResources.SslCredentials;
+
+            var channel = new Channel("localhost:50051", credentials);
+            var client = new Chatter.ChatterClient(channel);
 
             using (var chat = client.Chat())
             {
@@ -69,7 +67,7 @@ namespace Sample.Clients
             }
 
             Console.WriteLine("Shutting down");
-            //client.Dispose();
+            await channel.ShutdownAsync();
             Console.WriteLine("Press any key to exit...");
             Console.ReadKey();
 
--- a/src/Grpc.AspNetCore.Server.GrpcClient/DefaultGrpcClientFactory.cs
+++ b/src/Grpc.AspNetCore.Server.GrpcClient/DefaultGrpcClientFactory.cs
@@ -1,0 +1,55 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using Grpc.Core;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using System;
+using System.Net.Http;
+
+namespace Grpc.AspNetCore.Server.GrpcClient
+{
+    internal class DefaultGrpcClientFactory : IGrpcClientFactory
+    {
+        private readonly IServiceProvider _serviceProvider;
+        private readonly IHttpClientFactory _httpClientFactory;
+        private readonly IOptionsMonitor<GrpcClientOptions> _clientOptionsMonitor;
+
+        public DefaultGrpcClientFactory(IServiceProvider serviceProvider, IHttpClientFactory httpClientFactory, IOptionsMonitor<GrpcClientOptions> clientOptionsMonitor)
+        {
+            _serviceProvider = serviceProvider;
+            _httpClientFactory = httpClientFactory;
+            _clientOptionsMonitor = clientOptionsMonitor;
+        }
+
+        public TClient CreateClient<TClient>(string name) where TClient : ClientBase
+        {
+            var options = _clientOptionsMonitor.Get(name);
+            if (!options.ExplicitlySet)
+            {
+                throw new InvalidOperationException($"No gRPC client configured with name '{name}'.");
+            }
+
+            var httpClient = _httpClientFactory.CreateClient(name);
+
+            var typedHttpClientFactory = _serviceProvider.GetRequiredService<INamedTypedHttpClientFactory<TClient>>();
+
+            return typedHttpClientFactory.CreateClient(httpClient, name);
+        }
+    }
+}
--- a/src/Grpc.AspNetCore.Server.GrpcClient/Grpc.AspNetCore.Server.GrpcClient.csproj
+++ b/src/Grpc.AspNetCore.Server.GrpcClient/Grpc.AspNetCore.Server.GrpcClient.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.0</TargetFramework>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Grpc.AspNetCore.Server.GrpcClient/Grpc.AspNetCore.Server.GrpcClient.csproj
+++ b/src/Grpc.AspNetCore.Server.GrpcClient/Grpc.AspNetCore.Server.GrpcClient.csproj
@@ -1,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+
+    <ProjectReference Include="..\Grpc.NetCore.HttpClient\Grpc.NetCore.HttpClient.csproj" />
+    <PackageReference Include="Grpc.Core" Version="$(GrpcCorePackageVersion)" />
+  </ItemGroup>
+</Project>
--- a/src/Grpc.AspNetCore.Server.GrpcClient/Grpc.AspNetCore.Server.GrpcClient.csproj
+++ b/src/Grpc.AspNetCore.Server.GrpcClient/Grpc.AspNetCore.Server.GrpcClient.csproj
@@ -8,6 +8,7 @@
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
 
+    <ProjectReference Include="..\Grpc.AspNetCore.Server\Grpc.AspNetCore.Server.csproj" />
     <ProjectReference Include="..\Grpc.NetCore.HttpClient\Grpc.NetCore.HttpClient.csproj" />
     <PackageReference Include="Grpc.Core" Version="$(GrpcCorePackageVersion)" />
   </ItemGroup>
--- a/src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientOptions.cs
+++ b/src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientOptions.cs
@@ -46,5 +46,8 @@ namespace Grpc.AspNetCore.Server.GrpcClient
         /// A flag that indicates whether the request deadline should be propagated to client calls. Defaults to true.
         /// </summary>
         public bool UseRequestDeadline { get; set; } = true;
+
+        // This property is set internally. It is used to check whether named configuration was explicitly set by the user
+        internal bool ExplicitlySet { get; set; }
     }
 }
--- a/src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientOptions.cs
+++ b/src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientOptions.cs
@@ -1,0 +1,32 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System;
+using System.Security.Cryptography.X509Certificates;
+using Grpc.Core;
+
+namespace Grpc.AspNetCore.Server.GrpcClient
+{
+    public class GrpcClientOptions<TClient> where TClient : ClientBase<TClient>
+    {
+        public Uri BaseAddress { get; set; }
+        public X509Certificate Certificate { get; set; }
+        public bool UseRequestCancellationToken { get; set; } = true;
+        public bool UseRequestDeadline { get; set; } = true;
+    }
+}
--- a/src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientOptions.cs
+++ b/src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientOptions.cs
@@ -18,7 +18,6 @@
 
 using System;
 using System.Security.Cryptography.X509Certificates;
-using Grpc.Core;
 
 namespace Grpc.AspNetCore.Server.GrpcClient
 {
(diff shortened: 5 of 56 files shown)


src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientServiceExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            services.TryAdd(ServiceDescriptor.Transient(typeof(ITypedHttpClientFactory<TClient>), typeof(GrpcHttpClientFactory<TClient>)));
+            services.TryAdd(ServiceDescriptor.Transient(typeof(GrpcHttpClientFactory<TClient>.Cache), typeof(GrpcHttpClientFactory<TClient>.Cache)));
+
+            services.Configure(configureClient);
+
+            // Accessing the client options here allows the HttpContextAccessor to be added as needed
+            // Is there a way to get the value from services.Configure now?
+            var clientOptions = new GrpcClientOptions<TClient>();
+            configureClient(clientOptions);
```
JamesNK: Is there a better way of doing this? I want to register the configuration with DI - `services.Configure(configureClient);` - and I want to access the configuration immediately.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientServiceExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            // HttpContextAccessor has performance overhead. Only add it when required
+            if (RequireHttpContextAccessor(clientOptions))
+            {
+                services.AddHttpContextAccessor();
```
JamesNK: I have heard that `AddHttpContextAccessor` introduces overhead to each request so I made it conditionally added depending on whether the client will use values from the current request. Is this a true fact?

src/Grpc.AspNetCore.Server.GrpcClient/GrpcHttpClientFactory.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            var callInvoker = new HttpClientCallInvoker(httpClient);
+            if (_clientOptions.UseRequestCancellationToken)
+            {
+                callInvoker.CancellationToken = _httpContextAccessor.HttpContext.RequestAborted;
+            }
+
+            // TODO(JamesNK): Need to set deadline and context propagation token
+            // Either add HttpContextServerCallContext to HttpContext.Items or provide equivilent of IHttpContextAccessor
```
JamesNK: Thoughts here? After adding this comment it occurred to me that adding it as a feature to HttpContext might be a better solution.

src/Grpc.NetCore.HttpClient/GrpcClientFactory.cs
```diff
@@ -0,0 +1,72 @@
...
+using Grpc.Core;
+
+namespace Grpc.NetCore.HttpClient
+{
+    public static class GrpcClientFactory
+    {
+        public static TClient Create<TClient>(string baseAddress, X509Certificate certificate = null) where TClient : ClientBase<TClient>
+        {
+            // Needs to be set before creating the HttpClientHandler
+            AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2Support", true);
```
JamesNK: Will we always need this? Will it have an affect on other HttpClient instances in the application?

test/FunctionalTests/Infrastructure/GrpcTestFixture.cs
```diff
@@ -47,6 +47,8 @@ public GrpcTestFixture()
                 services.AddSingleton<DynamicGrpcServiceRegistry>();
 
                 services.AddSingleton<ILoggerFactory>(LoggerFactory);
+
+                services.AddSingleton<IPrimaryMessageHandlerProvider, TestPrimaryMessageHandlerProvider>(s => new TestPrimaryMessageHandlerProvider(_server));
```
JamesNK: Injecting the TestServer into the TestServer

image

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientServiceExtensions.cs
rynowak: No there really isn't a way to do this, I'd recommend against running the user code here if possible. What's the core concern? 

I have a theory that this is about capturing values from the current request to use in header propagation? Using the context accessor for this is problematic because nothing enforces the constraint that you're "in a request" or "on the request processing thread.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientServiceExtensions.cs
rynowak: It is a true fact. I want to know more about what this is for.

davidfowl: Yes there are also thread safety concerns. This is about cancellation propagation.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientServiceExtensions.cs
JamesNK: I wanted to access the config value here so the value could be used to decided whether AddHttpContextAccessor was added. If you want to propagate something from the current request then `AddHttpContextAccessor` would be called, otherwise no. Make it pay for play.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientServiceExtensions.cs
JamesNK: Microservices tend to call other microservices. A gRPC client using the current requests cancellation token will allow its calls to be cancelled when the current request is cancelled.

Cancellation token, deadline and context propagation token are values that could be set from the current request.

examples/Clients/Chatter/Program.cs
```diff
@@ -39,10 +43,8 @@ static async Task<int> Main(string[] args)
             var name = args[0];
 
             // Server will only support Https on Windows and Linux
-            var credentials = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? ChannelCredentials.Insecure : ClientResources.SslCredentials;
-
-            var channel = new Channel("localhost:50051", credentials);
-            var client = new Chatter.ChatterClient(channel);
+            var certificate = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? null : new X509Certificate2(Path.Combine(Resources.CertDir, "client.crt"));
+            var client = GrpcClientFactory.Create<Chatter.ChatterClient>(@"https://localhost:50051", certificate);
```
jtattermusch: Does it make sense to replace existing examples (that are fully working) with examples that use Grpc.NetCore.HttpClient and miss important functionality (trailers)?
It sounds more like we should fork one of the clients and create a separate example for Grpc.NetCore.HttpClient.

Same for other examples.

jtattermusch: In most gRPC languages, there's a concept of Channel (something heavyweight, roughly corresponding to a connection even though this parallel is not accurate due to channel laziness and e.g. loadbalancing) and a lightweight "stub" (the client class for given service).
We should consider exposing such concepts too if possible (Channel would be replaced by something that implements a CallInvoker).

src/Grpc.AspNetCore.Server.GrpcClient/GrpcHttpClientFactory.cs
rynowak: feature is the most first-class thing. Also see: https://github.com/aspnet/AspNetCore/pull/7921

src/Grpc.AspNetCore.Server.GrpcClient/GrpcHttpClientFactory.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            return (TClient)_cache.Activator(_services, new object[] { callInvoker });
+        }
+
+        // The Cache should be registered as a singleton, so it that it can
+        // act as a cache for the Activator. This allows the outer class to be registered
+        // as a transient, so that it doesn't close over the application root service provider.
+        public class Cache
```
rynowak: workarounds, amirite?

src/Grpc.AspNetCore.Server.GrpcClient/Grpc.AspNetCore.Server.GrpcClient.csproj
```diff
@@ -0,0 +1,15 @@
...
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+
+    <ProjectReference Include="..\Grpc.NetCore.HttpClient\Grpc.NetCore.HttpClient.csproj" />
+    <PackageReference Include="Grpc.Core" Version="$(GrpcCorePackageVersion)" />
+
+    <PackageReference Include="Microsoft.Extensions.Http" Version="3.0.0-preview.19074.2" />
```
rynowak: Is this just temporary? This should be part of the framework reference.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+using Grpc.Core;
+
+namespace Grpc.AspNetCore.Server.GrpcClient
+{
+    public class GrpcClientOptions<TClient> where TClient : ClientBase<TClient>
+    {
+        public Uri BaseAddress { get; set; }
+        public X509Certificate Certificate { get; set; }
+        public bool UseRequestCancellationToken { get; set; } = true;
+        public bool UseRequestDeadline { get; set; } = true;
```
rynowak: nit: I find it much less surprising for bool options to default to `false`. It's less confusing (IMO) from the point of view of the app developer.

```
options.SupressRequestCancellationToken = true;
```

Ignore me if this project is already using a different style.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientServiceExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+using Grpc.Core;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Http;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class GrpcClientServiceExtensions
+    {
+        public static IHttpClientBuilder AddGrpcClient<TClient>(this IServiceCollection services, Action<GrpcClientOptions<TClient>> configureClient) where TClient : ClientBase<TClient>
```
rynowak: You might want another overload that takes a `string name`. Typed clients use `nameof(TClient)` as the identifier for the client, so if you need to register multiple types with the same class name, or register the same class name multiple times you will be in trouble.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientServiceExtensions.cs
rynowak: Considering that the things you check for are the defaults, do you really need this?

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientServiceExtensions.cs
JamesNK: How do named typed clients work with constructor DI injection?

I'm going to guess that the named configs are usable if IHttpClientFactory is injected into a class and you create an instance from the factory.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcHttpClientFactory.cs
rynowak: If you do end up doing something like this, you need to capture the values up front to avoid safety problems.

src/Grpc.NetCore.HttpClient/GrpcClientFactory.cs
rynowak: Yes, these are global. I imagine you won't always need it.

src/Grpc.NetCore.HttpClient/Grpc.NetCore.HttpClient.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Authors>The gRPC Authors</Authors>
+    <Copyright>Copyright 2019 The gRPC Authors</Copyright>
+    <Description>gRPC support for ASP.NET Core</Description>
```
JunTaoLuo: Managed .NET client for gRPC

src/Grpc.NetCore.HttpClient/GrpcClientFactory.cs
JunTaoLuo: This will be needed until the BCL turn this on by default. I don't think that's happening until at least the trailer support is available. cc @shirhatti.

src/Grpc.NetCore.HttpClient/HttpClientCallInvoker.cs
```diff
@@ -0,0 +1,400 @@
...
+                default:
+                    // length%4 == 1 should be illegal
+                    throw new FormatException("Invalid base64 header value");
+            }
+
+            return Convert.FromBase64String(decodable);
+        }
+    }
+
+    internal class PipeClientStreamWriter<TRequest> : IClientStreamWriter<TRequest>
```
JunTaoLuo: These files need to be split up.

src/Grpc.AspNetCore.Server.GrpcClient/Grpc.AspNetCore.Server.GrpcClient.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
JamesNK: It has a dependency on the managed gRPC client. We don't nessessarily want the server to always bring in the client.

This package glues the two together. I think it would be something the metadata would bring in.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientOptions.cs
JamesNK: Sure, something to think about. Need to figure out what the default behavour should be before locking down on the names.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcHttpClientFactory.cs
JamesNK: I just copied and pasted the default implementation and modified as needed. I assume you know what you're doing here 😄

examples/Clients/Chatter/Program.cs
JamesNK: Figuring out this probably extends beyond the scope of this PR.

JamesNK: I've reverted the clients to use the c-core client.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcHttpClientFactory.cs
JamesNK: Ok! Will figure it out when the todo is looked at.

JamesNK:
🆙 📅 

I have reverted the example clients to use the c-core client. All changes are in two new projects. I'd like to get this in, and we can production-ize the managed gRPC client in later PRs.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcHttpClientFactory.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public TClient CreateClient(HttpClient httpClient)
+        {
+            if (httpClient == null)
+            {
+                throw new ArgumentNullException(nameof(httpClient));
+            }
+
+            var callInvoker = new HttpClientCallInvoker(httpClient);
+            if (_clientOptions.UseRequestCancellationToken)
```
JamesNK: @rynowak I need to access named `GrpcClientOptions` options here but I don't have the name. Is this a limitation of `ITypedHttpClientFactory<TClient>`?

JamesNK:
🆙 📅 

* Added named client configuration
* Added `IGrpcClientFactory` for resolving clients by name
* Added `IServerCallContextFeature` for accessing context from `HttpContext`
* Request deadline/cancellation token passed to client
* Unit tests

This (the HttpClientFactory integration) is feature complete and has tests. I'd like to get this in so please review 🙏

src/Grpc.NetCore.HttpClient/Grpc.NetCore.HttpClient.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
jtattermusch: Should we put a README.md under Grpc.NetCore.Http client that this is not expected to work yet?

Status and Trailers are not supported yet, so calls will always return StatusCode.OK, so (unlike Grpc.AspNetCore.Server) the managed client is not really usable yet. We should potential early adopters about this.

jtattermusch:
travis tests seem stuck, not sure why. Rebase to nudge travis a bit?

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>
+    /// Options used to configure a gRPC client.
+    /// </summary>
+    public class GrpcClientOptions
+    {
+        /// <summary>
+        /// The base address to use when making gRPC requests.
+        /// </summary>
+        public Uri BaseAddress { get; set; }
```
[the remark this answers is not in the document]
JamesNK: Yes you could. I like `BaseAddress` because it matches what developers are fimilar with from `HttpClient.BaseAddress`

src/Grpc.AspNetCore.Server.GrpcClient/TypeNameHelper.cs
```diff
@@ -0,0 +1,195 @@
...
+
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Grpc.AspNetCore.Server.GrpcClient
+{
+    // Copied from https://github.com/aspnet/Extensions/blob/aa18b93adac411b6b98e4088284a1ffb7d83249f/src/Shared/src/TypeNameHelper/TypeNameHelper.cs
```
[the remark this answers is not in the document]
JamesNK: The history of this code is it is entirely written by Microsoft developers. We're contributing it to Grpc like any other code we are writing for the project.

If you still have concerns then we can chat.

src/Grpc.NetCore.HttpClient/ClientAsyncStreamReader.cs
```diff
@@ -0,0 +1,66 @@
...
+                    var responseMessage = await _sendTask;
+                    _responseStream = await responseMessage.Content.ReadAsStreamAsync();
+                }
+
+                Current = _responseStream.ReadSingleMessage(_deserializer);
+                return true;
+            }
+            catch (Exception ex)
+            {
+                ex.StackTrace.ToString();
```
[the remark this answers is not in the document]
JamesNK: Whoops, old code from trying to debug the client. Will remove.

src/Grpc.NetCore.HttpClient/Grpc.NetCore.HttpClient.csproj
JamesNK: Good idea

build/dependencies.props
```diff
@@ -9,9 +9,10 @@
     <MicrosoftAspNetCoreTestHostPackageVersion>3.0.0-preview-18579-0056</MicrosoftAspNetCoreTestHostPackageVersion>
     <MicrosoftExtensionsLoggingTestingPackageVersion>2.0.0-*</MicrosoftExtensionsLoggingTestingPackageVersion>
     <MicrosoftNETTestSdkPackageVersion>16.0.0-preview-20181205-02</MicrosoftNETTestSdkPackageVersion>
+    <MicrosoftSourceLinkGitHubPackageVersion>1.0.0-beta2-18618-05</MicrosoftSourceLinkGitHubPackageVersion>
     <MoqPackageVersion>4.10.0</MoqPackageVersion>
     <NunitPackageVersion>3.11.0</NunitPackageVersion>
     <Nunit3TestAdapterPackageVersion>3.12.0</Nunit3TestAdapterPackageVersion>
-    <MicrosoftSourceLinkGitHubPackageVersion>1.0.0-beta2-18618-05</MicrosoftSourceLinkGitHubPackageVersion>
+    <SystemIOPipelinesPackageVersion>4.6.0-preview.19073.11</SystemIOPipelinesPackageVersion>
```
JunTaoLuo: 4.6.0-preview3.19128.7 looks like the newest version.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcClientServiceExtensions.cs
```diff
@@ -0,0 +1,192 @@
...
+                var httpClient = httpClientFactory.CreateClient(builder.Name);
+
+                var typedClientFactory = s.GetRequiredService<INamedTypedHttpClientFactory<TClient>>();
+                return typedClientFactory.CreateClient(httpClient, builder.Name);
+            });
+
+            return builder;
+        }
+
+        internal class DefaultHttpClientBuilder : IHttpClientBuilder
```
JunTaoLuo: nit: private class, or new files

src/Grpc.AspNetCore.Server.GrpcClient/GrpcHttpClientFactory.cs
```diff
@@ -0,0 +1,115 @@
...
+            if (namedOptions.PropagateCancellationToken)
+            {
+                if (serverCallContext == null)
+                {
+                    throw new InvalidOperationException("Cannot propagate the call cancellation token to the client. Cannot find the current gRPC ServerCallContext.");
+                }
+
+                callInvoker.CancellationToken = serverCallContext.CancellationToken;
+            }
+            if (namedOptions.PropagateDeadline)
```
JunTaoLuo: nit: new line after brace

src/Grpc.AspNetCore.Server.GrpcClient/GrpcHttpClientFactory.cs
```diff
@@ -0,0 +1,115 @@
...
+                callInvoker.Deadline = serverCallContext.Deadline;
+            }
+
+            return (TClient)_cache.Activator(_services, new object[] { callInvoker });
+        }
+
+        // The Cache should be registered as a singleton, so it that it can
+        // act as a cache for the Activator. This allows the outer class to be registered
+        // as a transient, so that it doesn't close over the application root service provider.
+        public class Cache
```
JunTaoLuo: Why is this public?

src/Grpc.AspNetCore.Server.GrpcClient/INamedTypedHttpClientFactory.cs
```diff
@@ -0,0 +1,27 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System.Net.Http;
+
+namespace Grpc.AspNetCore.Server.GrpcClient
```
JunTaoLuo: Internal namespace? Also, move to internal directory?

src/Grpc.AspNetCore.Server.GrpcClient/TypeNameHelper.cs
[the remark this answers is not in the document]
jtattermusch: Got that, my concern something along the line that automated license checkers won't know that and might evaluate as false positive.

build/dependencies.props
JamesNK: Pipelines will eventually be removed from the client so not important.

src/Grpc.AspNetCore.Server.GrpcClient/GrpcHttpClientFactory.cs
JamesNK: It's nested in an internal class so it becomes internal anyway.

src/Grpc.AspNetCore.Server.GrpcClient/DefaultGrpcClientFactory.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using Grpc.Core;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using System;
+using System.Net.Http;
+
+namespace Grpc.AspNetCore.Server.GrpcClient
+{
+    internal class DefaultGrpcClientFactory : IGrpcClientFactory
```
rynowak: I was going to complain about this because of parallel abstractions - but I think it seems fine since a Grpc client already is a different type from httpclient.

src/Grpc.AspNetCore.Server.GrpcClient/IGrpcClientFactory.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using Grpc.Core;
+
+namespace Grpc.AspNetCore.Server.GrpcClient
+{
+    /// <summary>
+    /// A factory abstraction for a component that can create gRPC client instances with custom
+    /// configuration for a given logical name.
+    /// </summary>
+    public interface IGrpcClientFactory
```
rynowak: Consider seriously making this an abstract class. We should have done with `IHttpClientFactory` .

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
```diff
@@ -18,6 +18,7 @@
 
 using System;
 using System.Threading.Tasks;
+using Grpc.AspNetCore.Server.Feature;
```
rynowak: Is there a really good reason to create a namespace for this?

JamesNK: Nothing other than features everywhere else live in a Features namespace.

I'm going to leave it here for now and we can change it in an API review closer to release.

rynowak: IMO this naming doesn't make sense in our code, but is excusable because we have lots of features. I can't imagine grpc having too many, but maybe I need to adopt a growth mindset!

PR #147, opened 2019-03-07
Improving the auth tests and and a new auth test with an invalid token

damienbod:
Added a new test for an invalid token
Use the Authorization header to send the Bearer token

### DIFF
--- a/test/FunctionalTests/AuthorizationTests.cs
+++ b/test/FunctionalTests/AuthorizationTests.cs
@@ -89,7 +89,7 @@ namespace Grpc.AspNetCore.FunctionalTests
         }
 
         [Test]
-        public async Task CallAuthorizedServiceWithInvalidToken_Success()
+        public async Task CallAuthorizedServiceWithInvalidToken_ReturnUnauthorized()
         {
             // Arrange
 
--- a/test/FunctionalTests/AuthorizationTests.cs
+++ b/test/FunctionalTests/AuthorizationTests.cs
@@ -71,7 +71,7 @@ namespace Grpc.AspNetCore.FunctionalTests
             MessageHelpers.WriteMessage(ms, requestMessage);
 
             var httpRequest = new HttpRequestMessage(HttpMethod.Post, "Authorize.AuthorizedGreeter/SayHello");
-            httpRequest.Headers.Add("access_token", token);
+            httpRequest.Headers.Add("Authorization", $"Bearer {token}");
             httpRequest.Content = new GrpcStreamContent(ms);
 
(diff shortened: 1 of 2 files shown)


thelinuxfoundation:

* User @damienbod isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

damienbod:
I have signed up for the cla/linuxfoundation and completed the form

test/FunctionalTests/AuthorizationTests.cs
```diff
@@ -87,5 +87,29 @@ public async Task CallAuthorizedServiceWithToken_Success()
 
             Assert.AreEqual(StatusCode.OK.ToTrailerString(), Fixture.TrailersContainer.Trailers[GrpcProtocolConstants.StatusTrailer].Single());
         }
+
+        [Test]
+        public async Task CallAuthorizedServiceWithInvalidToken_Success()
```
JamesNK:
```suggestion
        public async Task CallAuthorizedServiceWithInvalidToken_ReturnUnauthorized()
```

damienbod: updated, thanks

test/FunctionalTests/AuthorizationTests.cs
```diff
@@ -87,5 +87,29 @@ public async Task CallAuthorizedServiceWithToken_Success()
...
+
+            var httpRequest = new HttpRequestMessage(HttpMethod.Post, "Authorize.AuthorizedGreeter/SayHello");
+            httpRequest.Headers.Add("Authorization", $"Bearer SomeInvalidTokenHere");
+            httpRequest.Content = new GrpcStreamContent(ms);
+
+            // Act
+            var response = await Fixture.Client.SendAsync(httpRequest);
+
+            // Assert
+            Assert.AreEqual(HttpStatusCode.Unauthorized, response.StatusCode);
```
jtattermusch: a thought:  gRPC also has StatusCode.Unauthenticated (which is sent along with http status code 200), so the .NET Core built-in support for [Authorize]  doesn't quite behave as expected with gRPC. I'm not quite sure how to solve at this point, but we should discuss this in the sync meeting.
Would having a [GrpcAuthorize] attribute make sense?

JamesNK: The attribute only specifies the policy name.

Is it important to return a gRPC status instead of 401? If so create an issue and we can discuss what we can do to improve it.

PR #153, opened 2019-03-11
Add GrpcBindingOptions.BasePath

JamesNK:
Addresses #110

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
@@ -38,16 +38,23 @@ namespace Grpc.AspNetCore.Server.Internal
         private readonly ServiceMethodsRegistry _serviceMethodsRegistry;
         private readonly ServerCallHandlerFactory<TService> _serverCallHandlerFactory;
         private readonly IGrpcMethodModelFactory<TService> _serviceModelFactory;
+        private readonly string _basePath;
 
         internal IList<IEndpointConventionBuilder> EndpointConventionBuilders { get; } = new List<IEndpointConventionBuilder>();
         internal IList<IMethod> ServiceMethods { get; } = new List<IMethod>();
 
-        internal GrpcServiceBinder(IEndpointRouteBuilder builder, IGrpcMethodModelFactory<TService> serviceModelFactory, ServerCallHandlerFactory<TService> serverCallHandlerFactory, ServiceMethodsRegistry serviceMethodsRegistry)
+        internal GrpcServiceBinder(
+            IEndpointRouteBuilder builder,
+            IGrpcMethodModelFactory<TService> serviceModelFactory,
+            string basePath,
+            ServerCallHandlerFactory<TService> serverCallHandlerFactory,
+            ServiceMethodsRegistry serviceMethodsRegistry)
         {
             _builder = builder;
             _serviceMethodsRegistry = serviceMethodsRegistry;
             _serverCallHandlerFactory = serverCallHandlerFactory;
             _serviceModelFactory = serviceModelFactory;
+            _basePath = basePath;
         }
 
         public override void AddMethod<TRequest, TResponse>(Method<TRequest, TResponse> method, ClientStreamingServerMethod<TRequest, TResponse> handler)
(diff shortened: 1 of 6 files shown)


JamesNK:
This is configured per registered service.

Pros: can configure services separately if desired
Cons: if you want all your services under one base path then it needs to be repeated for each service

src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
```diff
@@ -82,13 +89,24 @@ private void AddMethodCore(IMethod method, RequestDelegate requestDelegate, List
...
 
             // IMethod is added as metadata for the endpoint
             resolvedMetadata.Add(method);
             resolvedMetadata.AddRange(metadata);
 
-            EndpointConventionBuilders.Add(_builder.MapPost(method.FullName, $"gRPC - {method.FullName}", requestDelegate, resolvedMetadata.ToArray()));
+            EndpointConventionBuilders.Add(_builder.MapPost(pattern, $"gRPC - {pattern}", requestDelegate, resolvedMetadata.ToArray()));
+        }
+
+        private static string Combine(string uri1, string uri2)
```
JunTaoLuo: This looks like `PathString` logic.

jtattermusch:
It looks like adding this feature will decrease compatibility with gRPC clients in other languages (see https://github.com/grpc/grpc/issues/14900 and comments in #153) and I don't think we want to support this option unless there are very compelling reasons why we need to support it.

An alternative would be to support/investigate overriding the ":authority" header instead (in Grpc.Core and C core this is supported as setting the "host" parameter for clients e.g. https://github.com/grpc/grpc/blob/ee500f1963b5ed070d05dde7904b46af92d80de5/src/csharp/Grpc.Core/ClientBase.cs#L72  and https://github.com/grpc/grpc/blob/ee500f1963b5ed070d05dde7904b46af92d80de5/include/grpc/grpc.h#L217) - which seems to be more aligned with what other gRPC implementations support.

JamesNK:
> I don't think we want to support this option unless there are very compelling reasons why we need to support it.

Ok. I'll close this for now. If users raise this more in the future we can re-examin.

MontyGvMC:
I would like to support the idea of having a configurable base path.

Let me try to explain the current use case I am facing.

For customers we are developing a plug-in based application. All plug-ins runs on the same .NET5 server (same Kestrell instance). On the current project there are 10 plug-ins. Each plug-in should run the same gRPC Service.

This means the gRPC service has the same operations and classes. We implemented it once as library (with built in logging etc.)  as abstract generic service base. Each gRPC service of the plug-in inherits from the generic abstract base class and ensures the corrent generic type gets injected into our generic base class.

Now we have the need to start the 10 inherited services as 10 different endpoints. Which fails because it seems there is no way to change/modify the URL and it takes the same path/service name for each inherited gRPC service.

A basePath would allow a result like (pattern: Host/BasePath/Service) 

```
LOCALHOST/PluginA/grpc.service/DoStuff
LOCALHOST/PluginB/grpc.service/DoStuff
LOCALHOST/PluginC/grpc.service/DoStuff
...
```

even better (for our case) would be to change the service name/path resulting in

```
LOCALHOST/PluginAService/DoStuff
LOCALHOST/PluginBService/DoStuff
LOCALHOST/PluginCService/DoStuff
...
```

thorenled:
> I would like to support the idea of having a configurable base path.
> 
> Let me try to explain the current use case I am facing.
> 
> For customers we are developing a plug-in based application. All plug-ins runs on the same .NET5 server (same Kestrell instance). On the current project there are 10 plug-ins. Each plug-in should run the same gRPC Service.
> 
> This means the gRPC service has the same operations and classes. We implemented it once as library (with built in logging etc.) as abstract generic service base. Each gRPC service of the plug-in inherits from the generic abstract base class and ensures the correct generic type gets injected into our generic base class (and implements some details we were not able to implement in the base).
> 
> Now we have the need to start the 10 inherited services as 10 different endpoints. Which fails because it seems there is no way to change/modify the URL and it takes the same path/service name for each inherited gRPC service.
> 
> A basePath would allow a result like (pattern: Host/BasePath/Service)
> 
> ```
> LOCALHOST/PluginA/grpc.service/DoStuff
> LOCALHOST/PluginB/grpc.service/DoStuff
> LOCALHOST/PluginC/grpc.service/DoStuff
> ...
> ```
> 
> even better (for our case) would be to change the service name/path resulting in
> 
> ```
> LOCALHOST/PluginAService/DoStuff
> LOCALHOST/PluginBService/DoStuff
> LOCALHOST/PluginCService/DoStuff
> ...
> ```
> 
> For those interested I have matching [question on StackOverflow](https://stackoverflow.com/questions/70334178/change-route-of-mapped-grpc-service-while-adding-endpoint).

That would be awesome to have this feature!

shushil:
Hello guys we are facing similar issue as @MontyGvMC has described above. We are also trying to upgrade move our code from WCF to gRPC but we are facing same scenario as @MontyGvMC is facing. 

@MontyGvMC would mind sharing which path did you took to solve your issue.

Issue #149, opened 2019-03-07
large_unary interop test timing out against most grpc clients


jtattermusch:
https://source.cloud.google.com/results/invocations/83cbf9c2-cf9c-48e6-9865-74f60f96a421/targets/github%2Fgrpc%2Finterop_test/tests;group=tests;test=cloud_to_cloud:c%2B%2B:aspnetcore_server:custom_metadata:tls;row=1

Not sure yet what's going on, but it can be reproduced with instructions here
https://github.com/grpc/grpc-dotnet/tree/master/testassets/InteropTestsWebsite

jtattermusch:
It looks like `ReadSingleMessageAsync` get stuck forever when reading a large unary request.

jtattermusch:
CC @JunTaoLuo

jtattermusch:
What I'm seeing is that ReadAsync() reads 98304 bytes in the first invocation and never returns from the second invocation.

jtattermusch:
I bisected the breakage down to 
https://github.com/grpc/grpc-dotnet/commit/851e23101dd472703fd7e6caa91aeffa13f041ec (which is basically just upgrade to preview 3)

So there are two options:
- either the usage pattern HttpContext's PipeReader of preview3 has changed
- there's a bug in preview3 that breaks our code

```
851e23101dd472703fd7e6caa91aeffa13f041ec is the first bad commit
commit 851e23101dd472703fd7e6caa91aeffa13f041ec
Author: Jan Tattermusch <jtattermusch@users.noreply.github.com>
Date:   Thu Mar 7 09:36:18 2019 +0100

    Use .NET Core 3 preview 3 (#146)
    
    * bump to .NET Core 3 preview 3
    
    * use Grpc 1.19.0 nugets
    
    * nits in test runner scripts

:040000 040000 1ceea3b717fa44ccdf2d95fda7a259dfcf40d0dd ee9ed0e3caf0f016f2a218fdc9ed1d3c2c2f8e9b M	build
:100755 100755 8ae1db70c98764cbd9d080abb72b127c194bcc8e dcb1403eed4f7e3401925becf1942436686c2a73 M	build_and_test.sh
:100644 100644 f660a138ee650ca393d9a591c37cc168de911ddc 433a84f36505e22f8a8c2468b9f1524ac7cf906c M	global.json
[...]

PR #154, opened 2019-03-11
Tweak window sizes to make interop tests pass

jtattermusch:
Fixes https://github.com/grpc/grpc-dotnet/issues/149.

I don't like this "fix" at all, because it just masks the fact that Grpc.AspNetCore.Server doesn't work with default params when client sends large messages. The point of "large_unary" test is to test that flow control works as expected (which it currently doesn't by default).

Can we come up with a better fix until https://github.com/aspnet/AspNetCore/pull/8200 is available?

### DIFF
--- a/testassets/InteropTestsWebsite/Program.cs
+++ b/testassets/InteropTestsWebsite/Program.cs
@@ -49,6 +49,12 @@ namespace InteropTestsWebsite
                         }
                         listenOptions.Protocols = HttpProtocols.Http2;
                     });
+
+                    // Tweak flow control options to make large_unary test pass.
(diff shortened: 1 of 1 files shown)


JunTaoLuo:
The fix in  aspnet/AspNetCore#8200 will be available in preview 4 so I'm not too concerned with finding an alternative.

jtattermusch:
@JunTaoLuo  what's the ETA for preview4?

JunTaoLuo:
Our preview cadence is about once a month and we just released preview 3 so I'd assume around middle of April.

PR #156, opened 2019-03-12
Add functional test for singleton scenario

JunTaoLuo:
Added a functional test for singleton services to ensure the scenario doesn't regress in the future. The unit tests in 
https://github.com/grpc/grpc-dotnet/blob/master/test/Grpc.AspNetCore.Server.Tests/DefaultGrpcServiceActivatorTests.cs#L48-L77 
already cover the unit test scenarios pretty well.

I also cleaned up the tests a little bit while I was writing the new one.

### DIFF
--- a/test/FunctionalTests/UnaryMethodTests.cs
+++ b/test/FunctionalTests/UnaryMethodTests.cs
@@ -23,8 +23,7 @@ using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
-using FunctionalTestsWebsite;
-using FunctionalTestsWebsite.Services;
+using FunctionalTestsWebsite. Services;
 using Google.Protobuf.WellKnownTypes;
 using Greet;
 using Grpc.AspNetCore.FunctionalTests.Infrastructure;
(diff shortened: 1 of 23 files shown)


test/FunctionalTests/UnaryMethodTests.cs
```diff
@@ -342,6 +336,34 @@ static Task<Empty> ReturnContextInfoInTrailers(Empty request, ServerCallContext
             Assert.AreEqual("localhost", Fixture.TrailersContainer.Trailers["Test-Host"].ToString());
         }
 
+        [Test]
+        public async Task SingletonService_PrivateFieldsPreservedBetweenCalls()
```
JunTaoLuo: Here is the actual new test, rest is refactoring.

testassets/FunctionalTestsWebsite/Services/SingletonCounterService.cs
```diff
@@ -0,0 +1,50 @@
...
+using System;
+using System.Threading.Tasks;
+using Google.Protobuf.WellKnownTypes;
+using Grpc.Core;
+using Microsoft.Extensions.Logging;
+using SingletonCount;
+
+namespace FunctionalTestsWebsite.Services
+{
+    public class SingletonCounterService : Counter.CounterBase, IDisposable
```
JunTaoLuo: And here's the refactoring.

testassets/FunctionalTestsWebsite/Services/SingletonCounterService.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public SingletonCounterService(ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger<CounterService>();
+        }
+
+        public void Dispose()
+        {
+            // Set the count value to some arbitrary value. This should never happen in any case.
+            _count = int.MinValue;
```
JamesNK: This isn't be calling each request which is good, but what about on per request services?

What does MVC do with controller's the implement IDisposable?

JunTaoLuo: Yes, that was https://github.com/grpc/grpc-dotnet/pull/91/files.

JamesNK: Cool beans.

I am finally happy that this is a A Thing That Works 😄

Issue #143, opened 2019-03-06
React to BodyPipe rename


JunTaoLuo:
The BodyPipe properties were renamed in https://github.com/aspnet/AspNetCore/pull/8227. We'll need to update to the new API when we ingest new ASP.NET Core versions.

Issue #144, opened 2019-03-06
React to Response.StartAsync change


JamesNK:
StartAsync behavior is changing with https://github.com/aspnet/AspNetCore/issues/7779

Delay calling StartAsync when buffering on the server so user can send HTTP headers afterwards.

davidfowl:
Well no, if you do that then it'll be inefficient. You want to keep calling StartAsync so that on kestrel, it'll be optimized.

JamesNK:
@jtattermusch What is the current behavior of c-core when it comes to buffering? I wanted to test out whether it prevents headers from being sent but in the following code it hangs on the first `await WriteAsync(message)`.

```cs
public override async Task SayHellos(HelloRequest request, IServerStreamWriter<HelloReply> responseStream, ServerCallContext context)
{
    responseStream.WriteOptions = new WriteOptions(WriteFlags.BufferHint);

    for (int i = 0; i < 3; i++)
    {
        var message = $"How are you {request.Name}? {i}";
        await responseStream.WriteAsync(new HelloReply
        {
            Message = message,
            Timestamp = Timestamp.FromDateTime(DateTime.UtcNow)
        });
    }

    await responseStream.WriteAsync(new HelloReply
    {
        Message = $"Goodbye {request.Name}!",
        Timestamp = Timestamp.FromDateTime(DateTime.UtcNow)
    });

    var metadata = new Metadata();
    metadata.Add("test", "value!");

    await context.WriteResponseHeadersAsync(metadata);
}
```

jkotalik:
@JamesNK this was merged just now.

jtattermusch:
> @jtattermusch What is the current behavior of c-core when it comes to buffering? I wanted to test out whether it prevents headers from being sent but in the following code it hangs on the first `await WriteAsync(message)`.
> 
> ```cs
> public override async Task SayHellos(HelloRequest request, IServerStreamWriter<HelloReply> responseStream, ServerCallContext context)
> {
>     responseStream.WriteOptions = new WriteOptions(WriteFlags.BufferHint);
> 
>     for (int i = 0; i < 3; i++)
>     {
>         var message = $"How are you {request.Name}? {i}";
>         await responseStream.WriteAsync(new HelloReply
>         {
>             Message = message,
>             Timestamp = Timestamp.FromDateTime(DateTime.UtcNow)
>         });
>     }
> 
>     await responseStream.WriteAsync(new HelloReply
>     {
[...]

PR #162, opened 2019-03-13
React to pipe renames

JunTaoLuo:
Addresses https://github.com/grpc/grpc-dotnet/issues/143. Vendors filed an issue so I'm creating the update.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "3.0.100-preview3-010431"
+    "version": "3.0.100-preview4-010737"
   }
 }
--- a/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
@@ -68,8 +68,8 @@ namespace Grpc.AspNetCore.Server.Internal.CallHandlers
                     throw new RpcException(new Status(StatusCode.Cancelled, "No message returned from method."));
                 }
 
-                var responseBodyPipe = httpContext.Response.BodyPipe;
-                await responseBodyPipe.WriteMessageAsync(response, serverCallContext, Method.ResponseMarshaller.Serializer);
+                var responseBodyWriter = httpContext.Response.BodyWriter;
+                await responseBodyWriter.WriteMessageAsync(response, serverCallContext, Method.ResponseMarshaller.Serializer);
             }
             catch (Exception ex)
             {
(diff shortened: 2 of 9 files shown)


JamesNK:
What about the request pipe? e.g. https://github.com/grpc/grpc-dotnet/blob/4e9f76586f6cca46a5a1f38d3d9823458e2fd568/src/Grpc.AspNetCore.Server/Internal/HttpContextStreamReader.cs#L56 isn't this changing to PipeReader?

JunTaoLuo:
Oops, I think I searched one file instead of the whole solution. Ran the tests now to ensure I didn't miss anything else. I don't think there's scripts that need updating since we run get-dotnet.sh already on our travis runs. I've updated the global.json which ensures the builds get the correct SDK.

global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "3.0.100-preview3-010431"
+    "version": "3.0.100-preview4-010737"
```
jtattermusch: Is this a nightly release? if merged as is, bumping would break the continuous interop test build.

jtattermusch: We should figure out how to do testing against the latest released preview and against the nightly build at the same time.
Should we have two builds - one against the latest nightly and one against the latest released preview and test against both (and use #ifdefs to make things buildable). Then, once the next preview is out, we can cleanup the #ifdef sites?

Btw, would that nightly have https://github.com/aspnet/AspNetCore/pull/8200?

JamesNK: Yes it is a nightly release.

What is the value of testing a build with custom #ifdefs against the official preview? Is it to keep the interop test build passing because it is not possible to install nightly .NET SDKs in that environment?

jtattermusch: I think the main value is maintain the ability to do a release on nuget.org at any time (there's probably no point in releasing something that won't with the latest official preview) and users still being able to build against the latest preview.

In the interop framework we could pull the never version of SDK if needed, but we'd like to avoid if possible because none of the other gRPC languages does that - all use a fixed environment.

jtattermusch: Btw, updating the dotnet SDK version in out performance benchmark harness is harder than updating for interop framework - which is yet another motivation to keep things buildable with the stable preview.

JamesNK: If we need to make a point release I think it should come from a branch from the 0.1.9 release. I'd rather not try to maintain two versions because we're going to have more and more minor breaking changes from the SDK:
* this rename
* https://github.com/grpc/grpc-dotnet/issues/164
* https://github.com/grpc/grpc-dotnet/issues/144

I don't think we should automatically use the latest nightly each build, but we are going to need to manually update to a newer nightly from time to time. This isn't really any different from manually updating to a newer official preview SDK.

JunTaoLuo: Updating should be as simple as running the build/get-dotnet.{ps1/sh} script.

JunTaoLuo:
The grpc PR is merged. I'll merge this as well.

PR #169, opened 2019-03-18
Add more kokoro job configurations

jtattermusch:
Prepare setup to be able to run more tests on PRs (nuget build, interop tests etc.).

### DIFF
--- a/kokoro/interop.sh
+++ b/kokoro/interop.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+# Copyright 2019 The gRPC Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


kokoro/interop.sh
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+# change to grpc repo root
+cd $(dirname $0)/..
+
+# TODO(jtattermusch): run interop tests
+echo "Interop tests not yet setup."
```
JunTaoLuo: So is this a WIP PR or should this be checked in as is?

jtattermusch: It can be merged. The PR test will be created as a followup.

PR #171, opened 2019-03-18
Run interop tests against grpc-dotnet

jtattermusch:


kokoro/interop.sh
```diff
@@ -15,8 +15,14 @@
 
 set -ex
 
-# change to grpc repo root
+# change to grpc-dotnet repo root
 cd $(dirname $0)/..
 
-# TODO(jtattermusch): run interop tests
-echo "Interop tests not yet setup."
+git clone https://github.com/grpc/grpc ./../grpc
```
JunTaoLuo: Is this a pattern that's common in grpc infrastructure? I'm a little skeptical of cloning entire repos in parent directories. How about submodules if we are going that route or is there a directory structure requirement for this to work?

I'm not opposed to it but it does seem like an odd pattern.

jtattermusch: Yes, this is a pattern we are using to run other interop test suites as well. So it's basically a "standard" pattern.
Basically the testing script expects the directories with clones of language implementations to live side by side (one advantage is that makes local testing easier).

https://github.com/grpc/grpc/blob/a41a381302376488a779422e50a6c2d4bd7049ca/tools/internal_ci/helper_scripts/prepare_build_interop_rc#L27

jtattermusch:
A passing adhoc run:
https://source.cloud.google.com/results/invocations/da2b0662-370d-48e6-841f-14e4dbe89ff6/targets/github%2Fgrpc%2Finterop_test/tests

I thinks this is ready to review and merge (I'll setup PR tests as a followup).

PR #173, opened 2019-03-18
Change server streaming sample to use cancellation

JamesNK:
Update the server streaming sample to use a cancellation token from the client

### DIFF
--- a/examples/Clients/Greeter/Program.cs
+++ b/examples/Clients/Greeter/Program.cs
@@ -38,8 +38,11 @@ namespace Sample.Clients
             var reply = client.SayHello(new HelloRequest { Name = "GreeterClient" });
             Console.WriteLine("Greeting: " + reply.Message);
 
-            var replies = client.SayHellos(new HelloRequest { Name = "GreeterClient" });
-            while (await replies.ResponseStream.MoveNext(CancellationToken.None))
+            var cts = new CancellationTokenSource();
+            cts.CancelAfter(TimeSpan.FromSeconds(3.5));
+
+            var replies = client.SayHellos(new HelloRequest { Name = "GreeterClient" }, cancellationToken: cts.Token);
+            while (await MoveNextSafe(replies.ResponseStream, cts.Token))
             {
                 Console.WriteLine("Greeting: " + replies.ResponseStream.Current.Message);
             }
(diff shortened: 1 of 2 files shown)


examples/Clients/Greeter/Program.cs
```diff
@@ -38,16 +38,29 @@ static async Task Main(string[] args)
             var reply = client.SayHello(new HelloRequest { Name = "GreeterClient" });
             Console.WriteLine("Greeting: " + reply.Message);
 
-            var replies = client.SayHellos(new HelloRequest { Name = "GreeterClient" });
-            while (await replies.ResponseStream.MoveNext(CancellationToken.None))
```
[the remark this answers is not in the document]
jtattermusch: The semantics of cancellation token passed to responseStream.MoveNext() is to trigger cancellation of the entire call - hence the RpcException being thrown, because it means the call actually ended with an error. Also, because the call is being terminated, it makes sense to throw  RpcException with corresponding status (that's the same behavior you'd get from a unary call - if you cancel it, it will throw RpcException once it get cancelled).

examples/Clients/Greeter/Program.cs
```diff
@@ -38,16 +38,29 @@ static async Task Main(string[] args)
...
             }
 
             Console.WriteLine("Shutting down");
-            channel.ShutdownAsync().Wait();
+            await channel.ShutdownAsync();
             Console.WriteLine("Press any key to exit...");
             Console.ReadKey();
         }
+
+        private static async Task<bool> MoveNextSafe<T>(IAsyncStreamReader<T> reader, CancellationToken cancellationToken)
```
jtattermusch: Looks like this can possibly eat errors coming from the server side?

JamesNK: Only if MoveNext would have returned an error and the cancellation token was cancelled before the error was received.

JamesNK: Is this a good approach to handle reading from a client when streaming with a cancellation token, or would you recommend using a try/catch instead?

jtattermusch: I'd say try catch is a better option (and if you want to, you can eat the exception if StatusCode=Cancelled status and you already requested cancellation on the client side - that will make sure you won't eat an exception with an unexpected status).

Btw, thinking out loud - for scenarios where you want to tell the server to stop sending and you still want the calls to finish gracefully (Ok status), it might be actually better to use a bi-di streaming call  and calling requestStream.CompleteAsync() on the client side (half close). The server will see that (MoveNext() returns false) and can react by ending the RPC gracefully.
I think cancellation is generally meant to be used for more "unexpected" scenarios, rather than being used as part of the "protocol" between client and server.

JamesNK:
🆙 📅 

Now using try/catch

// @jtattermusch

Issue #43, opened 2019-01-23
Managed client using HttpClient


JunTaoLuo:
This issue tracks the work required to provide a complete managed end to end experience. This is going to be blocked on Core support for trailing response header support. cc @shirhatti for an expected timeframe.

Also, I'm not certain we need to change the code gen to implement this feature. I think we can pass in an HttpClient based [call invoker](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core/CallInvoker.cs) to the [constructor](https://github.com/grpc/grpc/blob/afc5b0b5d4d36aedf94a70b50d7c40062590abb1/src/csharp/Grpc.Core/ClientBase.cs#L62) of the client. I don't have a client I can prototype this with yet so I'm not 100% certain but is there any reason this wouldn't work @jtattermusch @JamesNK ?

jtattermusch:
> This issue tracks the work required to provide a complete managed end to end experience. This is going to be blocked on Core support for trailing response header support. cc @shirhatti for an expected timeframe.
> 
> Also, I'm not certain we need to change the code gen to implement this feature. I think we can pass in an HttpClient based [call invoker](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core/CallInvoker.cs) to the [constructor](https://github.com/grpc/grpc/blob/afc5b0b5d4d36aedf94a70b50d7c40062590abb1/src/csharp/Grpc.Core/ClientBase.cs#L62) of the client. I don't have a client I can prototype this with yet so I'm not 100% certain but is there any reason this wouldn't work @jtattermusch @JamesNK ?

Good point, using HttpClient-based CallInvoker to create an instance of client stub should work. That should be sufficient for basic client functionality.

JamesNK:
Initial release in preview 6. Additional features can be tracked in there own issues.

mikebm:
Sorry to resurrect this, but I am not seeing a .net standard version of the Grpc.Net.Client available in Nuget. It only targets .net core 3 which is incompatible with projects such as Xamarin Forms. Am I missing something or is this not actually available as part of preview 6.

JamesNK:
https://www.nuget.org/packages/Grpc.Net.Client/0.1.22-pre1 - this version supports netstandard2.1

netstandard2.1 is required because the client needs to access trailers from the HTTP/2 response.

PR #180, opened 2019-03-20
Add document for client migration

JunTaoLuo:
Sounded like we wanted to start capturing design considerations in docs. I've summarized our email conversations so far for moving the types required for https://github.com/grpc/grpc-dotnet/issues/43.

I think documentation may be more suited for keeping track of our final decisions after we discuss in emails but we can try doing this on a PR to a markdown file.

cc @shirhatti @JamesNK FYI

### DIFF
--- a/doc/clientTypeMigrations.md
+++ b/doc/clientTypeMigrations.md
@@ -1,0 +1,33 @@
+# Migrating managed client types to Grpc.Core.Api
+
+Migrating types for managed client will be different from types required for managed server. The main difference is that there’s a default client base implementation whose functionality must be maintained for back compat. The types `ClientBase<T>` and `ClientBase` has constructors that take in Channel which contains abstractions for making native calls using c core which is then used by the `DefaultCallInvoker`. It is difficult to decouple `ClientBase` and `Channel` while maintaining current functionality. As an alternative, it may be necessary to modify codegen to use a different base class for the client.
+
+To elaborate:
+1. Create a new client base class `ManangedClientBase<T>` which doesn’t have the constructor overload that takes in a `Channel` but is otherwise similar to `ClientBase<T>`.
+2. The constructor overload that takes in a `CallInvoker` will be used in the managed client implementation
+3. Additional types potentially needed include:
+
+* UnimplementedCallInvoker
+  * Needed for parameterless  constructor of `ManagedClientBase<T>`
+* AsyncClientStreamingCall<TResponse>
+* AsyncDuplexStreamingCall<TResponse>
+* AsyncServerStreamingCall<TResponse>
+* AsyncUnaryCall<TResponse>
+* CallInvoker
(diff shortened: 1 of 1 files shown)


doc/clientTypeMigrations.md
```diff
@@ -0,0 +1,33 @@
...
+
+* UnimplementedCallInvoker
+  * Needed for parameterless  constructor of `ManagedClientBase<T>`
+* AsyncClientStreamingCall<TResponse>
+* AsyncDuplexStreamingCall<TResponse>
+* AsyncServerStreamingCall<TResponse>
+* AsyncUnaryCall<TResponse>
+* CallInvoker
+* IClientStreamWriter<T>
+* CallFlags
```
jtattermusch: the interesting part here is the WaitForReady behavior.

doc/clientTypeMigrations.md
```diff
@@ -0,0 +1,33 @@
...
+  * Internal class which may not be needed
+* CallCredentials?
+  * This type may not be needed as it seems like it’s used to add `Authorization` headers with a per call granularity. We provide functionality through the APIs on `HttpClient` to do this but for portability reasons, though tenuous, we could move this type too. If we do decide to move this type, we need to change the implementation and remove the ToNativeCredentials() method.
```
jtattermusch: the problem is that CallOptions depend on CallCredentials (and it feels odd if the "credentials" as one of the CallOptions' constructor param would be unimplemented).
How hard would be to implement CallCredentials over HttpClient?
Another problem is that CallCredentials has some public static factory methods that transitively pull in lot of the native functionality.

doc/clientTypeMigrations.md
```diff
@@ -0,0 +1,33 @@
+# Migrating managed client types to Grpc.Core.Api
+
+Migrating types for managed client will be different from types required for managed server. The main difference is that there’s a default client base implementation whose functionality must be maintained for back compat. The types `ClientBase<T>` and `ClientBase` has constructors that take in Channel which contains abstractions for making native calls using c core which is then used by the `DefaultCallInvoker`. It is difficult to decouple `ClientBase` and `Channel` while maintaining current functionality. As an alternative, it may be necessary to modify codegen to use a different base class for the client.
```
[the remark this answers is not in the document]
jtattermusch: Also 
https://github.com/grpc/grpc/blob/master/doc/keepalive.md

doc/clientTypeMigrations.md
jtattermusch: I think I have a reasonable solution for migrating CallCredentials to Grpc.Core.Api, I think I will have a PR soon.

JunTaoLuo:
This document is woefully out of date and probably not very valuable since we've elected to go with the ChannelBase approach now that we are making breaking changes anyway. Closing unless anyone objects.

PR #184, opened 2019-03-22
Add logging of gRPC methods during binding

JamesNK:
Marc commented that it was hard to figure out what was going on during startup.

This logs the gRPC methods and their routes.


src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
```diff
@@ -152,5 +164,16 @@ private GrpcContentTypeConstraint()
             {
             }
         }
+
+        private static class Log
```
JunTaoLuo: Let's use the partial pattern for this, and move the code to GrpcService.Binder.Log.cs.

JamesNK: I want to follow the convention SignalR uses for logging. If there are a lot of log messages for an already large class then split them out into a separate file. If there are only a few then the same file is fine.

PR #185, opened 2019-03-25
Set version to 0.1.19-pre2

jtattermusch:
Once this is merged, I'll build a nuget a push it to nuget.org

- note that the packages won't work with .NET Core preview3 (because of the API update we've merged recently), so users will need to use nightlies - I'm not thrilled about this.
- we should probably put the version in a .props file so that it's shared among all the nugets (we currently publish just one, but it would be good to get rid of the duplication).

### DIFF
--- a/src/Grpc.AspNetCore.Server/Grpc.AspNetCore.Server.csproj
+++ b/src/Grpc.AspNetCore.Server/Grpc.AspNetCore.Server.csproj
@@ -8,7 +8,7 @@
     <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
     <PackageProjectUrl>https://github.com/grpc/grpc-dotnet</PackageProjectUrl>
     <PackageTags>gRPC RPC HTTP/2 aspnetcore</PackageTags>
-    <VersionPrefix>0.1.20-dev</VersionPrefix>
+    <VersionPrefix>0.1.19-pre2</VersionPrefix>
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


JamesNK:
> I'm not thrilled about this.

Yeah I agree. People using the template today will see the new version, update, and break themselves.

I don't think this should be pushed to nuget.org. We should publish this to MyGet.

JunTaoLuo:
We don't have a myget set up yet, so I think nuget is fine if we also unlist it. The build number suffix should also make it clear it's not an official release.

jtattermusch:
Looks like we're facing the problem with releasing I described in https://github.com/grpc/grpc-dotnet/pull/162#discussion_r265671254.
If we push to nuget.org and unlist it, who is the user group that's going to benefit from that?

JamesNK:
> We don't have a myget set up yet

If Jan provides us with the package we could add it to an ASP.NET feed. We don't need a continous process. This would require the template to somehow temporarily include myGet as a package source.

> nuget is fine if we also unlist it

I forgot about unlisting. Yeah, if myGet isn't possible then that would be an ok workaround.

JamesNK:
> If we push to nuget.org and unlist it, who is the user group that's going to benefit from that?

The only people who would use the package would be those that explictly chose to. The new version wouldn't show up in search results and user's wouldn't be prompted it as an upgrade.

jtattermusch:
If we agree on the nuget version, we can merge this PR, I'll build the nuget and then we can decide where to upload it?

jtattermusch:
The signed nuget is here: 

https://storage.googleapis.com/grpc-testing-kokoro-prod/test_result_public/prod/grpc/core/experimental/windows/package_signing/grpc_sign_nugets/30/20190325-162949/piper/google3/third_party/grpc/google_specific/tools/internal_ci/windows/package_signing/artifacts/Grpc.AspNetCore.Server.0.1.19-pre2.nupkg

jtattermusch:
I uploaded the package to nuget.org and unlisted it.
https://www.nuget.org/packages/Grpc.AspNetCore.Server/0.1.19-pre2

JunTaoLuo:
I'm curious why this one doesn't have a build number suffix?

jtattermusch:
> I'm curious why this one doesn't have a build number suffix?

The timestamp suffix has only been setup for grpc/grpc nugets AFAIK.

Issue #186, opened 2019-03-27
Replace bi-directional streaming sample


JamesNK:
The broadcasting in the Chatter sample is flaky and is more suited to SignalR.

Replace with a more suitable sample.

PR #187, opened 2019-03-27
Update bi-directional streaming sample

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/186

I don't like the current bi-directional streaming sample (Chatter). Maintaining a collection of IServerStreamWriter instances and calls pushing writes to each other's writers is super not threadsafe.

Updated sample has calls subscribe to a singleton queue. In the real world you could imagine this might be an Azure Queue/service bus/whatever. The queue raises change notifications that calls subscribe to, and on notification calls push content to their own stream writer.

// @shirhatti


examples/Clients/Mailer/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+            Console.WriteLine("Connecting to mailbox...");
+            Console.WriteLine();
+
+            await channel.ConnectAsync();
+
+            Console.WriteLine("Connected");
+            Console.WriteLine("Press escape to exit. Press any other key to forward mail.");
+
+            var client = new Mailer.MailerClient(channel);
+            using (var mailbox = client.Mailbox(headers: new Metadata { new Metadata.Entry("mailbox-name", "DefaultMailBox") }))
```
JunTaoLuo: So this would never be configurable anyway? Maybe make this an optional command line arg?

examples/Server/Services/Mailer.cs
```diff
@@ -0,0 +1,85 @@
...
+        public async override Task Mailbox(
+            IAsyncStreamReader<ForwardMailMessage> requestStream,
+            IServerStreamWriter<MailboxMessage> responseStream,
+            ServerCallContext context)
+        {
+            var mailboxName = context.RequestHeaders.Single(e => e.Key == "mailbox-name").Value;
+
+            var mailQueue = _messageQueueRepository.GetMailQueue(mailboxName);
+
+            Func<(int totalCount, int fowardCount, MailboxMessage.Types.Reason reason), Task> reportChanges = async (state) =>
```
davidfowl: nit: Local function looks better

examples/Server/Services/Mailer.cs
```diff
@@ -0,0 +1,85 @@
...
+                });
+            };
+            
+            _logger.LogInformation($"Connected to {mailboxName}");
+
+            mailQueue.Changed += reportChanges;
+
+            try
+            {
+                while (await requestStream.MoveNext())
```
davidfowl: We really need an IAsyncEnumerable extension method

JamesNK:
This isn't for a talk. The community looks at our samples and modifies them, and the code in chatter isn't what we want to encourage.

examples/Proto/mail.proto
```diff
@@ -14,16 +14,23 @@
...
+// The mail service definition.
+service Mailer {
+  rpc Mailbox (stream ForwardMailMessage) returns (stream MailboxMessage) {}
 }
 
-// The chat message.
-message ChatMessage {
-  string name = 1;
-  string message = 2;
+message ForwardMailMessage {
```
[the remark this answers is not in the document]
JamesNK: IMO not using empty makes it more obvious what is going on.

examples/Server/Services/Mailer.cs
JamesNK: Yup

Issue #165, opened 2019-03-16
React to ServerCallContext.UserData


JamesNK:
https://github.com/grpc/grpc/pull/18086

Should expose implement on HttpContextServerCallContext via HttpContext.Items.

JamesNK:
Blocked on Grpc.Core.Api 1.20 preview.

Issue #72, opened 2019-02-06
Populate AuthContext on ServerCallContext


JunTaoLuo:
The last item that's not tracked by issues or pending PRs from the list of properties on ServerCallContext.

JamesNK:
@jtattermusch The `AuthContext` constructor is internal

https://github.com/grpc/grpc/blob/906c12568e9bdf1d8306a2f2ab45efe8beb43f74/src/csharp/Grpc.Core.Api/AuthContext.cs#L42

Should it be made public?

JamesNK:
We'll have to leave it out in the first preview.

PR #195, opened 2019-04-03
Implement AuthContext

JamesNK:
~Fixes https://github.com/grpc/grpc-dotnet/issues/165~
Fixes https://github.com/grpc/grpc-dotnet/issues/72


src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
```diff
@@ -163,5 +171,63 @@ public static void AppendStatusTrailers(HttpResponse response, Status status)
             response.AppendTrailer(GrpcProtocolConstants.StatusTrailer, status.StatusCode.ToTrailerString());
             response.AppendTrailer(GrpcProtocolConstants.MessageTrailer, status.Detail);
         }
+
+        public static AuthContext CreateAuthContext(X509Certificate2 clientCertificate)
```
jtattermusch: @jiangtaoli2016   this is an implementation of AuthContext that  should be compatible with C core's implementation - can you take a look?

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
```diff
@@ -27,6 +32,9 @@ namespace Grpc.AspNetCore.Server.Internal
 {
     internal static class GrpcProtocolHelpers
     {
+        // https://stackoverflow.com/a/54434121/11829
+        private static readonly Regex SubjectAlternativeNameRegex = new Regex("((DNS:|DNS Name=|CN=)(?<domain>.*?),?(?=\\s|$))+", RegexOptions.CultureInvariant, TimeSpan.FromSeconds(1));
```
JunTaoLuo: @blowdard @javiercn do you have expertise on how X509Certificate2 works? Or would you be able to check whether the logic here is correct?

jiangtaoli2016:
@yihuazhang could you take a look first?

yihuazhang:
SG, I will take a look.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
JamesNK: The unit tests now passing on my computer (Windows) and the CI (Linux) confirms it works. I searched and there is no alternative API for this with X509Certificate2.

JamesNK: @blowdart

blowdart: Ummmmmm no regexes please. What are you trying to do here?

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
```diff
@@ -163,5 +171,63 @@ public static void AppendStatusTrailers(HttpResponse response, Status status)
...
+                            if (peerIdentityPropertyName == null)
+                            {
+                                peerIdentityPropertyName = GrpcProtocolConstants.X509SubjectAlternativeNameKey;
+                            }
+                        }
+                    }
+                }
+            }
+
+            var commonName = clientCertificate.GetNameInfo(X509NameType.SimpleName, false);
```
blowdart: X509 standards state if you have SAN strings (the alternative names above) then the subject/common name is invalid. You are still parsing it, if a SAN exists. This should only run if there if there is no SAN string.

Also don't use GetNameInfo for this, there's a SubjectName property.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
```diff
@@ -163,5 +171,63 @@ public static void AppendStatusTrailers(HttpResponse response, Status status)
...
+
+            foreach (var extension in clientCertificate.Extensions)
+            {
+                if (extension.Oid.Value == GrpcProtocolConstants.X509SubjectAlternativeNameId)
+                {
+                    var formattedAlternativeNamesText = extension.Format(multiLine: true);
+
+                    // The format of subject names changes between operating systems
+                    // Parse subject names using a regex that supports all major operating systems (Windows/Mac/Linux)
+                    foreach (Match match in SubjectAlternativeNameRegex.Matches(formattedAlternativeNamesText))
```
blowdart: Seriously, no regular expressions. SAN strings are separated by semicolons, and prefixed by DNS=. Split, and substring.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
JamesNK: To be compatible with other implementations of AuthContext the common name property should be in the collection. You can see on line 215 that the common name will only be used as the peer identity if there is no SAN.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
blowdart: Also SANs are for HTTPS, not for client certs. This feels a bit like a misuse for client certificate authentication, if this is what's happening here.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
JamesNK: As far as I can tell the only way to access alternative names on X509Certificate2 is to call https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.asnencodeddata.format?view=netframework-4.7.2#System_Security_Cryptography_AsnEncodedData_Format_System_Boolean_

But there are slight differences in the string result depending on the operating system:
https://stackoverflow.com/questions/54358492/cant-read-certificate-when-running-in-linux-docker-container-works-on-windows/54434121#54434121

The regex handles paring those differences.

Is there a better solution?

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
blowdart: I'm going to veer a little here. If this client certificate authentication (which is coming to asp.net core via Chris and I, in which case there's functional duplication here), then it's not exactly correct. Certificates have usages (EKUs). You're not checking if the certificate is valid for use as a client authentication certificate, and that check should be performed.

This is done via checking the EKUs directly (bad) or using Validation Policies (good, because you can also check for revocation and expiry at the same time). See https://github.com/blowdart/idunno.Authentication/blob/master/src/idunno.Authentication.Certificate/CertificateAuthenticationHandler.cs#L178 for an example.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
blowdart: Ugh. Forgot that.

Does this indicate that GRPC does mistreats certificates for auth in general though? (Multiple subjects is weird here)

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
JamesNK: It is different between OSes.

DNS Name=xxx separated by new lines on Windows
DNS:xxx separated by commas on Mac/Linux

I could parse it without a regex if you want, but I don't think it impacts security. Removing the regex would mostly be for performance.

blowdart: Keep it, but it's horrible. @bartonjs should be bugged to try to make this all cross plat identical.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
JamesNK: I don't know. We (gRPC for ASP.NET) don't use AuthContext ourselves. It is only provided for compatibility with existing gRPC apps that port from the other .NET implementation.

How people use AuthContext is up to them.

blowdart: You make me cry. Is AuthContext from the gRPC spec, or is it just from other implementations? If it's the latter then frankly I would rather make some changes to be more secure, and have a "UseCrappyCertParsing" as an option.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
bartonjs: https://github.com/dotnet/corefx/issues/22068

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
JamesNK: It is from other implementations.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
bartonjs: This code will fail on a ja-JP machine, btw.  (Windows localizes "DNS Name")

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
JamesNK: The certificate used is comes from `HttpContext.Connection.ClientCertificate`

The C core C# gRPC implementation has an enum for setting how client certificates are handled: `new SslServerCredentials(pair, File.ReadAllText(@"Certs\ca.crt"), SslClientCertificateRequestType.RequestAndRequireAndVerify)`

It seems similar to ASP.NET Core's ClientCertificateMode: `httpsOptions.ClientCertificateMode = ClientCertificateMode.RequireCertificate;`

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
bartonjs: https://github.com/dotnet/wcf/blob/08371d989fc1d230dbe9520916644f7a7f5dc954/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs#L297-L332  and https://github.com/dotnet/wcf/blob/08371d989fc1d230dbe9520916644f7a7f5dc954/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs#L376-L484 are a working example of how to do this.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
blowdart: So that, to my mind, is not enough. Especially when it blindly accepts any type of certificate. Hell, I could use an SMIME certificate meant for signing email. This goes against all best practice, and a bunch of SDL requirements. This needs a wider discussion outside of this PR about what you're trying to achieve here, compatibility or actual security.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
JamesNK: Blarg.

Thanks for the links. At least @blowdart will get to see the regex disappear 😋

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
blowdart: Where's the implementation for SslClientCertificateRequestType.RequestAndRequireAndVerify?

JamesNK: Lets discuss offline at meeting on Monday.

JamesNK: @jiangtaoli2016 What kind of validation does gRPC C core do with the client certificate?

JamesNK: > Where's the implementation for SslClientCertificateRequestType.RequestAndRequireAndVerify?

It will be deep in some C++ code somewhere, or in another library like OpenSSL/BoringSSL. I will see if I can find it.

yihuazhang: The certificate validation logic happens in `tsi_create_ssl_client_handshaker_factory_with_options()` and `tsi_create_ssl_server_handshaker_factory_with_options()` in `ssl_transport_security.cc` for client and server.

yihuazhang: IIUC `CreateAuthContext()` implements the C core logic of parsing X.509 certificate into `tsi_peer_property` and then `grpc_auth_context`. If that's the case, I think we further need to take into account peer properties such as ALPN and session reuse if they are supported.

blowdart: So part of my concerns here are the internal guidelines in our SDL for client certification validation, but also, honestly, I'm rather confused by the enumeration of SANs. Sure, it's subject alternative name, but really, it's DNS entries. It just struck me as odd. Is the intent here that the majority of folks just use their HTTPS cert?

blowdart: To clarify, is the auth piece really adjacent? I noticed in the spec it's more a case of "Whatever is presented we'll put in the context" to allow for other formats, so if I read it right, the auth, certainly with regards to certificates is not a grpc concern, but a server concern.

JamesNK:
Updated to remove regex. Certificate parsing now works in non-English cultures

PR #200, opened 2019-04-07
Refactor local benchmarks project

JamesNK:
Refactor benchmarks project to call `IWorker`'s for different benchmark runs.

Workers:
* C core gRPC client
* HttpClient gRPC client
* Minimalist HttpClient gRPC client
* JSON client (used as comparison)

And add a basic JSON endpoint and MVC action to server benchmark app.

~Includes https://github.com/grpc/grpc-dotnet/pull/199~


perf/benchmarkapps/Shared/Certs/Readme.md
```diff
@@ -0,0 +1 @@
+These certs are generated via openssl according to https://stackoverflow.com/questions/37714558/how-to-enable-server-side-ssl-for-grpc. The server.crt and server.key were combined into server.pfx. The password is 1111. These certs are not secure, do not use in production.
```
jtattermusch: IIRC we were thinking of unifying the test certs being used with the ones from grpc/grpc (currently used for interop tests), is that still the plan or we just need a different set of certs here because of infrastructure reasons (e.g. needing to share the same certs with the rest of Asp.net core benchmarking infrastructure)?

JamesNK: It's planned, but low priority.

PR #204, opened 2019-04-09
Prepare for 0.1.20-pre1 release

jtattermusch:
I'm assuming we want to unlist the nuget again after pushing? (because it depends on unreleased version of .NET core)


JamesNK:
We can't make the release public until preview 4 has shipped.

> I'm assuming we want to unlist the nuget again after pushing? (because it depends on unreleased version of .NET core)

Correct. Once preview 4 has shipped then `0.1.20-pre1` would be changed to public.

JunTaoLuo:
Ah, excuse my poor language. I meant that we **eventually** need a public 0.1.20-pre1 release.

jtattermusch:
0.1.20-pre1 published and unlisted.

Issue #175, opened 2019-03-19
Improve get-dotnet.ps1 script to install onto default path


JamesNK:
By default https://github.com/grpc/grpc-dotnet/blob/master/build/get-dotnet.ps1 is installing into the user's profile directory. This is accessible when building with `dotnet run` in PS but not VS. Likely the problem is the user's profile directory is not in the path.

We should install the nessessary version of .NET into a location that works for building via the commandline in CI *and* developing in Visual Studio.

JunTaoLuo:
We need the -InstallDir <DIRECTORY> or --install-dir <DIR> options to be set.

JamesNK:
@mgravell as a temporary work around you could install using the script above and adding the profile install location to `Path` environment variable in Windows.

mgravell:
Thanks, will try.

mgravell:
no luck with path hacks; I tried manually copying the extra SDK folder, but that resulted in:

```
C:\code\grpc-dotnet>dotnet build
It was not possible to find any compatible framework version
The specified framework 'Microsoft.NETCore.App', version '3.0.0-preview4-27518-02' was not found.
  - Check application dependencies and target a framework version installed at:
      C:\Program Files\dotnet\
```

I can build from PS for now, but... yeah, this doesn't make playing easy.

mgravell:
I got it working by manually copying both the sdk and the relevant `\shared\Microsoft.NETCore.App` folder over to the program-files location

PR #205, opened 2019-04-10
Update development infrastructure

JunTaoLuo:
Addresses https://github.com/grpc/grpc-dotnet/issues/175. Most of the infrastructure is taken from aspnet/AspNetCore.

The get-dotnet.ps1/sh will now install locally to {repoRoot}.dotnet/ and the activate.sh/ps1 will set the appropriate environment variables to use the installed SDK. Edited README to reflect the new workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,25 +28,35 @@ Documentation and guides are coming soon! In the mean time we suggest creating a
 
 ## To develop gRPC for ASP.NET Core
 
-Installing .NET Core SDK:
-```
-# Run this script before building the project.
-./build/get-dotnet.sh or ./build/get-dotnet.ps1
-```
-
-Setting up local feed with unreleased Grpc.* packages:
+Setting up local feed with unreleased Grpc.* packages (May not be necessary):
 ```
 # We may depend on unreleased Grpc.* packages.
 # Run this script before building the project.
 ./build/get-grpc.sh or ./build/get-grpc.ps1
 ```
 
-To build:
+Installing .NET Core SDK:
+```
+# Run this script before building the project.
+./build/get-dotnet.sh or ./build/get-dotnet.ps1
+```
+
+Activate the development environment:
+```
+# Source this script to use the installed .NET Core SDK.
+source activate.sh or . activate.ps1
+```
+To launch Visual Studio with the installed SDK:
+```
+startvs.cmd Grpc.AspNetCore.sln
+```
+
+To build from the command line:
 ```
 dotnet build Grpc.AspNetCore.sln
 ```
 
-To run tests:
+To run tests from the command line:
 ```
 dotnet test Grpc.AspNetCore.sln
 ```
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Documentation and guides are coming soon! In the mean time we suggest creating a
 
 ## To develop gRPC for ASP.NET Core
 
-Setting up local feed with unreleased Grpc.* packages (May not be necessary):
+Setting up local feed with unreleased Grpc.* packages:
 ```
 # We may depend on unreleased Grpc.* packages.
 # Run this script before building the project.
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -28,25 +28,35 @@ Documentation and guides are coming soon! In the mean time we suggest creating a
 
 ## To develop gRPC for ASP.NET Core
 
+Setting up local feed with unreleased Grpc.* packages (May not be necessary):
```
JamesNK:
```suggestion
Setting up local feed with unreleased Grpc.* packages:
```

Adding "may not be necessary" will just confuse people. If it isn't necessary, but there is no harm in doing it, we might as well always recommend people do it.

README.md
```diff
@@ -28,25 +28,35 @@ Documentation and guides are coming soon! In the mean time we suggest creating a
...
+```
+
 Installing .NET Core SDK:
 ```
 # Run this script before building the project.
 ./build/get-dotnet.sh or ./build/get-dotnet.ps1
 ```
 
-Setting up local feed with unreleased Grpc.* packages:
+Activate the development environment:
```
JamesNK: Can we be more descriptive? I don't know what activate means.

Is this needed if someone is using startvs.cmd?

README.md
```diff
@@ -28,25 +28,35 @@ Documentation and guides are coming soon! In the mean time we suggest creating a
...
 ```
-# We may depend on unreleased Grpc.* packages.
-# Run this script before building the project.
-./build/get-grpc.sh or ./build/get-grpc.ps1
+# Source this script to use the installed .NET Core SDK.
+source activate.sh or . activate.ps1
```
JamesNK: Our other commands start with `./`. Should these to be consistent?

README.md
```diff
@@ -28,25 +28,35 @@ Documentation and guides are coming soon! In the mean time we suggest creating a
...
+```
+To launch Visual Studio with the installed SDK:
+```
+startvs.cmd Grpc.AspNetCore.sln
```
JamesNK: We only have one solution. An argument is unnecessary.

JunTaoLuo: The script is set up to require an argument. Also, it can be used to open up *.csproj in addition to *.slns

JunTaoLuo:
Hmm need to update https://github.com/grpc/grpc/blame/master/tools/dockerfile/interoptest/grpc_interop_aspnetcore/build_interop.sh#L27-L32 to fix the interop tests.

JunTaoLuo:
Interop test is not required. I think there just needs to be an approval.

README.md
JamesNK: But in aspnetcore there are startvs.cmd scripts in solution directories that don't require an argument.

There is only one valid argument for startvs.cmd in this repo. Why are we making people do extra work?

JunTaoLuo: Hmm I was basing the behaviour off of https://github.com/aspnet/AspNetCore/blob/master/startvs.cmd but I suppose I can combine the functionality of that script and https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/startvs.cmd with a default parameter.

jtattermusch:
Looks like this has broken the interop test build on grpc/grpc repository: https://github.com/grpc/grpc/issues/18431
We need to be more careful when updating the way things are built, because grpc/grpc also runs a continuous interop test against master of grpc-dotnet.  Basically whenever the script that runs interop tests in this repo is changing, same changes need to be done in grpc/grpc to avoid breakage.

jtattermusch:
I might have spoken too soon - I didn't notice https://github.com/grpc/grpc/pull/18714 earlier. But we should have synced more around merging this PR (and subsequently breaking grpc/grpc) anyway.

JamesNK:
Sorry for creating issues with this PR. We underestimated the impact of changing the build in other places.

I think for now we'll always get your approval before merging, except if the PR has low risk/impact changes such as:
- A minor refactor or fix
- Unit test changes
- Sample or documentation

Does that sound good to you?

jtattermusch:
> Sorry for creating issues with this PR. We underestimated the impact of changing the build in other places.
> 
> I think for now we'll always get your approval before merging, except if the PR has low risk/impact changes such as:
> 
> * A minor refactor or fix
> * Unit test changes
> * Sample or documentation
> 
> Does that sound good to you?

SGTM.

Issue #206, opened 2019-04-10
Interop test server failure seen on grpc/grpc interop suite


jtattermusch:
https://source.cloud.google.com/results/invocations/71ff4afa-806a-4947-a2d5-567c86e34735/log

This could be a bug in Grpc.AspNetCore.Server, as the server shouldn't just exit.

```
info: Microsoft.AspNetCore.Routing.EndpointMiddleware[0]
      Executing endpoint 'gRPC - /grpc.testing.TestService/FullDuplexCall'
info: Microsoft.AspNetCore.Routing.EndpointMiddleware[1]
      Executed endpoint 'gRPC - /grpc.testing.TestService/FullDuplexCall'
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
      Request finished in 0.8264ms 200 application/grpc
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/2 POST https://foo.test.google.fr/grpc.testing.TestService/FullDuplexCall application/grpc
info: Microsoft.AspNetCore.Routing.EndpointMiddleware[0]
      Executing endpoint 'gRPC - /grpc.testing.TestService/FullDuplexCall'
fail: Grpc.Testing.TestServiceImpl[3]
      Error when executing service method 'FullDuplexCall'.
System.IO.IOException: The client reset the request stream.
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestPipeReader.ReadAsync(CancellationToken cancellationToken)
   at Grpc.AspNetCore.Server.Internal.PipeExtensions.ReadStreamMessageAsync(PipeReader input, HttpContextServerCallContext context, CancellationToken cancellationToken) in /var/local/git/grpc-dotnet/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs:line 271
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at Grpc.AspNetCore.Server.Internal.HttpContextStreamReader`1.<MoveNext>g__MoveNextAsync|10_0(ValueTask`1 readStreamTask) in /var/local/git/grpc-dotnet/src/Grpc.AspNetCore.Server/Internal/HttpContextStreamReader.cs:line 48
[...]

PR #207, opened 2019-04-10
Add exception handling to deadline cancellation

JamesNK:
We don't want to throw an unhandled exception from the deadline timer, it will kill the process.

https://github.com/grpc/grpc-dotnet/issues/206
https://github.com/aspnet/AspNetCore/issues/9239

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.Log.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.Log.cs
@@ -42,7 +42,7 @@ namespace Grpc.AspNetCore.Server.Internal
                 LoggerMessage.Define<string>(LogLevel.Debug, new EventId(5, "EncodingNotInAcceptEncoding"), "Request grpc-encoding header value '{GrpcEncoding}' is not in grpc-accept-encoding.");
 
             private static readonly Action<ILogger, Exception> _deadlineCancellationError =
-                LoggerMessage.Define(LogLevel.Error, new EventId(6, "DeadlineCancellationError"), "Error canceling request because of the deadline.");
+                LoggerMessage.Define(LogLevel.Error, new EventId(6, "DeadlineCancellationError"), "Error occurred while trying to cancel the request due to deadline exceeded.");
 
             public static void DeadlineExceeded(ILogger logger, TimeSpan timeout)
             {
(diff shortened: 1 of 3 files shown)


src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.Log.cs
```diff
@@ -41,6 +41,9 @@ private static class Log
             private static readonly Action<ILogger, string, Exception> _encodingNotInAcceptEncoding =
                 LoggerMessage.Define<string>(LogLevel.Debug, new EventId(5, "EncodingNotInAcceptEncoding"), "Request grpc-encoding header value '{GrpcEncoding}' is not in grpc-accept-encoding.");
 
+            private static readonly Action<ILogger, Exception> _deadlineCancellationError =
+                LoggerMessage.Define(LogLevel.Error, new EventId(6, "DeadlineCancellationError"), "Error canceling request because of the deadline.");
```
jtattermusch: nit: the error message is not very readable.

jtattermusch: I think the meaning is something like "Error occurred while trying to cancel the request due to deadline exceeded"

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -263,8 +263,15 @@ private void DeadlineExceeded(object state)
...
-            // TODO(JamesNK): I believe this sends a RST_STREAM with INTERNAL_ERROR. Grpc.Core sends NO_ERROR
-            HttpContext.Abort();
+            try
+            {
+                // TODO(JamesNK): I believe this sends a RST_STREAM with INTERNAL_ERROR. Grpc.Core sends NO_ERROR
+                HttpContext.Abort();
+            }
+            catch (Exception ex)
+            {
+                Log.DeadlineCancellationError(_logger, ex);
```
davidfowl: This doesn't seem like the right fix

JamesNK: It doesn't fix why a NRE was thrown. But it at least stops the process exiting.

I want to figure out why Abort could throw in https://github.com/aspnet/AspNetCore/issues/9239

PR #208, opened 2019-04-10
Make Greeter sample calls more obvious

JamesNK:


examples/Clients/Greeter/Program.cs
```diff
@@ -35,9 +35,24 @@ static async Task Main(string[] args)
             var channel = new Channel("localhost:50051", credentials);
             var client = new Greeter.GreeterClient(channel);
 
-            var reply = client.SayHelloAsync(new HelloRequest { Name = "GreeterClient" });
-            Console.WriteLine("Greeting: " + (await reply.ResponseAsync).Message);
```
JamesNK: I didn't realize you could await AsyncUnaryCall directly.

examples/Clients/Greeter/Program.cs
```diff
@@ -35,9 +35,24 @@ static async Task Main(string[] args)
...
 
+            await ServerStreamingCall(client);
+
+            Console.WriteLine("Shutting down");
+            await channel.ShutdownAsync();
+            Console.WriteLine("Press any key to exit...");
+            Console.ReadKey();
+        }
+
+        private static async Task UnaryCall(Greeter.GreeterClient client)
```
jtattermusch: super nit: rename to UnaryCallExample so it's clear that "UnaryCall" is just a helper method and not part of gRPC API?

examples/Clients/Greeter/Program.cs
jtattermusch: Yes, there's a GetAwaiter()  method  for that purpose.

PR #212, opened 2019-04-16
Managed gRPC client

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/43

To do:
- ~Validating content type~
- ~Deadline (client timeout)~
- ~Deadline (copying C-core HTTP request header formatting)~
- ~CallOptions.CancellationToken~

Future:
- Compression
- CallOptions.WriteOptions
- CallOptions.PropagationToken
- CallOptions.Credentials
- CallOptions.IsWaitForReady


examples/Clients/Counter/Program.cs
```diff
@@ -34,7 +36,8 @@ static async Task Main(string[] args)
             // Server will only support Https on Windows and Linux
             var credentials = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? ChannelCredentials.Insecure : ClientResources.SslCredentials;
             var channel = new Channel("localhost:50051", credentials);
-            var client = new Counter.CounterClient(channel);
+            //var client = new Counter.CounterClient(channel);
+            var client = GrpcClientFactory.Create<Counter.CounterClient>("https://localhost:50051", new System.Security.Cryptography.X509Certificates.X509Certificate2(Path.Combine(Resources.CertDir, "client.crt")));
```
JamesNK: Ignore changes in the samples. They are me playing around. I'll clean them up when this is closer to done.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,314 @@
...
+                    message.Headers.Add(entry.Key, value);
+                }
+            }
+
+            if (Options.Deadline != null && Options.Deadline != DateTime.MaxValue)
+            {
+                var deadline = Options.Deadline.Value - _clock.UtcNow;
+
+                // JamesNK(todo) - Replicate C core's logic for formatting grpc-timeout
+                message.Headers.Add(GrpcDeadlineHeader, Convert.ToInt64(deadline.TotalMilliseconds) + "m");
```
JamesNK: @jtattermusch Could you point me towards where C core sets the grpc-timeout header? I want to replicate it here.

JamesNK:
Cleaned up. Ready for review

// @davidfowl

src/Grpc.NetCore.HttpClient/Grpc.NetCore.HttpClient.csproj
```diff
@@ -12,15 +12,14 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
```
JamesNK: Trailing headers is netcoreapp3.0 only today. I have an issue on dotnet/standard that proposes adding the property to netstandard2.1

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
JamesNK: https://github.com/grpc/grpc/blob/master/src/core/lib/transport/timeout_encoding.cc

src/Grpc.NetCore.HttpClient/Internal/HttpContextClientStreamReader.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            cancellationToken.ThrowIfCancellationRequested();
+
+            if (_responseStream == null)
+            {
+                var responseMessage = await _call.SendTask.ConfigureAwait(false);
+                _responseStream = await responseMessage.Content.ReadAsStreamAsync().ConfigureAwait(false);
+            }
+
+            using (cts)
```
JunTaoLuo: Interesting, it's safe to pass null to a using statement.

src/Grpc.NetCore.HttpClient/Grpc.NetCore.HttpClient.csproj
```diff
@@ -12,15 +12,17 @@
...
     <LangVersion>8.0</LangVersion>
     <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
   <ItemGroup>
+    <!-- Roslyn analyzer that enforces ConfigureAwait(false) -->
+    <PackageReference Include="ConfigureAwaitChecker.Analyzer" Version="3.0.0" PrivateAssets="All" />
```
JunTaoLuo: Use dependency.props to manage the version for this dependency.

test/Grpc.NetCore.HttpClient.Tests/AsyncDuplexStreamingCallTests.cs
```diff
@@ -0,0 +1,166 @@
...
+            Assert.IsNotNull(responseStream.Current);
+            Assert.AreEqual("Hello world 2", responseStream.Current.Message);
+
+            var moveNextTask3 = responseStream.MoveNext(CancellationToken.None);
+            Assert.IsFalse(moveNextTask3.IsCompleted);
+
+            await streamContent.AddDataAndWait(Array.Empty<byte>()).DefaultTimeout();
+
+            Assert.IsFalse(await moveNextTask3.DefaultTimeout());
+            Assert.IsFalse(await moveNextTask3.DefaultTimeout());
```
JunTaoLuo: Why the second assert

test/Grpc.NetCore.HttpClient.Tests/AsyncServerStreamingCallTests.cs
```diff
@@ -0,0 +1,157 @@
...
+            Assert.IsNotNull(responseStream.Current);
+            Assert.AreEqual("Hello world 2", responseStream.Current.Message);
+
+            var moveNextTask3 = responseStream.MoveNext(CancellationToken.None);
+            Assert.IsFalse(moveNextTask3.IsCompleted);
+
+            await streamContent.AddDataAndWait(Array.Empty<byte>()).DefaultTimeout();
+
+            Assert.IsFalse(await moveNextTask3.DefaultTimeout());
+            Assert.IsFalse(await moveNextTask3.DefaultTimeout());
```
JunTaoLuo: Same here.

test/Grpc.NetCore.HttpClient.Tests/Grpc.NetCore.HttpClient.Tests.csproj
```diff
@@ -0,0 +1,30 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Protobuf Include="Proto\*.proto" GrpcServices="Client" Generator="MSBuild:Compile" />
+    <None Remove="@(Protobuf)" />
```
JunTaoLuo: No need for the None and Content item groups. This is only for design time build and we don't really need it for the tests.

src/Grpc.NetCore.HttpClient/Internal/HttpContextClientStreamReader.cs
```diff
@@ -0,0 +1,89 @@
...
+            // HTTP response has finished
+            if (_call.ResponseFinished)
+            {
+                return false;
+            }
+
+            // User could have disposed call
+            _call.EnsureNotDisposed();
+
+            // Linking tokens is expensive. Only create a linked token is passed in cancellation token requires it
```
JunTaoLuo: Any test for this? Also, nit: the language here seems off. `Only create a linked token if the token passed in requires it`

src/Grpc.NetCore.HttpClient/Grpc.NetCore.HttpClient.csproj
JunTaoLuo: Also, are we allowed to use this tool? For our repos, we have strict requirements on which third party tools we were allowed to use.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,246 @@
...
+            // The task of this method is cached so there is no need to cache the headers here
+            return GrpcProtocolHelpers.BuildMetadata(_httpResponse.Headers);
+        }
+
+        public Status GetStatus()
+        {
+            ValidateTrailersAvailable();
+
+            var grpcStatus = SendTask.Result.TrailingHeaders.GetValues(GrpcProtocolConstants.StatusTrailer).FirstOrDefault();
+            var grpcMessage = SendTask.Result.TrailingHeaders.GetValues(GrpcProtocolConstants.MessageTrailer).FirstOrDefault();
```
JunTaoLuo: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses:
>The value portion of Status-Message is conceptually a Unicode string description of the error, physically encoded as UTF-8 followed by percent-encoding.

We should unescape the perccent-encoding here.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,246 @@
...
+            Options = options;
+            _clock = clock;
+        }
+
+        public CancellationToken CancellationToken
+        {
+            get { return _callCts.Token; }
+        }
+
+        public void SendUnary(System.Net.Http.HttpClient client, TRequest request)
```
jtattermusch: nit: why the "Send" prefix? Does "StartUnary" make more sense (especially for client streaming calls that don't immediately send a request)?

src/Grpc.NetCore.HttpClient/Internal/HttpContentClientStreamWriter.cs
```diff
@@ -0,0 +1,112 @@
...
+        public void Dispose()
+        {
+        }
+
+        private async Task WriteAsyncCore(TRequest message)
+        {
+            // Wait until the client stream has started
+            var writeStream = await _writeStreamTask.ConfigureAwait(false);
+
+            await SerialiationHelpers.WriteMessage<TRequest>(writeStream, message, _call.Method.RequestMarshaller.Serializer, _call.CancellationToken).ConfigureAwait(false);
```
jtattermusch: typo: SerialiationHelpers

src/Grpc.NetCore.HttpClient/Internal/SerialiationHelpers.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.Buffers.Binary;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Grpc.NetCore.HttpClient.Internal
+{
+    internal static class SerialiationHelpers
```
jtattermusch: typo: SerialiationHelpers

test/Grpc.NetCore.HttpClient.Tests/AsyncDuplexStreamingCallTests.cs
jtattermusch: The contract is that once MoveNext() returns false or throws, it will continue doing so forever. Add a comment?

src/Grpc.NetCore.HttpClient/Grpc.NetCore.HttpClient.csproj
jtattermusch: nit: add a TODO?

src/Grpc.NetCore.HttpClient/Grpc.NetCore.HttpClient.csproj
JamesNK: `PrivateAssets="All"` = build time dependency. ASP.NET Core will never see it.

@jtattermusch Do you have any issues with using this tool? It's MIT, and a Roslyn analyzer that can be removed at anytime. I want to use it, or something like it, to ensure we have `ConfigureAwait(false)` everywhere. See https://stackoverflow.com/questions/27851073/why-would-i-bother-to-use-task-configureawaitcontinueoncapturedcontext-false/27851460#27851460

jtattermusch: I'm fine with using that analyzer.

test/Shared/SyncPoint.cs
```diff
@@ -20,7 +20,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Grpc.AspNetCore.Server.Tests
+namespace Grpc.Tests
```
jtattermusch: nit: use a slightly less general namespace than Grpc.Tests?

test/Grpc.NetCore.HttpClient.Tests/Grpc.NetCore.HttpClient.Tests.csproj
JamesNK: Meh. When we can remove them we'll remove them everywhere.

test/Shared/SyncPoint.cs
JamesNK: Changed to Grpc.Tests.Shared.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
JamesNK: It is called Send because these methods call HttpClient.SendAsync.

Client streaming doesn't talk to the server until the first message is sent?

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
JamesNK: Added todo.

JamesNK:
Big update. @jtattermusch Can you review the interop test client? I'd like to merge this and get the HttpClient gRPC client running in your interop environment ASAP. If there are failures caused by other servers, and the problem is HttpClient then we need to get issues raised to the HttpClient team quickly.

Interop test status:

*   "empty_unary"
*  "large_unary"
*  "client_streaming"
*  "server_streaming"
*  "ping_pong" - Fails
*  "empty_stream"
*  "compute_engine_creds" - Fails
*  "jwt_token_creds" - Fails
*  "oauth2_auth_token" - Fails
*  "per_rpc_creds" - Fails
*  "cancel_after_begin"
*  "cancel_after_first_response" - Fails
*  "timeout_on_sleeping_server"
*  "custom_metadata"
*  "status_code_and_message"
*  "unimplemented_service"
*  "unimplemented_method"
*  "client_compressed_unary" - Fails
*  "client_compressed_streaming" - Fails

global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "3.0.100-preview4-011108"
+    "version": "3.0.100-preview6-011568"
```
JamesNK: Updated the SDK so HttpClient supports making HTTP/2 requests without TLS.

src/Grpc.NetCore.HttpClient/Internal/HttpContentClientStreamWriter.cs
```diff
@@ -0,0 +1,127 @@
...
+            lock (_writeLock)
+            {
+                // Pending writes need to be awaited first
+                if (IsWriteInProgress)
+                {
+                    return Task.FromException(new InvalidOperationException("Cannot complete client stream writer because the previous write is in progress."));
+                }
+
+                // Notify that the client stream is complete
+                _completeTcs.TrySetResult(true);
```
davidfowl: This is dispatching right?

src/Grpc.NetCore.HttpClient/Internal/HttpContentClientStreamWriter.cs
```diff
@@ -0,0 +1,127 @@
...
+using Grpc.Core;
+
+namespace Grpc.NetCore.HttpClient.Internal
+{
+    internal class HttpContentClientStreamWriter<TRequest, TResponse> : IClientStreamWriter<TRequest>
+    {
+        private readonly GrpcCall<TRequest, TResponse> _call;
+        private readonly Task<Stream> _writeStreamTask;
+        private readonly TaskCompletionSource<bool> _completeTcs;
+        private readonly object _writeLock;
```
davidfowl: SemaphoreSlim?

src/Grpc.NetCore.HttpClient/Internal/HttpContentClientStreamWriter.cs
JamesNK: Pardon?

It is used here: https://github.com/grpc/grpc-dotnet/blob/9812a63ea2aca64b4059594465c35690ddf379db/src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs#L302-L323

This writer writes to a stream, and the stream comes from the HttpContent on the request. The HttpContent is considered complete when its SerializeToStream method finishes. This TCS holds it open.

src/Grpc.NetCore.HttpClient/Internal/HttpContentClientStreamWriter.cs
JamesNK: The lock is used when a write starts, it doesn't lock the entire write. It isn't used for async locking.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,421 @@
...
+            }
+
+            ResponseFinished = true;
+
+            // Clean up call resources once this call is finished
+            // Call may not be explicitly disposed when used with unary methods
+            // e.g. var reply = await client.SayHelloAsync(new HelloRequest());
+            Dispose();
+
+            HttpResponse = httpResponseMessage;
```
BrennanConroy: This is odd, you've called dispose at this point but you're modifying the state of `this` after disposing. Is there a reason you need to set the field or can you just use the local for `GetStatusCore`?

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,421 @@
...
+            }
+
+            return message;
+        }
+
+        private void DeadlineExceeded(object state)
+        {
+            if (!_callCts.IsCancellationRequested)
+            {
+                // Flag is used to determin status code when generating exceptions
```
BrennanConroy:
```suggestion
                // Flag is used to determine status code when generating exceptions
```

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,421 @@
...
+                DeadlineReached = true;
+
+                _callCts.Cancel();
```
halter73: Is there any way user code could block this call to _callCts.Cancel()?

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,421 @@
...
+
+        public void Dispose()
+        {
+            if (!Disposed)
+            {
+                Disposed = true;
+
+                _callCts.Cancel();
+                _callCts.Dispose();
+                _ctsRegistration?.Dispose();
```
halter73: I think you need to dispose the _ctsRegistration before disposing _callCts to avoid a possible ODE in CancelCall.

Directory.Build.props
```diff
@@ -9,6 +9,9 @@
 
     <!-- Don't make missing XML docs a fatal build error, but still surface so we have visibility into undocumented APIs. -->
     <WarningsNotAsErrors>$(WarningsNotAsErrors);CS1591</WarningsNotAsErrors>
+
+    <!-- static local methods are a C# 8.0 feature -->
+    <LangVersion>preview</LangVersion>
```
halter73: Should we use 8.0 instead of preview to be more explicit?

src/Grpc.NetCore.HttpClient/GrpcClientFactory.cs
```diff
@@ -49,7 +49,10 @@ public static class GrpcClientFactory
                 handler.ClientCertificates.Add(certificate);
             }
 
-            return Cache<TClient>.Instance.Activator(new HttpClientCallInvoker(handler, new Uri(baseAddress, UriKind.RelativeOrAbsolute)));
+            var httpClient = new System.Net.Http.HttpClient(handler);
```
halter73: Nit:

```suggestion
            var httpClient = new HttpClient(handler);
```

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,421 @@
...
+                _writerCtsRegistration?.Dispose();
+                _deadlineTimer?.Dispose();
```
halter73: I think this should also be done before disposing _callCts.

src/Grpc.NetCore.HttpClient/Internal/HttpContextClientStreamReader.cs
```diff
@@ -0,0 +1,103 @@
...
+                {
+                    await _call.SendTask.ConfigureAwait(false);
+                    _httpResponse = _call.HttpResponse;
+                }
+                if (_responseStream == null)
+                {
+                    _responseStream = await _httpResponse.Content.ReadAsStreamAsync().ConfigureAwait(false);
+                }
+
+                using (cts)
```
halter73: Move this using up before all the stuff that can throw between here and CreateLinkedTokenSource.

src/Grpc.NetCore.HttpClient/Internal/HttpContextClientStreamReader.cs
```diff
@@ -0,0 +1,103 @@
...
+using System;
+using System.IO;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Grpc.Core;
+
+namespace Grpc.NetCore.HttpClient.Internal
+{
+    internal class HttpContextClientStreamReader<TRequest, TResponse> : IAsyncStreamReader<TResponse>
```
halter73: I just noticed this is HttpConte**x**tClientStreamReader while the writer is call HttpConte**n**tClientStreamWriter was that intentional.

src/Grpc.NetCore.HttpClient/Internal/HttpContentClientStreamWriter.cs
```diff
@@ -0,0 +1,127 @@
...
+                }
+                else if (_completeTcs.Task.IsCanceled)
+                {
+                    throw _call.CreateCanceledStatusException();
+                }
+
+                // Pending writes need to be awaited first
+                if (IsWriteInProgress)
+                {
+                    return Task.FromException(new InvalidOperationException("Cannot write message because the previous write is in progress."));
```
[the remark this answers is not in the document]
davidfowl: We can still do it. File a bug.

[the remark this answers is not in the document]
JamesNK: With IAsyncEnumerator, what happens if you do something similar. For example:

```cs
var t1 = e.MoveNextAsync();
var t2 = e.MoveNextAsync();

await t1;
await t2;
```

I'm guessing whether it is an error or not depends on the implementation. Is there guidance?

src/Grpc.NetCore.HttpClient/GrpcClientFactory.cs
JamesNK: The explicit namespace is there because HttpClient is the namespace. It will be renamed before release.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
JamesNK: Ah. Setting this isn't required anymore because it happens when awaiting the send task.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
JamesNK: I don't believe so. The cancellation token is only exposed internally.

src/Grpc.NetCore.HttpClient/Internal/HttpContentClientStreamWriter.cs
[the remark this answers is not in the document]
JamesNK: Nevermind. I tested what C Core does and it throws an error.

src/Grpc.NetCore.HttpClient/Internal/HttpContextClientStreamReader.cs
JamesNK: I've added a finally to the try/catch to ensure it is cleaned up.

Directory.Build.props
JamesNK: 8.0 creates warnings about static local methods. Will re-evaluate closer to 3.0

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,433 @@
...
+            }
+
+            ResponseFinished = true;
+
+            // Clean up call resources once this call is finished
+            // Call may not be explicitly disposed when used with unary methods
+            // e.g. var reply = await client.SayHelloAsync(new HelloRequest());
+            Dispose();
+
+            var status = GetStatusCore(HttpResponse);
```
BrennanConroy: Still using a disposed object here, what's wrong with moving this before `Dispose`

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,433 @@
...
+
+        public void StartDuplexStreaming(System.Net.Http.HttpClient client)
+        {
+            var message = CreateHttpRequestMessage();
+            ClientStreamWriter = CreateWriter(message);
+            StartSend(client, message);
+            ClientStreamReader = new HttpContentClientStreamReader<TRequest, TResponse>(this);
+        }
+
+        public void Dispose()
```
rynowak: Is there a reason why the type isn't disposable?

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,433 @@
...
+            if (options.Deadline != null && options.Deadline != DateTime.MaxValue)
+            {
+                var timeout = options.Deadline.Value - _clock.UtcNow;
+                _timeout = (timeout > TimeSpan.Zero) ? timeout : TimeSpan.Zero;
+            }
+
+            if (_timeout != null)
+            {
+                // Deadline timer will cancel the call CTS
+                _deadlineTimer = new Timer(DeadlineExceeded, null, _timeout.Value, Timeout.InfiniteTimeSpan);
```
rynowak: What's the reason to use this instead of `.CancelAfter`?

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -0,0 +1,433 @@
...
+        {
+            var message = CreateHttpRequestMessage();
+            ClientStreamWriter = CreateWriter(message);
+            StartSend(client, message);
+            ClientStreamReader = new HttpContentClientStreamReader<TRequest, TResponse>(this);
+        }
+
+        public void Dispose()
+        {
+            if (!Disposed)
```
rynowak: It's not super clear to me what the threading model is for this (since you've got cancellation and timeouts). How does this work without locking?

src/Grpc.NetCore.HttpClient/Internal/GrpcProtocolHelpers.cs
```diff
@@ -0,0 +1,111 @@
...
+
+#endregion
+
+using System;
+using System.Net.Http.Headers;
+using Grpc.Core;
+
+namespace Grpc.NetCore.HttpClient.Internal
+{
+    internal static class GrpcProtocolHelpers
```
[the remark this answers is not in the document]
rynowak: The world is a better place when you just drop the word "helpers" from these apis.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
JamesNK: A different status is returned depending on whether the call was cancelled via a token or via the deadline being exceeded. A CTS doesn't provide a reason why it was cancelled (Cancel vs CancelAfter)

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
JamesNK: It was never be used as IDisposable (cast as IDisposable, or used with a using statement). On the other hand it doesn't hurt so I can change it.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
halter73: I'm concerned about this too.

As long as the GrpcCall object itself is accessed from a single logical thread and then disposed after it's used, `_ctsRegistration?.Dispose()` should block until any possibly-already-triggered calls to `CancelCall` complete. This effectively locks on our behalf.

I don't know if `_deadlineTimer?.Dispose()` will similarly wait for `DeadlineExceeded` to complete. I don't think it does. It would be tempting to use a lock in Dispose and DeadlineExceeded to protect against ODE's, but you risk deadlocking if you're not extremely careful.

After a lot of effort trying to properly dispose the CTS backing the RequestAborted token without ODEs or deadlocking, @jkotalik and I found the best way to avoid deadlocking is to simply not even try disposing CTS that's already been canceled. https://github.com/aspnet/AspNetCore/pull/9333

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
JamesNK: The only property being used from the HttpResponse is trailers which dispose doesn't effect. However, there is no reason status couldn't be fetched before the dispose and then an error is thrown afterwards.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
halter73: It might be better to construct a new CTS with the specified timeout and register DeadlineExceeded with that instead of using a timer. That way when you dispose the registration, you wait for DeadlineExceeded to complete (if it's currently executing) automatically.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
halter73: I don't think you need to do anything as complicated as what we did for the RequestAborted token since you don't try to lazily allocate the CTS nor do you expose the CancellationToken to the end user.

I do think that using a CTS instead of a Timer for DeadlineExceeded might help you avoid an ODE in DeadlineExceeded though.

JamesNK: Dispose, cancellation and error handling is immature at the moment.

Dispose can happen 3 ways:
1. Response headers are invalid (bad HttpStatusCode or bad content type)
2. Response is received and we finish reading the body
3. User explicitly calls dispose on the public API, which gets routed here

Then there are additional ways _callCts can be canceled:
1. The deadline could be exceeded
2. The user's passed in cancellation token is cancelled

src/Grpc.NetCore.HttpClient/Internal/ISystemClock.cs
```diff
@@ -0,0 +1,27 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System;
+
+namespace Grpc.NetCore.HttpClient.Internal
+{
+    internal interface ISystemClock
```
JunTaoLuo: We keep having this exact interface all around our stack...

src/Grpc.NetCore.HttpClient/Internal/GrpcProtocolHelpers.cs
```diff
@@ -0,0 +1,188 @@
...
+                    return EncodeTimeoutSeconds(x / MillisecondsPerSecond);
+                }
+                else
+                {
+                    return FormatTimeout(x, 'm');
+                }
+            }
+        }
+
+        public static string EncodeTimeout(long timeout)
```
JunTaoLuo: Where is this used?

src/Grpc.NetCore.HttpClient/Internal/HttpContentClientStreamWriter.cs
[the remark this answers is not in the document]
halter73: https://github.com/aspnet/AspNetCore/issues/9876 is the HttpResponse issue.

testassets/InteropTestsClient/IChannel.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
```
JunTaoLuo: nit: copyright

src/Grpc.NetCore.HttpClient/Internal/ISystemClock.cs
JamesNK: aspnet/specs is waiting for you Microsoft.Extensions.Clock PR 😋

examples/Clients/Counter/Program.cs
```diff
@@ -36,9 +37,24 @@ static async Task Main(string[] args)
             var channel = new Channel("localhost:50051", credentials);
             var client = new Counter.CounterClient(channel);
 
-            var reply = client.IncrementCount(new Google.Protobuf.WellKnownTypes.Empty());
+            await UnaryCallExample(client);
```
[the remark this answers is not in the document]
JamesNK: When this is ready all our sample clients should use it.

JamesNK:
Merging so this can be setup in the interop tests.

Still work to do in follow up PRs to get thread-safety perfect but that can be handled in follow up work.

Thanks for everyone's feedback.

PR #216, opened 2019-04-19
Special status message

JunTaoLuo:
Percent encode the status message. Addresses https://github.com/grpc/grpc-dotnet/issues/161.

I have tested locally with a custom client but I'll see if I can run the interop tests as well.

PR #218, opened 2019-04-24
Revert "Special status message"

jtattermusch:
Reverts grpc/grpc-dotnet#216

See https://github.com/grpc/grpc-dotnet/pull/216#issuecomment-486184313

Fixes https://github.com/grpc/grpc/issues/18843


jtattermusch:
CC @JunTaoLuo

jtattermusch:
Analysis of the problem:
From the protocol spec:
`Percent-Byte-Unencoded → 1*( %x20-%x24 / %x26-%x7E ) ; space and VCHAR, except %`
Looks like some characters (e.g. space) should not be encoded. 
The problem is only revealed by the dart client which (unlike other languages) doesn't percent-decode the characters received as part of status-message, so `test%20status%20message` is surfaced instead of `test status message`.

JunTaoLuo:
Hmm interesting. A custom encoding might be needed then.

Issue #44, opened 2019-01-24
Support interceptors in AspNetCore.Server


JamesNK:
https://github.com/grpc/proposal/blob/master/L12-csharp-interceptors.md

https://grpc.io/grpc/csharp/api/Grpc.Core.Interceptors.html

JunTaoLuo:
This might impact #1

JamesNK:
@jtattermusch Do you think this is important to implement in gRPC ASP.NET Server? In ASP.NET we will have middleware that will fulfill the same purpose.

jtattermusch:
I think it's fine to not have this in the first version and we can revisit later. It would be useful to do a quick check though to see if this is possible from the API perspective (i.e. what types we would need to move to Grpc.Core.Api and see if they are easily movable).

JamesNK:
ok! thanks

Kadajski:
Middleware fulfills a _similar_ purpose, though how would you get access to the actual request or response objects in your middleware? Seems like the deserialization/serialization all happens inside of e.g. `UnaryServerCallHandler` so on either end you'd only have access to bytes

Kadajski:
> I think it's fine to not have this in the first version and we can revisit later. It would be useful to do a quick check though to see if this is possible from the API perspective (i.e. what types we would need to move to Grpc.Core.Api and see if they are easily movable).

I am pretty sure the only things required that were mentioned to have not been moved over are:

| Types |
|---------------------------------------------------------------------------|
| Grpc.Core.Interceptors.Interceptor |
| Grpc.Core.Interceptors.Interceptor+AsyncClientStreamingCallContinuation`2 |
| Grpc.Core.Interceptors.Interceptor+AsyncDuplexStreamingCallContinuation`2 |
| Grpc.Core.Interceptors.Interceptor+AsyncServerStreamingCallContinuation`2 |
| Grpc.Core.Interceptors.Interceptor+AsyncUnaryCallContinuation`2 |
| Grpc.Core.Interceptors.Interceptor+BlockingUnaryCallContinuation`2 |

The interceptor is pretty stand-alone. The only thing referencing anything else are the extension methods which would need to be rewritten to match this new repo as the old `ServerServiceDefinition` doesn't seem to be used anymore.

I did some investigations into how this _could_ work and came up with some sample code for a unary call. https://github.com/Kadajski/grpc-dotnet/commit/68a398bb7f4ce7ba2c6052a579276ca6bacbc0b9

[...]

PR #220, opened 2019-04-26
Interceptors for Grpc.AspNetCore.Server

JunTaoLuo:
Addresses #44 

A few notes:
- Interceptors can be scoped or singletons (if added as a singleton service)
- Supports constructor injection from service container
- Separate interceptor pipeline and aspnetcore middleware pipeline. Interleaving of middlewares and interceptors not supported
- Middlewares operate before serialization, interceptors operate after serialization
- Interceptors configured via GrpcServiceOptions


examples/Server/Startup.cs
```diff
@@ -27,14 +29,28 @@ public class Startup
         // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddGrpc();
+            services
+                .AddGrpc(options =>
```
shirhatti: Why is this a fluent API? It looks really weird

JamesNK: It allows configuration to be customized per service. It's no different than the fluent API hanging off AddSignalR. Most people will set the global config in AddGrpc and not customize per service.

Do you have an alternative API?

src/Grpc.AspNetCore.Server/GrpcServiceOptions.cs
```diff
@@ -73,6 +76,34 @@ public List<ICompressionProvider> CompressionProviders
...
+        /// <summary>
+        /// List of interceptors to be executed. Interceptors executed in order.
+        /// </summary>
+        internal List<Type> Interceptors { get; private set; } = new List<Type>();
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        public void AddInterceptor<T>()
```
JamesNK: I think of this as like UseMiddleware, and like UseMiddleware I think we should take a collection of args that get passed to the constructor.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.usemiddlewareextensions.usemiddleware

src/Grpc.AspNetCore.Server/GrpcServiceOptions.cs
```diff
@@ -73,6 +76,34 @@ public List<ICompressionProvider> CompressionProviders
...
+        {
+            AddInterceptor(typeof(T));
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="interceptors"></param>
+        public void AddInterceptor(params Type[] interceptors)
```
JamesNK: IMO remove the params and just accept one interceptor. Basically be a non-generic version of `AddInterceptor<T>()`

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
```diff
@@ -57,8 +70,22 @@ protected override async Task HandleCallAsyncCore(HttpContext httpContext)
 
                 service = activator.Create();
 
-                response = await _invoker(
+                ClientStreamingServerMethod<TRequest, TResponse> resolvedInvoker = delegate (IAsyncStreamReader<TRequest> resolvedRequestStream, ServerCallContext resolvedContext)
```
JamesNK: image

JamesNK: Even if there are no interceptors it looks like we're still wrapping the underlying method call. Can we make interceptors zero cost when they aren't present? (which should be most of the time)

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
```diff
@@ -31,14 +33,25 @@ internal class ClientStreamingServerCallHandler<TService, TRequest, TResponse> :
         where TService : class
     {
         private readonly ClientStreamingServerMethod<TService, TRequest, TResponse> _invoker;
+        private readonly IEnumerable<Type> _interceptors;
```
JamesNK: I'd prefer to store a list or array over an enumerable.

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
```diff
@@ -57,8 +70,22 @@ protected override async Task HandleCallAsyncCore(HttpContext httpContext)
...
+                {
+                    return _invoker(
                     service,
+                    resolvedRequestStream,
+                    resolvedContext);
+                };
+
+                foreach (var interceptor in _interceptors)
+                {
+                    var interceptorInstance = (Interceptor)ActivatorUtilities.CreateInstance(httpContext.RequestServices, interceptor);
```
JamesNK: We have IGrpcServiceActivator that allows users to change how services are created, i.e. override the default per-request creation to singletons. I feel like we have an equivalent for interceptors. Either introduce a new interface, or perhaps IGrpcServiceAcivator should be responsible for activating the service's interceptors as well.

src/Grpc.AspNetCore.Server/Internal/ClientInterceptorContext.cs
```diff
@@ -0,0 +1,63 @@
...
+using System;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Grpc.Core.Interceptors
+{
+    /// <summary>
+    /// Carries along the context associated with intercepted invocations on the client side.
+    /// </summary>
+    public struct ClientInterceptorContext<TRequest, TResponse>
```
JamesNK: Double check: This is temporary and will come from Grpc.Core.Api?

src/Grpc.AspNetCore.Server/Internal/Interceptor.cs
```diff
@@ -0,0 +1,402 @@
...
+#endregion
+
+using System.Threading.Tasks;
+
+namespace Grpc.Core.Interceptors
+{
+    /// <summary>
+    /// Serves as the base class for gRPC interceptors.
+    /// </summary>
+    public abstract class Interceptor
```
JamesNK: Double check: This is temporary and will come from Grpc.Core.Api?

JamesNK:
I think the service configuration is fine, but a more idiomatic ASP.NET way of defining filters would be placing attributes on the service implementation class. Should investigate what that would look like.

examples/Server/Startup.cs
JamesNK: Defining an interceptor via an attributes would be the more user friendly way of configuring interceptors per service. See https://github.com/grpc/grpc-dotnet/pull/220#issuecomment-487243597

shirhatti: My bad, I didn't notice the `IGrpcServerBuilder`. On first glance I assumed everything hangs off `IServiceCollection` which felt weird.

As an aside, I'm not really a fan that some `IServiceCollection` extension methods return builders whereas others return `IServiceCollection`. It feels inconsistent to me, but I just realized we do the same thing in MVC, SignalR, et al.

build/dependencies.props
```diff
@@ -2,9 +2,9 @@
   <PropertyGroup>
     <BenchmarkDotNetPackageVersion>0.11.3</BenchmarkDotNetPackageVersion>
     <GoogleProtobufPackageVersion>3.7.0</GoogleProtobufPackageVersion>
-    <GrpcCorePackageVersion>1.20.0-pre3</GrpcCorePackageVersion>
-    <GrpcCoreApiPackageVersion>1.20.0-pre3</GrpcCoreApiPackageVersion>
-    <GrpcToolsPackageVersion>1.20.0-pre3</GrpcToolsPackageVersion>
+    <GrpcCorePackageVersion>1.21.0-dev201904210946</GrpcCorePackageVersion>
```
jtattermusch: what exactly do we need from the newer packages?

examples/Server/Startup.cs
```diff
@@ -27,14 +29,28 @@ public class Startup
...
+                 {
+                     options.AddInterceptor<GreetingCachingInterceptor>();
+                 });
             services.AddSingleton<IncrementingCounter>();
             services.AddSingleton<MailQueueRepository>();
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app)
+        public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
```
davidfowl: Why did you add this?

build/dependencies.props
JunTaoLuo: `CallOptions`.

examples/Server/Startup.cs
JunTaoLuo: I was using it to test interceptors at one point

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
JunTaoLuo: I'm thinking we'll go with a custom collection.

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
JunTaoLuo: Yup, will need to optimize the fast path.

src/Grpc.AspNetCore.Server/Internal/ClientInterceptorContext.cs
JunTaoLuo: Yes https://github.com/grpc/grpc/pull/18914

jtattermusch:
FYI https://github.com/grpc/grpc/pull/18914 has been merged.

src/Grpc.AspNetCore.Server/GrpcServiceOptions.cs
```diff
@@ -73,6 +76,27 @@ public List<ICompressionProvider> CompressionProviders
...
+        /// </summary>
+        internal List<Type> Interceptors { get; private set; } = new List<Type>();
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        public void AddInterceptor<T>(params object[] args)
+        {
+            // TODO: handle args
```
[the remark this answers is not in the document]
JamesNK: They are passed to the interceptor's constructor.

examples/Clients/Mailer/Program.cs
```diff
@@ -35,7 +35,7 @@ static async Task Main(string[] args)
             // Server will only support Https on Windows and Linux
             var credentials = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? ChannelCredentials.Insecure : ClientResources.SslCredentials;
 
-            var channel = new Channel("localhost:50051", credentials);
+            var channel = new Channel("localhost:5002", credentials);
```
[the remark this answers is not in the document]
JunTaoLuo: This PR isn't ready for final reviews yet and all the code change in examples are for local testing, I opened it to illustrate what I had in mind for the server side implementation. This port change was due to 50051 port being blocked on my machine for some reason so I changed the port I was testing on. I'll update the examples to something more reasonable when I mark this PR as ready for review.

examples/Server/Interceptors/GreetingCachingInterceptor.cs
```diff
@@ -0,0 +1,45 @@
...
+        private static IDictionary<string, string> _cache = new Dictionary<string, string>();
+
+        public GreetingCachingInterceptor(ILogger<GreetingCachingInterceptor> logger)
+        {
+            _logger = logger;
+        }
+
+        public override async Task<TResponse> UnaryServerHandler<TRequest, TResponse>(TRequest request, ServerCallContext context, UnaryServerMethod<TRequest, TResponse> continuation)
+        {
+            var greetingRequest = request as HelloRequest;
```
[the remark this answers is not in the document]
JamesNK: I'm not a fan of this interceptor. I think it should be left out of samples.

examples/Server/Interceptors/LoggingInterceptor.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Server.Interceptors
+{
+    public class LoggingInterceptor : Interceptor
+    {
+        ILogger _logger;
+        string _prefix;
+
+        public LoggingInterceptor(ILogger<LoggingInterceptor> logger, string prefix)
```
JamesNK: Is the prefix to test passing values via AddInterceptor? How about changing prefix to environment.

Log message can then be something like: `Beginning request to {context.Method} on {environment}`

examples/Server/Server.csproj
```diff
@@ -25,4 +25,8 @@
     <PackageReference Include="Grpc.Tools" Version="$(GrpcToolsPackageVersion)" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Properties\" />
```
JamesNK: remove?

src/Grpc.AspNetCore.Server/GrpcServiceOptions.cs
```diff
@@ -73,6 +76,11 @@ public List<ICompressionProvider> CompressionProviders
             }
             set => _compressionProviders = value;
         }
+
+        /// <summary>
+        /// Collections of interceptors to be executed. Interceptors are executed in order.
```
JamesNK:
```suggestion
        /// Gets a collection of interceptors that are executed with every call. Interceptors are executed in order.
```

src/Grpc.AspNetCore.Server/IGrpcInterceptorActivator.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// A <typeparamref name="TInterceptor"/> activator abstraction.
+    /// </summary>
+    /// <typeparam name="TInterceptor">The interceptor type.</typeparam>
+    public interface IGrpcInterceptorActivator<TInterceptor> where TInterceptor : Interceptor
+    {
+        /// <summary>
+        /// Creates a interceptor.
+        /// </summary>
+        /// <returns>The created service.</returns>
+        TInterceptor Create(params object[] args);
```
JamesNK: Document args

src/Grpc.AspNetCore.Server/IGrpcInterceptorActivator.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Grpc.AspNetCore.Server
+{
+    /// <summary>
+    /// A <typeparamref name="TInterceptor"/> activator abstraction.
+    /// </summary>
+    /// <typeparam name="TInterceptor">The interceptor type.</typeparam>
+    public interface IGrpcInterceptorActivator<TInterceptor> where TInterceptor : Interceptor
+    {
+        /// <summary>
+        /// Creates a interceptor.
```
JamesNK:
```suggestion
        /// Creates an interceptor.
```

src/Grpc.AspNetCore.Server/IGrpcInterceptorActivator.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// </summary>
+        /// <returns>The created service.</returns>
```
JamesNK:
```suggestion
        /// <returns>The created interceptor.</returns>
```

src/Grpc.AspNetCore.Server/IGrpcInterceptorActivator.cs
```diff
@@ -0,0 +1,41 @@
...
+        TInterceptor Create(params object[] args);
+
+        /// <summary>
+        /// Releases the specified interceptor.
+        /// </summary>
+        /// <param name="service">The interceptor to release.</param>
+        void Release(TInterceptor service);
```
JamesNK:
```suggestion
        void Release(TInterceptor interceptor);
```

src/Grpc.AspNetCore.Server/InterceptorCollection.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Grpc.AspNetCore.Server
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class InterceptorCollection : IEnumerable<InterceptorRegistration>
```
JamesNK: @rynowak Did you envision the interceptor registration collection this as a fully fledged `ICollection<T>` collection or something like this?

src/Grpc.AspNetCore.Server/InterceptorCollection.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        private static readonly IEnumerator<InterceptorRegistration> EmptyEnumerator = Enumerable.Empty<InterceptorRegistration>().GetEnumerator();
+
+        private List<InterceptorRegistration> _store;
```
JamesNK: A maximum of one of these is created per service. Does lazily creating the underlying list provide much benefit?

src/Grpc.AspNetCore.Server/InterceptorCollection.cs
```diff
@@ -0,0 +1,97 @@
...
+        public IReadOnlyList<InterceptorRegistration> Snapshot => _store?.AsReadOnly(); 
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="args"></param>
+        public void Add<T>(params object[] args)
+        {
+            // TODO: handle args
```
JamesNK: What do you mean here?

src/Grpc.AspNetCore.Server/InterceptorRegistration.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public Type Type { get; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        internal object[] Args { get; }
```
JamesNK: Why internal? You could internally save as `object[]` and publicly expose as `IReadOnlyList<object>` to keep it immutable.

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
```diff
@@ -56,11 +69,46 @@ protected override async Task HandleCallAsyncCore(HttpContext httpContext)
...
+                        service,
+                        resolvedRequestStream,
+                        resolvedContext);
+                    };
+
+                    for (var i = _interceptors.Count - 1; i >= 0; i--)
+                    {
+                        // TODO: Need to check if there's a way to avoid reflection
+                        var registration = _interceptors[i];
+                        var activatorType = typeof(IGrpcInterceptorActivator<>).MakeGenericType(registration.Type);
```
JamesNK: Eek. Ok, how about internally caching this type on the registration: `typeof(IGrpcInterceptorActivator<>).MakeGenericType(registration.Type)`. It doesn't change.

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
```diff
@@ -56,11 +69,46 @@ protected override async Task HandleCallAsyncCore(HttpContext httpContext)
...
+                        var interceptorActivator = httpContext.RequestServices.GetRequiredService(activatorType);
+                        var createMethod = activatorType.GetMethod("Create");
```
JamesNK: Add a non-generic `IGrpcInterceptorActivator` interface that `IGrpcInterceptorActivator<T>` inherits from. It would have non-generic Create and Release methods. You can cast to the non-generic version and use the methods without reflection.

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
```diff
@@ -31,14 +33,25 @@ internal class ClientStreamingServerCallHandler<TService, TRequest, TResponse> :
...
+            };
+        };
 
         public ClientStreamingServerCallHandler(
             Method<TRequest, TResponse> method,
             ClientStreamingServerMethod<TService, TRequest, TResponse> invoker,
             GrpcServiceOptions serviceOptions,
-            ILoggerFactory loggerFactory) : base(method, serviceOptions, loggerFactory)
+            ILoggerFactory loggerFactory,
+            IReadOnlyList<InterceptorRegistration> interceptors)
```
JamesNK: Could the interceptors be fetched from the already passed in GrpcServiceOptions?

src/Grpc.AspNetCore.Server/InterceptorCollection.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// 
+        /// </summary>
+        /// <returns></returns>
+        public bool IsEmpty => _store == null;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <returns></returns>
+        public IReadOnlyList<InterceptorRegistration> Snapshot => _store?.AsReadOnly();
```
JamesNK: What is this for? At the very least this should be a method instead of a property.

src/Grpc.AspNetCore.Server/Internal/DefaultGrpcInterceptorActivator.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public TInterceptor Create(params object[] args)
+        {
+            Debug.Assert(!_created.HasValue, "Grpc interceptor activator must not be reused.");
+
+            _created = false;
+            var interceptor = _serviceProvider.GetService<TInterceptor>();
+            if (interceptor == null)
+            {
+                interceptor = ActivatorUtilities.CreateInstance<TInterceptor>(_serviceProvider, args); // TODO: Is there a way to factory-ize this?
```
JamesNK: Can you use ActivatorUtilities.CreateFactory? Convert the arg values to type with GetType.

src/Grpc.AspNetCore.Server/InterceptorCollection.cs
JunTaoLuo: I think the impact would be small enough at startup that I wouldn't worry about it for now.

src/Grpc.AspNetCore.Server/InterceptorRegistration.cs
JunTaoLuo: I was thinking about this and couldn't come up with a use where the arguments would need to be inspected so I took the approach of not exposing unless needed.

src/Grpc.AspNetCore.Server/InterceptorCollection.cs
JunTaoLuo: I wanted to start off by exposing the least amount of API/functionality and add them as need arises. For example, I'm not sure there's a need for Remove. Also, currently, there's no need for the user to instantiate a new instance of InterceptorRegistration explicitly.

I'm not opposed to these changes though.

JunTaoLuo: Actually now I think about it, though strange, it might make more sense to make it an IReadonlyList<InterceptorRegistration> but with an explicit Add method.

src/Grpc.AspNetCore.Server/Internal/DefaultGrpcInterceptorActivator.cs
JunTaoLuo: There are some interesting implications here. For example, it's possible in the c-core pipeline to add two instances of the same interceptor that were constructed with different args to the pipeline. If we were to support that, I don't think we can use the same trick as DefaultGrpcServiceActivator to create a static factory, in which case compiling a lambda every call may be more inefficient.

examples/Server/Interceptors/GreetingCachingInterceptor.cs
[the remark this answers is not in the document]
JunTaoLuo: I'm mostly using these interceptors to test and see what the API would look like for registration and implementation. I don't plan on checking them in when I mark the PR is ready for review but are there any interceptors that would be interesting to implement as an actual example?

examples/Server/Interceptors/LoggingInterceptor.cs
JunTaoLuo: More for testing the Add<TInterceptor>(params object[] args) method. If it were just environment, it would be resolved from DI not the args parameter. I agree that the example is very contrived but again, I don't think this example will remain.

examples/Server/Server.csproj
JunTaoLuo: Ugh, I hate how VS automatically adds launchsettings when you debug with F5. The file is ignored by git but VS still creates a folder in the project.

src/Grpc.AspNetCore.Server/Internal/DefaultGrpcInterceptorActivator.cs
JamesNK: It is not perfectly efficient but I doubt it will be a problem. How many interceptors is even a very large app going to register? 20? 50?

I think it is an acceptable way of doing it for now, and we can change the delegate caching later if it does become problematic.

src/Grpc.AspNetCore.Server/Internal/DefaultGrpcInterceptorActivator.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            _serviceProvider = serviceProvider;
+        }
+
+        public Interceptor Create(params object[] args)
+        {
+            Debug.Assert(!_created.HasValue, "Grpc interceptor activator must not be reused.");
+
+            _created = false;
+            var interceptor = _serviceProvider.GetService<TInterceptor>();
```
JamesNK: This will skip any args you have provided. I think that if you provide arguments then CreateFactory should always be used.

That should be documented in `AddInterceptor<T>`

src/Grpc.AspNetCore.Server/Internal/DefaultGrpcInterceptorActivator.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (interceptor == null)
+            {
+                if (_objectFactory == null)
+                {
+                    var argumentTypes = new Type[args.Length];
```
JamesNK: If length is zero then skip allocating an array.

src/Grpc.AspNetCore.Server/Internal/DefaultGrpcInterceptorActivator.cs
```diff
@@ -0,0 +1,79 @@
...
+
+                    for (var i = 0; i < args.Length; i++)
+                    {
+                        argumentTypes[i] = args[i].GetType();
```
JamesNK: Need to validate that args are never null. That should happen in `AddInterceptor<T>`. Put a comment here saying that args have already been validated

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerStreamingServerCallHandler.cs
```diff
@@ -56,11 +72,44 @@ protected override async Task HandleCallAsyncCore(HttpContext httpContext)
...
+                {
+                    await _invoker(
+                        service,
+                        request,
+                        new HttpContextStreamWriter<TResponse>(serverCallContext, Method.ResponseMarshaller.Serializer),
+                        serverCallContext);
+                }
+                else
+                {
+                    ServerStreamingServerMethod<TRequest, TResponse> resolvedInvoker = delegate (TRequest request, IServerStreamWriter<TResponse> responseStream, ServerCallContext resolvedContext)
```
JamesNK: nit: change these to use lambda syntax

src/Grpc.AspNetCore.Server/InterceptorCollection.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="args"></param>
+        public void Add<T>(params object[] args)
+        {
+            var interceptorType = typeof(T);
+            if (!interceptorType.IsSubclassOf(typeof(Interceptor)))
```
JamesNK: Why to place a generic constraint that T must be an `Interceptor`?

src/Grpc.AspNetCore.Server/Internal/DefaultGrpcInterceptorActivator.cs
JunTaoLuo: Hmm... I'm not sure if we should just reject null as arguments. For example, we sometimes use nullables to indicate a setting or a lack thereof.

JamesNK: How could null map to a value in the activator factory?

What does UseMiddleware do when there is a null argument?

JunTaoLuo: Looking at our middleware implementation: https://github.com/aspnet/AspNetCore/blob/master/src/Http/Http.Abstractions/src/Extensions/UseMiddlewareExtensions.cs#L92 we don't disallow nulls.

examples/Server/Interceptors/MaxConcurrentCallsInterceptor.cs
```diff
@@ -0,0 +1,110 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Server.Interceptors
+{
+    public class MaxConcurrentCallsInterceptor : Interceptor
+    {
+        private static readonly RpcException _maxConcurrentCallsExceededException = new RpcException(new Status(StatusCode.ResourceExhausted, "Maximum number of concurrent calls exceeded."), "Maximum number of concurrent calls exceeded.");
+        private static SemaphoreSlim _concurrentCalls;
```
JamesNK: This is static but assigned in the constructor?

examples/Server/Interceptors/MaxConcurrentCallsInterceptor.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(maxConcurrentCalls), maxConcurrentCalls, $"{nameof(maxConcurrentCalls)} must be a positive number.");
+            }
+
+            _concurrentCalls = new SemaphoreSlim(maxConcurrentCalls, maxConcurrentCalls);
+        }
+
+        public override async Task<TResponse> UnaryServerHandler<TRequest, TResponse>(TRequest request, ServerCallContext context, UnaryServerMethod<TRequest, TResponse> continuation)
+        {
+            if (!_concurrentCalls.Wait(0))
```
JamesNK: nit: move if check into a method and reuse

examples/Server/Interceptors/MaxStreamingRequestTimeoutInterceptor.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            public void Dispose()
+            {
+            }
+
+            public async Task<bool> MoveNext(CancellationToken cancellationToken)
+            {
+                var task = _inner.MoveNext(cancellationToken);
+
+                if (task.IsCompleted || Debugger.IsAttached)
```
JamesNK: Should Debugger.IsAttached still be here?

examples/Server/Interceptors/UnaryCachingInterceptor.cs
```diff
@@ -0,0 +1,60 @@
...
+        private readonly ILogger _logger;
+
+        public UnaryCachingInterceptor(ILogger<UnaryCachingInterceptor> logger)
+        {
+            _logger = logger;
+        }
+
+        public override async Task<TResponse> UnaryServerHandler<TRequest, TResponse>(TRequest request, ServerCallContext context, UnaryServerMethod<TRequest, TResponse> continuation)
+        {
+            if (_cache.TryGetValue<TResponse>(request, out var cachedResponse))
```
JamesNK: This would require the the request type to implement Equals/GetHashCode, yes? e.g. HelloRequest is equal to another HelloRequest when the name is equal. Should probably have a comment explaining that

JamesNK: Or do the generated types implement Equals/GetHashCode automatically?

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
```diff
@@ -31,14 +33,24 @@ internal class ClientStreamingServerCallHandler<TService, TRequest, TResponse> :
         where TService : class
     {
         private readonly ClientStreamingServerMethod<TService, TRequest, TResponse> _invoker;
+        private readonly IReadOnlyList<InterceptorRegistration> _interceptors;
```
JamesNK: Move to protected property on the base class. Also property should be `List<InterceptorRegistration>` to avoid interface dispatch overhead of IReadOnlyList

src/Grpc.AspNetCore.Server/Internal/CallHandlers/DuplexStreamingServerCallHandler.cs
```diff
@@ -52,11 +68,45 @@ protected override async Task HandleCallAsyncCore(HttpContext httpContext)
...
+                        service,
+                        requestStream,
+                        responseStream,
+                        resolvedContext);
+                    };
+
+                    // The list is reversed during construction so the first interceptor is invoked first
+                    for (var i = _interceptors.Count - 1; i >= 0; i--)
+                    {
+                        var registration = _interceptors[i];
```
JamesNK: Can 93-100 be moved to a protected method on the base class and reused in each call handler?

src/Grpc.AspNetCore.Server/Internal/DefaultGrpcInterceptorActivator.cs
```diff
@@ -0,0 +1,72 @@
...
+            if (args.Length == 0)
+            {
+                var globalInterceptor = _serviceProvider.GetService<TInterceptor>();
+                if (globalInterceptor != null)
+                {
+                    return globalInterceptor;
+                }
+            }
+
+            var interceptor = ActivatorUtilities.CreateInstance<TInterceptor>(_serviceProvider, args);
```
JamesNK: Hmm, this uses reflection every time. We can figure out perf here, and add some micro-benchmarks, in a follow up PR

examples/Server/Interceptors/UnaryCachingInterceptor.cs
JunTaoLuo: Yes the generated types implement this.

examples/Server/Interceptors/MaxStreamingRequestTimeoutInterceptor.cs
JunTaoLuo: I don't see why not, it's annoying to have to deal with timeouts when debugging.

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
JunTaoLuo: Or just use the ServiceOptions on the base class.

JunTaoLuo:
Need to look into why kokoro builds are failing (maybe because of the dependency update to a nightly?). Also looking into why the travis ci isn't triggering.

build_and_test.sh
```diff
@@ -18,7 +18,7 @@ set -ex
 source activate.sh
 
 # Required when using nightly builds of gRPC packages
-# ./build/get-grpc.sh
+./build/get-grpc.sh
```
JunTaoLuo: This would be temporary until 1.21.0-pre1 is released.

src/Grpc.AspNetCore.Server/InterceptorCollection.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        /// <inheritdoc />
+        public InterceptorRegistration this[int index] => _store?[index];
+
+        /// <summary>
+        /// Add an interceptor to the end of the pipeline.
+        /// </summary>
+        /// <typeparam name="TInterceptor">The interceptor type.</typeparam>
+        /// <param name="args">The list of arguments to pass to the interceptor constructor when creating an instance.</param>
+        public void Add<TInterceptor>(params object[] args) where TInterceptor : Interceptor
```
JunTaoLuo: Actually we shouldn't have this API. This was taken from Middlewares that does not implement IMiddleware: https://github.com/aspnet/AspNetCore/blob/master/src/Http/Http.Abstractions/src/Extensions/UseMiddlewareExtensions.cs#L59-L110. This is used to construct instances of middleware types which uses DI to inject services in the Invoke/InvokeAsync() methods. We are opting for constructor injection like IMiddleware.

src/Grpc.AspNetCore.Server/InterceptorCollection.cs
```diff
@@ -56,6 +56,20 @@ public class InterceptorCollection : IReadOnlyList<InterceptorRegistration>
             _store.Add(new InterceptorRegistration(typeof(TInterceptor), args));
         }
 
+        /// <summary>
+        /// Add a registration for an interceptor to the end of the pipeline.
+        /// </summary>
+        /// <param name="registration">The registration for the interceptor to add.</param>
+        public void Add(InterceptorRegistration registration)
```
JamesNK: Why make this public? The ctor for `InterceptorRegistration` is internal

JunTaoLuo:
Added a microbenchmark and here are the results:
Before:
```
---------------- |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 HandleCallAsync | 1.713 us | 0.0341 us | 0.0365 us | 583,885.6 |      0.0439 |           - |           - |               280 B |
```
After:
```
---------------- |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 HandleCallAsync | 1.766 us | 0.0335 us | 0.0329 us | 566,280.2 |      0.0496 |           - |           - |               320 B |

--------------------------- |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 InterceptedHandleCallAsync | 2.185 us | 0.0362 us | 0.0338 us | 457,706.6 |      0.1106 |           - |           - |               696 B |
```

PR #221, opened 2019-04-29
Filter known headers from request headers

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/81

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/GrpcProtocolConstants.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/GrpcProtocolConstants.cs
@@ -16,6 +16,10 @@
 
 #endregion
 
+using System;
+using System.Collections.Generic;
+using Microsoft.Net.Http.Headers;
+
 namespace Grpc.AspNetCore.Server.Internal
 {
     internal static class GrpcProtocolConstants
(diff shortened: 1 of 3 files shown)


src/Grpc.AspNetCore.Server/Internal/GrpcProtocolConstants.cs
```diff
@@ -31,5 +35,16 @@ internal static class GrpcProtocolConstants
         internal const string MessageAcceptEncodingHeader = "grpc-accept-encoding";
 
         internal const string IdentityGrpcEncoding = "identity";
+
+        internal static readonly HashSet<string> FilteredHeaders = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
JamesNK: @jtattermusch I compared Grpc.Core vs Grpc.AspNetCore.Server and found that Grpc.Core filtered a lot of "known" headers. I've included headers that were present in Grpc.AspNetCore.Server to filter in the list below. Do you know of others, or where in C Core logic this filter happens?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -27,6 +27,7 @@
 using Grpc.Core;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
+using Microsoft.Net.Http.Headers;
```
davidfowl: Where is this used?

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolConstants.cs
jtattermusch: There's not authoritative list, but I think in practice most of "well known" metadata entries won't be exposes to the user in C-core https://github.com/grpc/grpc/blob/master/src/core/lib/transport/static_metadata.h

JamesNK: I can see user-agent in that file, and I know it is exposed.

Does the filter I have here need to be changed or can it be merged?

jtattermusch: I think it's good be be merged, the headers listed seem to be the most important ones, we can add others later.

Issue #179, opened 2019-03-20
Publish nightly packages to https://packages.grpc.io/


JunTaoLuo:
@jtattermusch Can we add this repo to the infrastructure that builds nightly packages and pushes to https://packages.grpc.io/? This would help users to try out our latest features as we work on them. We've had more and more people asking to try out the nightlies instead of the latest preview.

jtattermusch:
I can take a look, but it might be non-trivial, as the the infrastructure to publish to  packages.grpc.io lives in the grpc/grpc repository and it is designed for packages generated from the same repository.

JunTaoLuo:
Can we duplicate the infrastructure in this repo?

JunTaoLuo:
Actually we need this to fix issues like https://github.com/aspnet/AspNetCore/issues/8693.

JunTaoLuo:
@shirhatti can you coordinate with @eilon and see if there's anything we can do here to help?

JamesNK:
@jtattermusch Have you made any progress with using Artifactory to host a NuGet feed?

jtattermusch:
The artifactory repository has been setup, this is the nuget source url:

https://grpc.jfrog.io/grpc/api/nuget/grpc-nuget-dev

nuget v3
https://grpc.jfrog.io/grpc/api/nuget/v3/grpc-nuget-dev 

For now the packages still need to be uploaded to that feed manually (I uploaded e.g. `Grpc 1.21.0-dev201904151532` manually), automatic uploads of nightly builds haven't been setup yet.

PR #222, opened 2019-04-29
Publish nightly build of grpc-dotnet nugets to dev feed

jtattermusch:
Fixes https://github.com/grpc/grpc-dotnet/issues/179 

- doesn't push dev packages to packages.grpc.io (which is meant for packages coming from grpc/grpc repository), but pushes them to the artifactory nuget dev feed.

- introduce version.props that has the version of all nugets from grpc/grpc-dotnet repository. If the package has a -dev suffix, the package version is expanded to contain current timestamp.

Dev nuget feed: https://grpc.jfrog.io/grpc/api/nuget/v3/grpc-nuget-dev

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,7 @@
 <Project>
   <Import Project="build\sources.props" />
   <Import Project="build\dependencies.props" />
+  <Import Project="build\version.props" />
 
   <PropertyGroup>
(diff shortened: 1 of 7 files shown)


JamesNK:
When will new packages be pushed to the NuGet feed? Is it a timer, or a new version per commit?

jtattermusch:
@JamesNK  It's basically a daily cron job, so there will be one build per day.

```
periodic_build_schedule: "@daily"
```

If that's desirable, running daily AND after each PR merged to master branch is also something we could configure.

Issue #161, opened 2019-03-13
support "special_status_message" interop test


jtattermusch:
https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md

Not a high priority (not all grpc languages currently implement this test).

JamesNK:
What are we failing to do to get the test to pass?

jtattermusch:
I'm not sure if we are handling the "Status-Message" encoding correctly in case special characters are used.

> The value portion of Status-Message is conceptually a Unicode string description of the error, physically encoded as UTF-8 followed by percent-encoding. Percent-encoding is specified in RFC 3986 §2.1, although the form used here has different restricted characters. When decoding invalid values, implementations MUST NOT error or throw away the message. At worst, the implementation can abort decoding the status message altogether such that the user would received the raw percent-encoded form. Alternatively, the implementation can decode valid portions while leaving broken %-encodings as-is or replacing them with a replacement character (e.g., '?' or the Unicode replacement character).

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

jtattermusch:
I verified that currently Grpc.AspNetCore.Server does not pass the test:
https://source.cloud.google.com/results/invocations/08537a88-6131-448b-bf7a-33aa1022ada9/targets/github%2Fgrpc%2Finterop_test/tests;group=tests;test=cloud_to_cloud:go:aspnetcore_server:special_status_message:tls;row=19

Here are a few clients that support the special_status_message interop test and fail against grpc.aspnetcore.server (it's an adhoc run with the test case enabled):

Go
```
FATAL: 2019/03/20 14:43:01 &{0xc000134300}.UnaryCall(_, response_status:<code:2 message:"\t\ntest with whitespace\r\nand Unicode BMP \342\230\272 and non-BMP \360\237\230\210\t\n" > ) = _, rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: INTERNAL_ERROR, want _, rpc error: code = Unknown desc =
test with whitespace
and Unicode BMP  and non-BMP
exit status 1
```

Java
```
Running test special_status_message
Exception in thread "main" java.lang.AssertionError: expected:<UNKNOWN> but was:<INTERNAL>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
[...]

PR #223, opened 2019-04-30
 Special status message II

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/161


src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool IsUnreservedCharacter(char c)
+        {
+            return ((PercentEncodingUnreservedBytes[c / 8] >> (c % 8)) & 1) != 0;
+        }
+
+        // TODO(JamesNK): Remove WIP decode once we are certain Uri.UnescapeData works correctly
+#if Decode
+        public static string PercentDecode(string value)
```
JamesNK: I did most of decode, then realized that I needed to worry about multi-byte characters, surrogate characters, UTF8 validation, etc. Ain't nobody got time for that.

Uri.UnescapeData should probably work fine for decoding. Leaving this here until we are sure it is not needed.

jtattermusch: Yes, I think using `Uri.UnescapeData` data should work as the only difference in the percent-encoding is the choice of characters that get percent-encoded.

test/Grpc.AspNetCore.Server.Tests/PercentEncodingHelpersTests.cs
```diff
@@ -0,0 +1,84 @@
...
+        public void DecodeInvalidMessageTrailer(string expectedDecodedValue, string value)
+        {
+            // Arrange & Act
+            var decodedValue = Uri.UnescapeDataString(value);
+
+            // Assert
+            Assert.AreEqual(expectedDecodedValue, decodedValue);
+        }
+
+        private static TestCaseData[] ValidPercentEncodingTestCases =
```
jtattermusch: Some of the java test cases might come handy
https://github.com/grpc/grpc-java/blob/80c3c992a66aa21ccf3e12e38000316e45f97e64/api/src/test/java/io/grpc/StatusTest.java#L65

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
```diff
@@ -0,0 +1,205 @@
...
+        private const int AsciiMaxValue = 127;
+
+        private static readonly byte[] PercentEncodingUnreservedBytes =
+        {
+            0x00, 0x00, 0x00, 0x00, 0xdf, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+            0xff, 0xff, 0xff, 0xff, 0x7f, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
+        };
+
+        public static string PercentEncode(string value)
```
jtattermusch: FYI, this is what Java does:
https://github.com/grpc/grpc-java/blob/b363f80764bcc8452ef54511b6c6d6b596f5f177/api/src/main/java/io/grpc/Status.java#L574

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
```diff
@@ -0,0 +1,205 @@
...
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal static class PercentEncodingHelpers
+    {
+        private static readonly char[] HexChars = "0123456789ABCDEF".ToCharArray();
+        private const int MaxUnicodeCharsReallocate = 40; // Maximum batch size when working with unicode characters
+        private const int MaxUtf8BytesPerUnicodeChar = 4;
+        private const int AsciiMaxValue = 127;
+
+        private static readonly byte[] PercentEncodingUnreservedBytes =
```
jtattermusch: make it obvious that this is a bitfield?

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
```diff
@@ -0,0 +1,205 @@
...
+            {
+            }
+
+            return unicodeCharCount;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool IsUnreservedCharacter(char c)
+        {
+            return ((PercentEncodingUnreservedBytes[c / 8] >> (c % 8)) & 1) != 0;
```
jtattermusch: how about this? https://github.com/grpc/grpc-java/blob/b363f80764bcc8452ef54511b6c6d6b596f5f177/api/src/main/java/io/grpc/Status.java#L592, seems simpler

JamesNK: I copied the bitfield from the C Core implementation. I can add a comment linking to the original implementation, or replace it with Java's. Up to you.

test/Grpc.AspNetCore.Server.Tests/PercentEncodingHelpersTests.cs
JamesNK: Ok, I'll add them.

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
JamesNK: Looks like a lot of allocations 😋 Not that matters since successful messages tend to not return a message with OK.

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
jtattermusch: I think the java way is more readable and the C core approach might be a premature optimization in our case IMHO. Leaving the final decision up to you though.

JamesNK: FYI they aren't equivalent. Java encodes ~ and C Core doesn't.

JunTaoLuo:
cc @grabyourpitchforks since we love adding custom encoding logic.

JamesNK:
I don't think this should be too controversial because it still uses Encoding.UTF8.GetBytes to convert non-ASCII chars to bytes. Those bytes are then encoded with percent encoding.

Decoding continues to use `Uri.UnescapeDataString`.

But would like an expert to double check before merging 😬

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
```diff
@@ -207,7 +207,21 @@ public static void SetStatusTrailers(HttpResponse response, Status status)
...
+
+            string escapedDetail;
+            if (!string.IsNullOrEmpty(status.Detail))
+            {
+                // https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses
+                // The value portion of Status-Message is conceptually a Unicode string description of the error,
+                // physically encoded as UTF-8 followed by percent-encoding.
+                escapedDetail = PercentEncodingHelpers.PercentEncode(status.Detail);
+            }
+			else
```
JunTaoLuo: why not just declare it as null and set it to a value in the if block

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
```diff
@@ -0,0 +1,218 @@
...
+        private static void EscapeAsciiChar(ref Span<char> span, ref int writePosition, char current)
+        {
+            span[writePosition++] = '%';
+            span[writePosition++] = HexChars[current >> 4];
+            span[writePosition++] = HexChars[current & 15];
+        }
+
+        private static int GetUnicodeCharCount(string value, int currentIndex)
+        {
+            var unicodeCharCount = 0;
```
JunTaoLuo: can start at 1?

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
```diff
@@ -0,0 +1,218 @@
...
+                }
+            }
+
+            return unicodeCharCount;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool IsUnreservedCharacter(char c)
+        {
+            return ((PercentEncodingUnreservedBitField[c / 8] >> (c % 8)) & 1) != 0;
```
JunTaoLuo: is doing all this arithmetic really better than just a 128 long bool array?

JamesNK: If it changes, it will change to something like discussed at https://github.com/grpc/grpc-dotnet/pull/223#discussion_r279640292

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
JamesNK: Because this will force it to null if it was empty, and avoid writing a header with an empty value.

JamesNK: I guess what you suggest does the same thing. I suppose I like code that forces each path of the `if` to make a choice 🤷‍♂

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        // TODO(JamesNK): Remove WIP decode once we are certain Uri.UnescapeData works correctly
+#if Decode
+        public static string PercentDecode(string value)
```
jtattermusch: Looks like we only need the percent decoding for grpc client (which currently cannot access classes from Grpc.AspNetCore.Server anyway) and so far Uri.UnescapeDataString seems to work just fine, so let's remove the PercentDecode method entirely (no reason to check in dead code) and only add it once we are sure we actually need it (add a TODO to https://github.com/grpc/grpc-dotnet/blob/23099fefd3bc0353580cf5c7600716a9ea8f4306/src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs#L434 or file an issue to double check).
    
If we decide to add a custom PercentDecode method, there should be some test cases for "PercentDecode" too, including making sure that implementation conforms with this part of spec: *"When decoding invalid values, implementations MUST NOT error or throw away the message. At worst, the implementation can abort decoding the status message altogether such that the user would received the raw percent-encoded form. Alternatively, the implementation can decode valid portions while leaving broken %-encodings as-is or replacing them with a replacement character (e.g., '?' or the Unicode replacement character)."*

jtattermusch:
I did a manual check to verify this doesn't break status_code_and_message test against grpc-dart (like #216 did) and the tests have passed. We should be good on that front.

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
JunTaoLuo: if you want to keep empty, just initialize escapeDetail = status.Detail. Also, nit: tabs.

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
```diff
@@ -0,0 +1,218 @@
...
+                var writePosition = 0;
+                for (var i = 0; i < s.Length; i++)
+                {
+                    var current = s[i];
+                    if (current > AsciiMaxValue)
+                    {
+                        // Get additional unicode characters
+                        var unicodeCharCount = GetUnicodeCharCount(s, i);
+
+                        var numberOfBytes = Encoding.UTF8.GetBytes(s.AsSpan(i, unicodeCharCount), unicodeBytesBuffer);
```
GrabYourPitchforks: The behavior of this method is that invalid UTF-16 data (such as unpaired `'\uD800'` chars) will get converted to the UTF-8 byte sequence `[ EF BF BD ]`, then escaped as `"%EF%BF%BD"`. The receiver will decode it as the replacement char � (`'\uFFFD'`).

This is _probably_ appropriate behavior for this scenario, but I wanted to call it out explicitly just so you could confirm that was your intention.

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
```diff
@@ -0,0 +1,218 @@
...
+#if Decode
+        public static string PercentDecode(string value)
+        {
+            var writeLength = 0;
+            var readIndex = 0;
+            while (readIndex < value.Length)
+            {
+                if (value[readIndex] == '%')
+                {
+                    if (readIndex + 2 < value.Length && IsValidHex(value[readIndex + 1]) && IsValidHex(value[readIndex + 2]))
```
GrabYourPitchforks: Possible integer overflow: should instead read `if (readIndex < value.Length - 2 && ...)` or `if ((uint)(readIndex + 2) < (uint)value.Length && ...)`.

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
```diff
@@ -0,0 +1,218 @@
...
+            }
+
+            if (writeLength == value.Length)
+            {
+                return value;
+            }
+
+            return string.Create(writeLength, value, Decode);
+
+            static void Decode(Span<char> span, string s)
```
GrabYourPitchforks: The implementation of this method is incorrect. It's decoding individual bytes and widening those bytes to chars when producing the destination string. In effect, that means it's decoding data as ISO-8859-1. This is an asymmetry with the encoding method, which uses UTF-8.

Additionally, per Jan's comment at https://github.com/grpc/grpc-dotnet/pull/223/files#r280808299, if you want to be compliant with the spec you need to include error handling behavior here if you're given invalid input.

I personally believe an error should be returned in this case, even though the spec explicitly calls it out as a disallowed behavior. My opinion is that the current wording of the spec is bound to introduce security vulnerabilities into implementations which are trying to be fully compliant because it encourages such implementations to introduce complexity into code paths which process untrusted input. But if I am alone in this opinion I'm not going to force the argument.

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
JamesNK: This is intended. There are unit tests below that explicitly test for this - https://github.com/grpc/grpc-dotnet/pull/223/files#diff-4269317ee063a7595ca58194903e29f6R70

I'll add a comment here that discusses this possibility.

src/Grpc.AspNetCore.Server/Internal/PercentEncodingHelpers.cs
JamesNK: All the decode logic here is being replaced with Uri.UnescapeData

JamesNK:
@GrabYourPitchforks Thanks for the feedback, all resolved.

PR #230, opened 2019-05-03
Fix interop test client with TLS enabled

JamesNK:
Interop test client was not using TLS even when the use_tls was true.


testassets/InteropTestsClient/InteropClient.cs
```diff
@@ -768,5 +787,27 @@ private static Metadata CreateTestMetadata()
                 {"x-grpc-test-echo-trailing-bin", new byte[] {0xab, 0xab, 0xab}}
             };
         }
+
+        private byte[] GetBytesFromPem(string pemString, string section)
```
jtattermusch: looks like this is a hacky way to read the certificate?  use a library function or at least or add a TODO to implement this properly?

PR #233, opened 2019-05-06
Add C Core interop server

JamesNK:
Interop test server that uses C Core. This is for comparing ASP.NET Core vs C Core grpc server/client interaction without having to compile all of grpc/grpc.

Will likely be removed when .NET Core gRPC client passes all interop tests.


JamesNK:
I already have grpc/grpc compiling locally, but to work through HttpClient issues I will need to give reproduction steps to other developers, and it is easier for them to compile grpc/grpc-dotnet than to compile grpc/grpc with all its native components.

I would find it easier to have this available in all branches compared to one, and temporarily adding it and then removing it doesn't seem like a big deal. The temporarily addition is isolated to  testassets/InteropTestsNativeWebsite.

testassets/InteropTestsNativeWebsite/InteropServer.cs
```diff
@@ -0,0 +1,81 @@
+﻿#region Copyright notice and license
```
jtattermusch: nit: InteropTestsNativeWebsite -> InteropTestsNativeServer  (the Grpc.Core-based interop server is not a website).

jtattermusch: Add a readme.md to the InteropTestsNativeServer directory that will describe that this is a copy of InteropServer logic from grpc/grpc to allow for easier testing in the initial phase of managed client development and that we plan to remove it the future (how long are you planning to keep it)?

Also add a note that the authoritative C# interop server and the server here might diverge (e.g. if new functionality is added) in which case the Grpc.Core interop server code is the source of truth.

Once we have this readme, I think I'm fine with adding the native interop server.

PR #236, opened 2019-05-07
Add logging to client

JamesNK:
Add logging to client with extensions logging API
Output logs to console in interop tests


src/Grpc.NetCore.HttpClient/Internal/GrpcCallScope.cs
```diff
@@ -0,0 +1,72 @@
...
+#endregion
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using Grpc.Core;
+
+namespace Grpc.NetCore.HttpClient.Internal
+{
+    internal class GrpcCallScope : IReadOnlyList<KeyValuePair<string, object>>
```
JunTaoLuo: All this work to avoid an allocation of an IDictionary 😄

JamesNK: Plus the cached ToString for log writers.

JamesNK: Who is the logging expert these days? I'd like to double check I am using log scope correctly.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -302,51 +326,77 @@ private void ValidateHeaders()
...
             _callCts.Cancel();
 
             // Canceling call will cancel pending writes to the stream
             _completeTcs?.TrySetCanceled();
             _writeStreamTcs?.TrySetCanceled();
         }
 
+        internal IDisposable StartScope()
+        {
+            return Logger.BeginScope(_logScope);
```
davidfowl: check if the level is at least critical otherwise return null

src/Grpc.NetCore.HttpClient/Internal/GrpcCallScope.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        private readonly string _cachedToString;
+        private readonly MethodType _methodType;
+        private readonly Uri _uri;
+
+        public GrpcCallScope(MethodType methodType, Uri uri)
+        {
+            _methodType = methodType;
+            _uri = uri;
+            _cachedToString = FormattableString.Invariant($"{GrpcMethodTypeKey}:{methodType},{GrpcUriKey}:{uri}");
```
davidfowl: Make this lazy.

src/Grpc.NetCore.HttpClient/Internal/GrpcCallScope.cs
JunTaoLuo: Hmmm @brennanconroy?

BrennanConroy: This class looks good, but some of the usage is questionable.

For example:
https://github.com/grpc/grpc-dotnet/pull/236/files#diff-9ca9d0e546493f5532bd42e33e273dc1R389
You dispose the scope right after this, so any logs that happen after the method goes async wont have the scope attached to them.

JamesNK: > You dispose the scope right after this, so any logs that happen after the method goes async wont have the scope attached to them.

They do have the scope. This message that the response headers have been received happens after the call goes async and I see the scope in the console output:

```
trce: Grpc.NetCore.HttpClient.Internal.GrpcCall[2]
      => GrpcMethodType:ServerStreaming,GrpcUri:/grpc.testing.TestService/StreamingOutputCall
      Response headers received.
```

I'm not really sure how a scope works. I initially called logger.AddScope when the call started, cached the disposable response on GrpcCall, and then disposed it when the call ended, but that didn't work.

JamesNK: I'm going to guess it is an async local and async/await handles flowing the execution context.

JamesNK:
@jtattermusch Do you have any thoughts here? I'd like to have you double check the PR before I merge it.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -302,51 +340,91 @@ private void ValidateHeaders()
...
         }
 
         private void CancelCall()
         {
-            _callCts.Cancel();
+            // Checking if cancellation has already happened isn't threadsafe
+            // but there is no adverse effect other than an extra log message
+            if (!_callCts.IsCancellationRequested)
+            {
+                System.Diagnostics.Debugger.Launch();
```
jtattermusch: is this intended?

JamesNK: No. Good catch

PR #240, opened 2019-05-10
Fallback to reading status from headers

JamesNK:
Fix more interop test scenarios.

Double check that this is expected - https://github.com/grpc/grpc/issues/19003

### DIFF
--- a/src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
+++ b/src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
@@ -493,15 +493,28 @@ namespace Grpc.NetCore.HttpClient.Internal
 
         private static Status GetStatusCore(HttpResponseMessage httpResponseMessage)
         {
-            // A gRPC server may return trailers in the headers when the response stream is empty
+            // A gRPC server may return gRPC status in the headers when the response stream is empty
             // For example, C Core server returns them together in the empty_stream interop test
-            var grpcStatus = GetHeaderValue(httpResponseMessage.TrailingHeaders, GrpcProtocolConstants.StatusTrailer)
-                ?? GetHeaderValue(httpResponseMessage.Headers, GrpcProtocolConstants.StatusTrailer);
+            HttpResponseHeaders statusHeaders;
 
-            // grpc-status is a required trailer
-            if (grpcStatus == null)
+            var grpcStatus = GetHeaderValue(httpResponseMessage.TrailingHeaders, GrpcProtocolConstants.StatusTrailer);
+            if (grpcStatus != null)
             {
-                throw new InvalidOperationException("Response did not have a grpc-status trailer.");
+                statusHeaders = httpResponseMessage.TrailingHeaders;
+            }
+            else
+            {
+                grpcStatus = GetHeaderValue(httpResponseMessage.Headers, GrpcProtocolConstants.StatusTrailer);
+
+                // grpc-status is a required trailer
+                if (grpcStatus != null)
+                {
+                    statusHeaders = httpResponseMessage.Headers;
+                }
+                else
+                {
+                    throw new InvalidOperationException("Response did not have a grpc-status trailer.");
+                }
             }
 
             int statusValue;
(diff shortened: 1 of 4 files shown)


src/Grpc.NetCore.HttpClient/Internal/GrpcProtocolHelpers.cs
```diff
@@ -91,7 +91,12 @@ public static Metadata BuildMetadata(HttpResponseHeaders responseHeaders)
                 // ASP.NET Core includes pseudo headers in the set of request headers
                 // whereas, they are not in gRPC implementations. We will filter them
                 // out when we construct the list of headers on the context.
-                if (header.Key.StartsWith(":", StringComparison.Ordinal))
+                if (header.Key.StartsWith(':'))
+                {
+                    continue;
+                }
+                // Exclude grpc related headers
+                if (header.Key.StartsWith("grpc-", StringComparison.OrdinalIgnoreCase))
```
JunTaoLuo: Is this strictly necessary?

JamesNK: The C Core client excludes grpc headers from GetTrailers and ResponseHeadersAsync.

test/Grpc.NetCore.HttpClient.Tests/AsyncServerStreamingCallTests.cs
```diff
@@ -176,5 +169,34 @@ public void ClientStreamReader_WriteWithInvalidHttpStatus_ErrorThrown()
 
             Assert.AreEqual("Bad gRPC response. Expected HTTP status code 200. Got status code: 404", ex.Message);
         }
+
+        [Test]
+        public async Task AsyncServerStreamingCall_NoContent_TrailersReturnedWithHeaders()
```
jtattermusch: nit: let's call this case "TrailersOnly" as that's how the case is called in the spec.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
```diff
@@ -415,7 +415,11 @@ private void DeadlineExceeded(object state)
 
         private static Status GetStatusCore(HttpResponseMessage httpResponseMessage)
         {
-            string grpcStatus = GetHeaderValue(httpResponseMessage.TrailingHeaders, GrpcProtocolConstants.StatusTrailer);
+            // A gRPC server may return trailers in the headers when the response stream is empty
+            // For example, C Core server returns them together in the empty_stream interop test
+            var grpcStatus = GetHeaderValue(httpResponseMessage.TrailingHeaders, GrpcProtocolConstants.StatusTrailer)
+                ?? GetHeaderValue(httpResponseMessage.Headers, GrpcProtocolConstants.StatusTrailer);
```
jtattermusch: This is correct, but too permissive.
Looks like this can potentially allow a malformed RPC where  grpc-status is in the headers, but there is also payload and there are some trailers. In the "Trailers-Only case" from the spec, the `TrailingHeaders` need to be empty and there must be no payload. Let's add some check in that sense.

In c-core, it seems that we only allow "Trailers-only" case if  the headers frame being analyzed is the last frame of the stream.
https://github.com/grpc/grpc/blob/50e16f6d3a99335f09060d62cfa495c8977e3b2b/src/core/ext/transport/chttp2/transport/parsing.cc#L630
https://github.com/grpc/grpc/blob/50e16f6d3a99335f09060d62cfa495c8977e3b2b/src/core/ext/transport/chttp2/transport/parsing.cc#L553

test/Grpc.NetCore.HttpClient.Tests/AsyncServerStreamingCallTests.cs
jtattermusch: let's add a negative test case where grpc-status is added to the headers, but there's also a payload and actual trailers (where grpc-status is missing) and verify that scenario fails.

src/Grpc.NetCore.HttpClient/Internal/GrpcCall.cs
tuchschererd: Help me please I’ve gotten assistance from github before and now I have several questions but am very new to coding as I used to do it long time ago. Would appreciate input on where and how to start so I can participate and get my questions answered

JamesNK:
All feedback applied. I'd like to merge now to see if all interop tests pass.

If anything isn't as expected then further changes can be in a new PR

Issue #107, opened 2019-02-19
Improve discoverability of BindService


JamesNK:
Right now we use reflection to look up a service's type hierarchy to find `BindService`. This is brittle because other code gen tools could have placed `BindService` in a different location.

What about adding an attribute that is added via code-gen to the service base type that specifies exactly where to find the BindService method.

```cs
public class BindServiceAttribute : Attribute
{
    public BindServiceAttribute(Type type, string methodName)
    {
        // ...
    }
}
```

Code gen service:

```cs
public static class Counter
{
    [BindService(typeof(Counter), nameof(Counter.BindService))]
    public abstract partial class CounterBase
    {
    }

    public static void BindService(grpc::ServiceBinderBase serviceBinder, CounterBase serviceImpl)
    {
        // ...
    }
}
```

Other .NET code generators can then be encouraged to output this attribute so they can be more easily used with Grpc.AspNetCore.

JamesNK:
@davidfowl @jtattermusch

JamesNK:
I still like this idea as a better alternative to how we currently bind services in gRPC. The way we currently search for the bind services method using reflection is flaky.

@jtattermusch we can submit a PR to Grpc.Core.Api to add the attribute and Grpc.Tools to update the code gen. Grpc.AspNetCore.Server would behave the same as it does today when it can't find the attribute (search for bind via the inheritance tree) so no generated code that works today would break.

davidfowl:
Agreed.

JamesNK:
Example of calling bind method with reflection via an attribute:
```cs
var serviceAttribute = typeof(GreetService).GetCustomAttributes(typeof(BindServiceAttribute), inherit: true).FirstOrDefault();
if (serviceAttribute != null)
{
    var bindMethod = serviceAttribute.ServiceType.GetMethod(serviceAttribute.BindMethodName);
    bindMethod.Invoke(null);
}
else
{
    // Existing reflection logic for backwards compat
}
```

PR #242, opened 2019-05-12
Use BindServiceMethodAttribute to search for bind method

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/107

### DIFF
--- a/src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
+++ b/src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
@@ -90,38 +90,12 @@ namespace Microsoft.AspNetCore.Builder
             return new CompositeEndpointConventionBuilder(serviceBinder.EndpointConventionBuilders);
         }
 
-        private static Type GetServiceBaseType(Type serviceImplementation)
-        {
-            // TService is an implementation of the gRPC service. It ultimately derives from Foo.TServiceBase base class.
-            // We need to access the static BindService method on Foo which implicitly derives from Object.
-            var baseType = serviceImplementation.BaseType;
-
-            // Handle services that have multiple levels of inheritence
-            while (baseType?.BaseType?.BaseType != null)
-            {
-                baseType = baseType.BaseType;
-            }
-
-            return baseType;
-        }
-
         private static void ReflectionBind<TService>(ServiceBinderBase binder, TService service)
         {
-            var baseType = GetServiceBaseType(typeof(TService));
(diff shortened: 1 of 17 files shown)


src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
```diff
@@ -90,38 +90,12 @@ public static class GrpcEndpointRouteBuilderExtensions
...
-            // Handle services that have multiple levels of inheritence
-            while (baseType?.BaseType?.BaseType != null)
-            {
-                baseType = baseType.BaseType;
-            }
-
-            return baseType;
-        }
-
         private static void ReflectionBind<TService>(ServiceBinderBase binder, TService service)
```
JunTaoLuo: This is no longer just reflection bind. It's using attribute first but falling back to reflection.

test/Grpc.AspNetCore.Server.Tests/TestObjects/Services/WithAttribute/GreeterWithAttribute.cs
```diff
@@ -0,0 +1,70 @@
...
+            __Marshaller_Greet_HelloReply);
+
+        static readonly Method<global::Greet.HelloRequest, global::Greet.HelloReply> __Method_SayHellos = new Method<global::Greet.HelloRequest, global::Greet.HelloReply>(
+            MethodType.ServerStreaming,
+            __ServiceName,
+            "SayHellos",
+            __Marshaller_Greet_HelloRequest,
+            __Marshaller_Greet_HelloReply);
+
+        [BindServiceMethod(typeof(GreeterWithAttribute), "BindService")]
```
JunTaoLuo: Instead of marking this on the generated class file, I would imagine it would be marked on a separate file with a partial GreeterBase class.

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
JamesNK: Searching for an attribute is still reflection yo

test/Grpc.AspNetCore.Server.Tests/TestObjects/Services/WithAttribute/GreeterWithAttribute.cs
JamesNK: This file is purposefully similar to what Grpc.Tools generates

PR #247, opened 2019-05-14
Run managed client interop tests too

jtattermusch:
@JamesNK  @JunTaoLuo  I think this should do it.

### DIFF
--- a/kokoro/interop.sh
+++ b/kokoro/interop.sh
@@ -27,4 +27,4 @@ cd ../grpc
 
 source tools/internal_ci/helper_scripts/prepare_build_linux_rc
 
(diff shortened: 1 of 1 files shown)


jtattermusch:
Just FTR, the test reports seem to be indicating that both managed and native client are running fine: https://source.cloud.google.com/results/invocations/b27e29a7-1564-456d-a7d9-6d11b5d79fa1/targets/github%2Fgrpc%2Finterop_test/tests

(btw, we could add more languages to the suite running on this repo, it's basically a tradeoff between getting more signal and tests taking longer - also if we include other languages, we will be occasionally seeing test flakes coming from those other languages and not related to code changes in this repo).

Issue #226, opened 2019-05-02
Remove System.Net.Http.SocketsHttpHandler.Http2Support appcontext switch


JunTaoLuo:
This has been set to true by default so we don't have to do it anymore. Will require an ingestion of new SDK when it's available.

JamesNK:
Where/when was that change made? I tried removing the setting with a new SDK and it didn't work.

JunTaoLuo:
This change is now in master https://github.com/dotnet/corefx/blob/d943b4d090dc8e88cbf02e6e24702543c59c49fb/src/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionSettings.cs#L126-L127 but it doesn't seem to be available yet.

PR #253, opened 2019-05-16
Remove System.Net.Http.SocketsHttpHandler.Http2Support appcontext switch

JunTaoLuo:
Addresses https://github.com/grpc/grpc-dotnet/issues/226

As part of updating the dependencies, I also removed some of the scripts that won't be used anymore now that we have a nuget feed.
Finally, I was testing the greeter client and changed it to use HttpClient, but maybe it's a little premature?


examples/Clients/Greeter/Program.cs
```diff
@@ -31,16 +33,13 @@ class Program
         static async Task Main(string[] args)
         {
             // Server will only support Https on Windows and Linux
-            var credentials = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? ChannelCredentials.Insecure : ClientResources.SslCredentials;
-            var channel = new Channel("localhost:50051", credentials);
-            var client = new Greeter.GreeterClient(channel);
+            var certificate = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? null : new X509Certificate2(Resources.ServerPFXPath, "1111");
+            var client = GrpcClientFactory.Create<Greeter.GreeterClient>("https://localhost:50051", certificate: certificate);
```
JamesNK: Undo this. The samples should use the C Core client until duplex streaming is ready. Once that is done we can recommend people to use it.

Issue #229, opened 2019-05-03
Enable nullable reference types


JamesNK:
https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/nullable-reference-types

Any reason not to? Better to do sooner rather than later. The more code added means the more code to update later.

PR #254, opened 2019-05-16
Enable non-nullable reference types

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/229


JamesNK:
I would like to merge this soon rather than later. It touches a lot of files and will turn into **Conflict City** if it hangs around.

IMO you can save yourself some time and ignore the test/benchmark projects.

test/FunctionalTests/FunctionalTestBase.cs
```diff
@@ -26,9 +26,9 @@ namespace Grpc.AspNetCore.FunctionalTests
 {
     public class FunctionalTestBase
     {
-        private VerifyNoErrorsScope _scope;
+        private VerifyNoErrorsScope? _scope;
 
-        protected GrpcTestFixture<FunctionalTestsWebsite.Startup> Fixture { get; private set; }
```
JunTaoLuo: Non-nullability doesn't play well with [OneTimeSetUp] does it.

test/FunctionalTests/FunctionalTestBase.cs
```diff
@@ -26,9 +26,9 @@ namespace Grpc.AspNetCore.FunctionalTests
 {
     public class FunctionalTestBase
     {
-        private VerifyNoErrorsScope _scope;
+        private VerifyNoErrorsScope? _scope;
```
JunTaoLuo: Why is _scope nullable? We always expect that it will be initialized via Setup right?

test/Grpc.AspNetCore.Server.Tests/DefaultGrpcInterceptorActivatorTests.cs
```diff
@@ -77,7 +77,7 @@ public void GrpcInterceptorCanResolveArgumentsFromArg()
                 .Setup(sp => sp.GetService(typeof(Mutex)))
                 .Returns(mutex);
 
-            var interceptor = new DefaultGrpcInterceptorActivator<GrpcInterceptor>(mockServiceProvider.Object).Create(10) as GrpcInterceptor;
+            var interceptor = (GrpcInterceptor)new DefaultGrpcInterceptorActivator<GrpcInterceptor>(mockServiceProvider.Object).Create(10);
```
JunTaoLuo: I'm wondering if IGrpcInterceptorActivator<T> should have `T Create()` and `Release(T)` instead of just the base interface IGrpcInterceptor's `Interceptor Create()` and `Release(Interceptor )` methods?

test/Grpc.AspNetCore.Server.Tests/HttpContextStreamWriterTests.cs
```diff
@@ -107,9 +106,9 @@ public async Task WriteAsync_BufferHintWriteOptions_DoesNotFlush()
             var pipeReader = new StreamPipeReader(ms);
 
             var writtenMessage1 = await MessageHelpers.AssertReadStreamMessageAsync<HelloReply>(pipeReader);
-            Assert.AreEqual("Hello world 1", writtenMessage1.Message);
+            Assert.AreEqual("Hello world 1", writtenMessage1!.Message);
```
JunTaoLuo: Should this be ? instead of ! 

It seems to me the AssertReadStreamMessageAsync's API's intention is to return null if no message is read. In that case we should use ? right? 

This appears in a lot of places.

test/Grpc.AspNetCore.Server.Tests/Infrastructure/TaskExtensions.cs
```diff
@@ -1,95 +1,95 @@
-﻿#region Copyright notice and license
+﻿//#region Copyright notice and license
```
JunTaoLuo: What's this about?

test/Grpc.AspNetCore.Server.Tests/Infrastructure/TestHttpContextAccessor.cs
```diff
@@ -22,6 +22,6 @@ namespace Grpc.AspNetCore.Server.Tests.Infrastructure
 {
     internal class TestHttpContextAccessor : IHttpContextAccessor
     {
-        public HttpContext HttpContext { get; set; }
+        public HttpContext? HttpContext { get; set; }
```
JunTaoLuo: I think we can use a constructor for this and make it a non-nullable. There's not much use use otherwise it seems.

test/Grpc.AspNetCore.Server.Tests/TestObjects/TestEndpointRouteBuilder.cs
```diff
@@ -25,7 +25,7 @@ namespace Grpc.AspNetCore.Server.Tests.TestObjects
 {
     public class TestEndpointRouteBuilder : IEndpointRouteBuilder
     {
-        public IServiceProvider ServiceProvider { get; set; }
+        public IServiceProvider? ServiceProvider { get; set; }
```
JunTaoLuo: Based on usage, it seems like we would also like for this property to never be null either. Use a constructor?

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
```diff
@@ -109,21 +110,21 @@ public static bool IsGrpcContentType(string contentType)
             return false;
         }
 
-        public static bool IsValidContentType(HttpContext httpContext, out string error)
+        public static bool IsInvalidContentType(HttpContext httpContext, [NotNullWhenTrue]out string? error)
```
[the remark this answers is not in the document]
JunTaoLuo: Oh is it to use NotNullWhenTrue?

src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
```diff
@@ -178,7 +178,7 @@ private GrpcContentTypeConstraint()
```
JunTaoLuo: The `AddMethod` methods are causing errors in VS:
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS8634	The type 'TResponse' cannot be used as type parameter 'TResponse' in the generic type or method 'ServerCallHandlerFactory<TService>.CreateUnary<TRequest, TResponse>(Method<TRequest, TResponse>, UnaryServerMethod<TService, TRequest, TResponse>)'. Nullability of type argument 'TResponse' doesn't match 'class' constraint.	Grpc.AspNetCore.Server	C:\gh\grpc-dotnet\src\Grpc.AspNetCore.Server\Internal\GrpcServiceBinder.cs	86	Active
```

I think you need to add more constraints.

src/Grpc.NetCore.HttpClient/HttpClientCallInvoker.cs
```diff
@@ -75,7 +75,7 @@ public HttpClientCallInvoker(System.Net.Http.HttpClient client, ILoggerFactory l
         /// <summary>
         /// Token for propagating parent call context.
         /// </summary>
-        public ContextPropagationToken PropagationToken { get; set; }
+        public ContextPropagationToken? PropagationToken { get; set; }
```
JunTaoLuo: There's also several errors in this file when building in VS:
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS8634	The type 'TRequest' cannot be used as type parameter 'TRequest' in the generic type or method 'HttpClientCallInvoker.CreateGrpcCall<TRequest, TResponse>(Method<TRequest, TResponse>, CallOptions)'. Nullability of type argument 'TRequest' doesn't match 'class' constraint.	Grpc.NetCore.HttpClient	C:\gh\grpc-dotnet\src\Grpc.NetCore.HttpClient\HttpClientCallInvoker.cs	86	Active
```

src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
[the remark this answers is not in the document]
JamesNK: Yes. You can have a NotNullWhenFalse but I inverted it rather than add another attribute.

test/Grpc.AspNetCore.Server.Tests/HttpContextStreamWriterTests.cs
JamesNK: I know in the test it is not null so this is to suppress the warning. It *could be* ? instead but meh. Either way it fails on null.

test/Grpc.AspNetCore.Server.Tests/Infrastructure/TaskExtensions.cs
JamesNK: Forgot to remove

test/Grpc.AspNetCore.Server.Tests/Infrastructure/TestHttpContextAccessor.cs
JamesNK: Sure. It doesn't really matter either way

test/Grpc.AspNetCore.Server.Tests/TestObjects/TestEndpointRouteBuilder.cs
JamesNK: Sure. It doesn't really matter either way

test/FunctionalTests/DeadlineTests.cs
```diff
@@ -148,7 +148,7 @@ static async Task WriteUntilError(HelloRequest request, IServerStreamWriter<Hell
                 return writeContext.LoggerName == typeof(DeadlineTests).FullName &&
                        writeContext.EventId.Name == "ErrorExecutingServiceMethod" &&
                        writeContext.State.ToString() == "Error when executing service method 'WriteUntilError'." &&
-                       writeContext.Exception.Message == "Cannot write message after request is complete.";
+                       writeContext.Exception!.Message == "Cannot write message after request is complete.";
```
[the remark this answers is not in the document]
JamesNK: Should be nullable. Not every log message has an exception.

src/Grpc.AspNetCore.Server/Internal/NotNullWhenTrueAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+namespace System.Runtime.CompilerServices
+{
+    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false)]
+    internal class NotNullWhenTrueAttribute : Attribute
```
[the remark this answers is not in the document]
JamesNK: What do you mean declare? The only thing that matters is the attribute's name and namespace.

I don't know if Roslyn will ship it. If it does then this will start generating a duplicate type error on compile and it can be fixed then.

test/FunctionalTests/FunctionalTestBase.cs
JamesNK: No

test/FunctionalTests/FunctionalTestBase.cs
JamesNK: Because it isn't set in the constructor. If someone did not call SetUp then it would be null.

src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
JamesNK: It builds on my computer and the build server. What is special to your computer that fails?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -309,7 +309,7 @@ public void Dispose()
             _deadlineTimer?.Dispose();
         }
 
-        internal string GetRequestGrpcEncoding()
+        internal string? GetRequestGrpcEncoding()
```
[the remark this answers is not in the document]
JamesNK: Checking the nullable result and throwing for null automatically turns it back to a non-null reference type. This is fine.

src/Grpc.AspNetCore.Server/GrpcBindingOptions.cs
```diff
@@ -25,14 +25,14 @@ namespace Grpc.AspNetCore.Server
...
     /// Options used to configure the binding of a gRPC service.
     /// </summary>
-    public class GrpcBindingOptions<TService>
+    public class GrpcBindingOptions<TService> where TService : class
     {
         /// <summary>
         /// The action invoked to get service metadata via <see cref="ServiceBinderBase"/>.
         /// </summary>
-        public Action<ServiceBinderBase, TService> BindAction { get; set; }
+        public Action<ServiceBinderBase, TService?>? BindAction { get; set; }
```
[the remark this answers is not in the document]
JamesNK: It's a public type. Adding a constructor will become problematic if more public properties are added and new constructors need to be added to keep things not null.

test/Grpc.AspNetCore.Server.Tests/DefaultGrpcInterceptorActivatorTests.cs
JamesNK: Because then you need to create a generic type and use reflection to call it in the call handlers.

src/Grpc.AspNetCore.Server/Internal/GrpcServiceBinder.cs
JunTaoLuo: Yea it always built from the command line. It seems like my VS was out of date, updating to 16.1 preview 3 fixed it.

PR #265, opened 2019-05-24
Make IGrpcMethodModelFactory public to allow non-reflection based method binding.

ASuurkuusk:
We have just published a preview of a code first implementation of gRPC at [SciTechSoftware/SciTech.Rpc](/SciTechSoftware/SciTech.Rpc). It is based on grpc-dotnet for the ASP.NET Core server side implementation. To allow gRPC HTTP/2 calls to be forwarded to our implementation methods we need to provide our own GrpcEndpointModels. Unfortunately, GrpcEndpointModel and IGrpcMethodModelFactory are internal in grpc-dotnet.

This pull request changes the accessibility from internal to public for IGrpcMethodModelFactory and related types and properties. Considering the comment on GrpcBindingOptions.ModelFactory ("Currently internal. It is set in tests via InternalVisibleTo. Can be made public if there is demand for it"), you may be willing to make it public.

You can see how SciTech.Rpc implements IGrpcMethodModelFactory in [NetGrpcEndpointRouteBuilderExtensions.cs](/SciTechSoftware/SciTech.Rpc/blob/master/src/SciTech.Rpc.NetGrpc/NetGrpc/Server/NetGrpcEndpointRouteBuilderExtensions.cs)

The IGrpcMethodModelFactory and GrpcEndpointModel  types are still in the "Internal" namespace. If you believe they should be moved out of "Internal" due to this request, I can modify the pull request to do that.

Best regards,

Andreas Suurkuusk
SciTech Software AB

### DIFF
--- a/src/Grpc.AspNetCore.Server/GrpcBindingOptions.cs
+++ b/src/Grpc.AspNetCore.Server/GrpcBindingOptions.cs
@@ -32,7 +32,9 @@ namespace Grpc.AspNetCore.Server
         /// </summary>
         public Action<ServiceBinderBase, TService?>? BindAction { get; set; }
 
-        // Currently internal. It is set in tests via InternalVisibleTo. Can be made public if there is demand for it
-        internal IGrpcMethodModelFactory<TService>? ModelFactory { get; set; }
+        /// <summary>
+        /// Gets or sets a factory for creating models that describe and invoke service methods on .NET types.
+        /// </summary>
+        public IGrpcMethodModelFactory<TService>? ModelFactory { get; set; }
     }
 }
--- a/src/Grpc.AspNetCore.Server/Internal/IGrpcMethodModelFactory.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/IGrpcMethodModelFactory.cs
@@ -26,23 +26,71 @@ namespace Grpc.AspNetCore.Server.Internal
     /// An interface for creating models that describe and invoke service methods on .NET types.
     /// </summary>
     /// <typeparam name="TService">The service type.</typeparam>
-    internal interface IGrpcMethodModelFactory<TService>
+    public interface IGrpcMethodModelFactory<TService>
     {
+        /// <summary>
+        /// Creates a model that describe and invoke a unary operation.
+        /// </summary>
+        /// <typeparam name="TRequest">The request type.</typeparam>
+        /// <typeparam name="TResponse">The response type.</typeparam>
+        /// <param name="method">Unary remote method description.</param>
+        /// <returns>A model that describe and invoke a unary operation.</returns>
         GrpcEndpointModel<UnaryServerMethod<TService, TRequest, TResponse>> CreateUnaryModel<TRequest, TResponse>(Method<TRequest, TResponse> method);
+
+
+        /// <summary>
+        /// Creates a model that describe and invoke a client streaming operation.
+        /// </summary>
+        /// <typeparam name="TRequest">The request type.</typeparam>
+        /// <typeparam name="TResponse">The response type.</typeparam>
+        /// <param name="method">Client streaming remote method description.</param>
+        /// <returns>A model that describe and invoke a client streaming operation.</returns>
         GrpcEndpointModel<ClientStreamingServerMethod<TService, TRequest, TResponse>> CreateClientStreamingModel<TRequest, TResponse>(Method<TRequest, TResponse> method);
+
+        /// <summary>
+        /// Creates a model that describe and invoke a server streaming operation.
+        /// </summary>
+        /// <typeparam name="TRequest">The request type.</typeparam>
+        /// <typeparam name="TResponse">The response type.</typeparam>
+        /// <param name="method">Server streaming remote method description.</param>
+        /// <returns>A model that describe and invoke a server streaming operation.</returns>
         GrpcEndpointModel<ServerStreamingServerMethod<TService, TRequest, TResponse>> CreateServerStreamingModel<TRequest, TResponse>(Method<TRequest, TResponse> method);
+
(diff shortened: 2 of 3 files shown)


JamesNK:
Hi

I don't like the current way of customizing how methods are discovered. I've opened a PR that changes the process completely. There is no more IGrpcMethodModelFactory.

Could you take a look at it and see whether you can extend gRPC using this model.

https://github.com/grpc/grpc-dotnet/pull/268

ASuurkuusk:
Thank you for looking into this. I have updated our project to use the new IServiceMethodProvider in PR #268. It is definitely a better and more straight-forward way of  customizing gRPC method implementations compared to IGrpcMethodModelFactory. Previously I had to associate a hidden invoker delegate with the method definition when calling GrpcServiceBinder.AddMethod (since the handler provided to AddMethod was ignored). I did this by deriving classes from Grpc.Core.Method<,>, which felt like a hack.

If you are interested in how IServiceMethodProvider is used, you can look at the files in the directories [/src/SciTech.Rpc.NetGrpc/NetGrpc/Server](/SciTechSoftware/SciTech.Rpc/tree/method-discovery/src/SciTech.Rpc.NetGrpc/NetGrpc/Server/) and[/src/SciTech.Rpc.NetGrpc/NetGrpc/Server/Internal](/SciTechSoftware/SciTech.Rpc/tree/method-discovery/src/SciTech.Rpc.NetGrpc/NetGrpc/Server/Internal/) in the "method-discovery" branch of [SciTech.Rpc](/SciTechSoftware/SciTech.Rpc).

Since SciTech.Rpc has its own lookup of service instances, it will only map a single service type using MapGrpcService (a NetGrpcServiceActivator service). However, the associated IServiceMethodProvider will bind gRPC methods with several different service names that are all mapped to NetGrpcServiceActivator. As far as I can see this should not cause any problems, but this may be something that you don't expect. Will this work well in the future as well?

If you merge PR #268 you can close this PR.

Best regards,

Andreas Suurkuusk
SciTech Software AB

JamesNK:
> Will this work well in the future as well?

I couldn't guarantee anything. What you're doing is kind of a hack. Some logic is still being executed every request to create the service instance - NetGrpcServiceActivator in your case - and pass it to the delegate. You just happen to be ignoring it and using a type you create yourself.

A more integrated solution would be to register the RPC types with ASP.NET Core DI:

```cs
services.AddSingleton<GreeterServiceImpl, IGreeterService>();
services.AddSingleton<MailBoxManager, IMailBoxManagerService>();
```

Then in UseEndpoints you could map a gRPC endpoint for each interface:

```
app.UseEndpoints(endpoints =>
{
    endpoints.MapGrpcService<IGreeterService>();
    endpoints.MapGrpcService<IMailBoxManagerService>();
});
```

The logic you inject into each delegate would then adapt from ServiceCallContext/reader/writer to calling the service instance that was passed in. I guess you are probably doing this already.

For creating the service instance, `DefaultGrpcServiceActivator<T>` will resolve from DI first so it will handle creating `GreeterServiceImpl` singleton for `IGreeterService`.

ASuurkuusk:
We're not ignoring the NetGrpcServiceActivator that your gRPC implementation has created and we definitely want your logic behind the request (otherwise we would have used IEndpointRouteBuilder.MapPost directly). The service activator is used to look up the actual SciTech.RPC implementation instance. It may use the ASP.NET Core service provider, or it may use some other mechanism depending on how the service is published. 

SciTech.Rpc is designed as an alternative to WCF and .NET Remoting (primarily for .NET Core) and to make the .NET use of gRPC more similar to WCF and .NET remoting. It offers ways of publishing remote instances that cannot be resolved solely using dependency injection, for instance it is possible to publish non-singleton instances which are identified by an object id. In this case, our service activator has to investigate the request message to route the call to the correct instance.

From your side the NetGrpcServiceActivator service should look like an ordinary gRPC service implementation, but with the difference that multiple service names may be used when binding the methods, which I agree may seem like a hack. I'm working on a new implementation where each call to MapGrpcService will only bind methods for a single service name. For example:

```csharp
app.UseEndpoints(endpoints =>
{
    // Will bind service methods under the "GreeterService" service name
    endpoints.MapGrpcService<NetGrpcServiceActivator<IGreeterService>>();
    // Will bind service methods under the "MailBoxManagerService" service name
    endpoints.MapGrpcService<NetGrpcServiceActivator<IMailBoxManagerService>>();
});
```
However, this also means that we have to register an IServiceMethodProvider and a service activator for each service, and not just one common IServiceMethodProvider and one common activator. For example:

```csharp
services.AddSingleton<IServiceMethodProvider<NetGrpcServiceActivator<IGreeterService>>, 
    NetGrpcServiceMethodProvider<IGreeterService>>();
services.AddScoped(s => new NetGrpcServiceActivator<IGreeterService>(s));

services.AddSingleton<IServiceMethodProvider<NetGrpcServiceActivator<IMailBoxManagerService>>, 
    NetGrpcServiceMethodProvider<IMailBoxManagerService>>();
services.AddScoped(s => new NetGrpcServiceActivator<IMailBoxManagerService>(s));
```

Is this an implementation you would think work better, and be better supported in future releases of grpc-dotnet?

JamesNK:
You can add an open generic registration to `IServiceMethodProvider<T>`:

```cs
services.TryAddEnumerable(ServiceDescriptor.Singleton(typeof(IServiceMethodProvider<>), typeof(NetGrpcServiceMethodProvider<>)));
```

In `NetGrpcServiceMethodProvider<T>` you would then check whether the type is discoverable, e.g. does the generic argument have a `NetGrpcServiceActivator<T>` value.

JamesNK:
The service method provider has been merged. This PR is no longer needed.

PR #266, opened 2019-05-24
Add factory deadline and cancellation token propagation

JamesNK:


src/Grpc.NetCore.HttpClient/HttpClientCallInvoker.cs
```diff
@@ -156,11 +152,44 @@ public HttpClientCallInvoker(System.Net.Http.HttpClient client, ILoggerFactory?
...
+            if (Deadline < options.Deadline)
+            {
+                options = options.WithDeadline(Deadline);
+            }
+
+            if (CancellationToken.CanBeCanceled)
+            {
+                if (options.CancellationToken.CanBeCanceled)
+                {
+                    // If both propogated and options cancellation token can be canceled
```
JunTaoLuo: nit: propagated

test/Grpc.AspNetCore.Server.ClientFactory.Tests/GrpcClientServiceExtensionsTests.cs
```diff
@@ -0,0 +1,92 @@
...
+using System;
+using System.Security.Cryptography.X509Certificates;
+using Greet;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using NUnit.Framework;
+
+namespace Grpc.AspNetCore.Server.ClientFactory.Tests
+{
+    [TestFixture]
```
JunTaoLuo: nit: Why is this attribute needed? I don't see any setup code. Not a big deal since we do this everywhere though.

JamesNK: It is required by nunit on test classes

JunTaoLuo: Should be optional if there are no setup code. I tried removing it on some tests and the tests still run.

JamesNK: https://nunit.org/docs/2.6/testFixture.html

> Beginning with NUnit 2.5, the TestFixture attribute is optional for non-parameterized, non-generic fixtures. So long as the class contains at least one method marked with the Test, TestCase or TestCaseSource attribute, it will be treated as a test fixture.

Hmm. I'm going to be old school and say we should keep using it everywhere.

PR #269, opened 2019-05-27
Remove NUnit dependency from InteropClient

JamesNK:
An NUnit dependency in a non-test project was causing an error in the VS2019 test discovery.

Replace NUnit with some basic custom asserts.


JamesNK:
This is the best solution I know of. There is the Trace.Assert(bool) method in System.Diagnostics but it pops up a dialog when an error is thrown which isn't what we want.

Issue #234, opened 2019-05-06
gRPC dotnet global tool


JamesNK:

shirhatti:
## Overview

This file documents the functionality of CLI tooling for adding Service Reference to projects. The tooling will be provided as a global dotnet tool. Each service reference type will have its own CLI sub-command.

## gRPC

### Functionality

Add the following package references to the project if not present:

- Google.Protobuf
- Grpc.AspNetCore.Server
- Grpc.Tools (With the metadata of "PrivateAssets=All")

Add the gRPC file to the @(Protobuf) ItemGroup in the project. If a URL is specified, download the protobuf file to disk first.
If a URL is specified and the --reflection-api options is specified, use the reflection API to obtain the protobuf file.
Remote URL(s) need to be saved as an attribute on the protobuf reference added. An additional attribute needs to be specified if reflection API was used to download the file.

### Usage

```
Usage: dotnet grpc <PROJECT> add [options] <PROTOBUF>
Usage: dotnet grpc <PROJECT> remove [options] <PROTOBUF>
Usage: dotnet grpc <PROJECT> refresh [options] <PROTOBUF>
```

Note: refresh command is only applicable for URL based protobuf files.

### Arguments

```
Arguments:
  <PROJECT>        The project file to operate on. If a file is not specified, the command will search the current directory for one.
  <PROTOBUF>       The protobuf file(s) to add. These can be a path to glob for local protobuf file(s), URL(s) to remote protobuf file(s) or URL(s) to gRPC endpoints with Reflection. Wild cards are supported for paths. // Arguments to the refresh command must be URLs if specified but could also be empty.
```

### Options

All commands:
```
Options:
  -h, --help                            Show command line help.
```

Add commands:
```
Options:
  -h, --help                            Show command line help.
  -n, --no-restore                      Add the reference without performing restore preview and compatibility check.
  -s, --services <SERVICE_TYPE>         The type of gRPC services that should be generated. Valid values are: Both, Server (Default?), Client, None.
  --use-reflection                      The Reflection API should be used to obtain the protobuf file from the URL(s).
  --download-dir <DIR>                  The directory to download the remote protobuf file to for the provided URL(s).
  --access <ACCESS_MODIFIER>            The access modifier to use for the generated C# classes. Valid values are: Public (Default), Internal.
  -a, --additional-import-dirs <DIRS>   Additional directories to be used when resolving imports for the protobuf files.
[...]

PR #271, opened 2019-05-28
gRPC ServiceReference tool

JunTaoLuo:
Addresses https://github.com/grpc/grpc-dotnet/issues/234

Sample help text:
```console
(grpc-dotnet) C:\gh\tp\NewConsole2> dotnet grpc --help
Usage:
  dotnet-grpc [options] [command]

Options:
  --version    Display version information

Commands:
  add-file <files>        Add protobuf file reference(s) to the gRPC project.
  add-url <url>           Add a protobuf url reference to the gRPC project.
  refresh <references>    Check remote protobuf references(s) for updates and replace them if a newer version is available. If no file or url is provided, all remote protobuf files will be updated.
  remove <references>     Remove protobuf references(s).
```
```console
(grpc-dotnet) C:\gh\tp\NewConsole2> dotnet grpc add-file --help
add-file:
  Add protobuf file reference(s) to the gRPC project.

Usage:
  dotnet-grpc add-file [options] <files>

Arguments:
  <files>    The protobuf file reference(s). These can be a path to glob for local protobuf file(s).

Options:
  -p, --project <project>                     The path to the project file to operate on. If a file is not specified, the command will search the current directory for one.
  -s, --services <Both|Client|None|Server>    The type of gRPC services that should be generated. Valid values are: Both (Default), Server, Client, None.
  -a, --additional-import-dirs <dirs>         Additional directories to be used when resolving imports for the protobuf files. This is a semicolon separated list of paths.
  --access <Internal|Public>                  The access modifier to use for the generated C# classes. Valid values are: Public (Default), Internal.
```
```console
(grpc-dotnet) C:\gh\tp\NewConsole2> dotnet grpc add-url --help
add-url:
  Add a protobuf url reference to the gRPC project.

Usage:
  dotnet-grpc add-url [options] <url>

Arguments:
  <url>    The URL to a remote protobuf file.

Options:
  -o, --output <path>                         Specify the download path for the remote protobuf file. This is a required option.
  -p, --project <project>                     The path to the project file to operate on. If a file is not specified, the command will search the current directory for one.
  -s, --services <Both|Client|None|Server>    The type of gRPC services that should be generated. Valid values are: Both (Default), Server, Client, None.
  -a, --additional-import-dirs <dirs>         Additional directories to be used when resolving imports for the protobuf files. This is a semicolon separated list of paths.
  --access <Internal|Public>                  The access modifier to use for the generated C# classes. Valid values are: Public (Default), Internal.
```
```console
(grpc-dotnet) C:\gh\tp\NewConsole2> dotnet grpc refresh --help
refresh:
  Check remote protobuf references(s) for updates and replace them if a newer version is available. If no file or url is provided, all remote protobuf files will be updated.

Usage:
  dotnet-grpc refresh [options] <references>

Arguments:
  <references>    The URL(s) or file path(s) to remote protobuf references(s) that should be updated. Leave this argument empty to refresh all remote references.

Options:
  -p, --project <project>    The path to the project file to operate on. If a file is not specified, the command will search the current directory for one.
  --dry-run                  Output a list of file(s) that will be updated without downloading any new content.
```
```console
(grpc-dotnet) C:\gh\tp\NewConsole2> dotnet grpc remove --help
remove:
  Remove protobuf references(s).

Usage:
  dotnet-grpc remove [options] <references>

Arguments:
  <references>    The URL(s) or file path(s) of the protobuf references to remove.

Options:
  -p, --project <project>    The path to the project file to operate on. If a file is not specified, the command will search the current directory for one.
  --remove-file              Delete the protobuf file that was referenced from disk.
```

TODOs:
- Add more tests


src/dotnet-grpc/dotnet-grpc.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <PackageProjectUrl>https://github.com/grpc/grpc-dotnet</PackageProjectUrl>
+    <PackageTags>gRPC RPC CLI</PackageTags>
+    <VersionPrefix>$(GrpcDotnetVersion)</VersionPrefix>
+
+    <!-- Ignore missing XML documentation -->
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <OutputType>exe</OutputType>
+    <!--<PackAsTool>true</PackAsTool>-->
```
JunTaoLuo: This need to be uncommented to ship. Currently disabled to allow for easier debugging.

Directory.Build.props
```diff
@@ -13,6 +13,7 @@
 
     <!-- static local methods are a C# 8.0 feature -->
     <LangVersion>preview</LangVersion>
+    <NullableContextOptions>enable</NullableContextOptions>
```
JunTaoLuo: Added this since I was switching between older and newer compilers. Will remove.

src/dotnet-grpc/Commands/AddFileCommand.cs
```diff
@@ -0,0 +1,79 @@
...
+            command.AddOption(CommonOptions.ServiceOption());
+            command.AddOption(CommonOptions.AdditionalImportDirsOption());
+            command.AddOption(CommonOptions.AccessOption());
+
+            command.Handler = CommandHandler.Create<IConsole, FileInfo, Services, Access, string, string[]>(new AddFileCommand().AddFile);
+
+            return command;
+        }
+
+        public int AddFile(IConsole console, FileInfo? project, Services services, Access access, string additionalImportDirs, string[] files)
```
JamesNK: It feels odd that console is passed in on a method and set on a property. I know you did it to avoid having to pass it around everywhere, and this method should only ever be called once.

src/dotnet-grpc/Commands/AddFileCommand.cs
```diff
@@ -0,0 +1,79 @@
...
+                    Name = "files",
+                    Description = "The protobuf file reference(s). These can be a path to glob for local protobuf file(s).",
+                });
+
+            command.AddOption(CommonOptions.ProjectOption());
+            command.AddOption(CommonOptions.ServiceOption());
+            command.AddOption(CommonOptions.AdditionalImportDirsOption());
+            command.AddOption(CommonOptions.AccessOption());
+
+            command.Handler = CommandHandler.Create<IConsole, FileInfo, Services, Access, string, string[]>(new AddFileCommand().AddFile);
```
JamesNK: You could add an `Initialize(IConsole, FileInfo)` to CommandBase. It would handle setting the console and project file. It would be called here. You would need to move the try/catch to return 1 here as well.

Alternatively you could do the initialization in the command constructors. That would allow them to be marked as not null.

The downside would be specifying an anonymous delegate rather than passing a method directly like you're currently doing.

What do you think? There are pros and cons.

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,358 @@
...
+
+            if (removeFile)
+            {
+                File.Delete(Path.IsPathRooted(protobufRef.UnevaluatedInclude) ? protobufRef.UnevaluatedInclude : Path.Combine(Project.DirectoryPath, protobufRef.UnevaluatedInclude));
+            }
+        }
+
+        public static bool IsUrl(string reference)
+        {
+            return reference.StartsWith("http") && Uri.TryCreate(reference, UriKind.Absolute, out var _);
```
JamesNK: Call TryCreate first and then you could test for http with Uri.Scheme.

https://docs.microsoft.com/en-us/dotnet/api/system.uri.scheme?view=netframework-4.8

Also, StartsWith should ignore culture and be case insensitive. Add a test for `HTTP://LOCALHOST/BLAH.PROTO`

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,358 @@
...
+                return;
+            }
+
+            if (!dryRun)
+            {
+                using (var stream = await HttpClient.GetStreamAsync(url))
+                using (var fileStream = File.Open(destination, FileMode.Create, FileAccess.Write))
+                {
+                    await stream.CopyToAsync(fileStream);
+                    fileStream.Flush();
```
JamesNK: FlushAsync or does it not matter? Add a comment if it doesn't matter

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,358 @@
...
+            {
+                throw new InvalidOperationException("Internal error: Project not set.");
+            }
+
+            if (Console == null)
+            {
+                throw new InvalidOperationException("Internal error: Console not set.");
+            }
+
+            var packageReference = Project.GetItems("PackageReference").SingleOrDefault(i => i.UnevaluatedInclude == packageName);
```
JamesNK: `i.UnevaluatedInclude == packageName` does culture or case matter here?

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,358 @@
...
+            }
+
+            if (!File.Exists(Path.IsPathRooted(file) ? file : Path.Join(Project.DirectoryPath, file)))
+            {
+                throw new CLIToolException($"The reference {file} does not exist.");
+            }
+
+            if (!Project.GetItems("Protobuf").Any(i => i.UnevaluatedInclude == file))
+            {
+                if (Path.GetExtension(file) != ".proto")
```
JamesNK: Culture/case?

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,358 @@
...
+                }
+
+                if (!string.IsNullOrEmpty(additionalImportDirs))
+                {
+                    newItem.Xml.AddMetadata("AdditionalImportDirs", additionalImportDirs, expressAsAttribute: true);
+                }
+
+                if (!string.IsNullOrEmpty(url))
+                {
+                    newItem.Xml.AddMetadata("SourceURL", url);
```
JamesNK: Is SourceURL something we have come up with? Why SourceURL instead of SourceUrl?

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,358 @@
...
+            }
+
+            var resolvedReferences = new List<ProjectItem>();
+            var protobufItems = Project.GetItems("Protobuf");
+
+            foreach (var reference in GlobReferences(references))
+            {
+                if (IsUrl(reference))
+                {
+                    var remoteItem = protobufItems.SingleOrDefault(p => p.GetMetadataValue("SourceURL") == reference);
```
JamesNK: `p.GetMetadataValue("SourceURL") == reference`

Culture/case?

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,358 @@
...
+                    {
+                        Console.Out.WriteLine($"Warning: Could not find a reference that uses the source url `{reference}`.");
+                        continue;
+                    }
+
+                    resolvedReferences.Add(remoteItem);
+                    continue;
+                }
+
+                var localItem = protobufItems.SingleOrDefault(p => p.UnevaluatedInclude == reference);
```
JamesNK: `p.UnevaluatedInclude == reference`

Culture/case?

src/dotnet-grpc/Commands/RefreshCommand.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        public async Task<int> Refresh(IConsole console, FileInfo? project, bool dryRun, string[] references)
+        {
+            Console = console;
+
+            try
+            {
+                Project = ResolveProject(project);
+                var refsToRefresh = references.Length == 0 ? Project.GetItems("Protobuf").Where(p => p.HasMetadata("SourceURL")) : ResolveReferences(references);
```
JamesNK: Consider moving "Protobuf", "SourceURL", etc to a constants file.

src/dotnet-grpc/Commands/AddFileCommand.cs
JunTaoLuo: I like the constructor approach.

Directory.Build.props
JunTaoLuo: We have moved to the newest compiler now. This should no longer be required.

src/dotnet-grpc/Commands/CommandBase.cs
JunTaoLuo: This name is pilfered from https://github.com/aspnet/AspNetCore/pull/10015/files. But yea I should change the casing.

test/dotnet-grpc.Tests/CommandBaseTests.cs
```diff
@@ -0,0 +1,367 @@
...
+using Microsoft.Build.Evaluation;
+using Microsoft.Build.Locator;
+using NUnit.Framework;
+
+namespace Grpc.Dotnet.Cli.Tests
+{
+    [TestFixture]
+    public class BindMethodFinderTests
+    {
+        private static readonly string ProtoUrl = "https://raw.githubusercontent.com/grpc/grpc-dotnet/edf4b478e5d35f19e69943eb807a99709fc8de3b/examples/Proto/greet.proto";
```
JunTaoLuo: I should remove the part where tests fetch content from the web, otherwise it's going to be a reliability nightmare.

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -63,11 +65,14 @@ internal CommandBase(IConsole console, Project project)
         internal IConsole Console { get; set; }
         internal Project Project { get; set; }
 
-        public void EnsureNugetPackages()
+        public void EnsureNugetPackages(Services services)
```
JunTaoLuo: Fix tests due to this change.

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,343 @@
...
+
+                var destinationDirectory = Path.GetDirectoryName(resolveDestination);
+                if (!Directory.Exists(destinationDirectory))
+                {
+                    Directory.CreateDirectory(destinationDirectory);
+                }
+            }
+            else
+            {
+                using (var stream = await GetStreamAsync(url))
```
JamesNK: Does this return a friendly error message if the URL is a 404, 500, etc

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,343 @@
...
+            if (contentNotModified)
+            {
+                Console.Log(CoreStrings.LogSkipDownload, destination, url);
+                return;
+            }
+
+            Console.Log(CoreStrings.LogDownload, destination, url);
+            if (!dryRun)
+            {
+                using (var stream = await GetStreamAsync(url))
```
JamesNK: Does this return a friendly error message if the URL is a 404, 500, etc

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,343 @@
...
+                    continue;
+                }
+
+                try
+                {
+                    expandedReferences.AddRange(
+                        Directory.GetFiles(Project.DirectoryPath, reference)
+                            .Select(r => r.Replace(Project.DirectoryPath + Path.DirectorySeparatorChar, string.Empty)));
+                }
+                catch (DirectoryNotFoundException) { }
```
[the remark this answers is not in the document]
JunTaoLuo: This was added to handle the cases where the given reference doesn't point to a valid directory. I suppose we could add a Directory.Exists check instead.

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,343 @@
...
+            return expandedReferences.ToArray();
+        }
+
+        public void RemoveProtobufReference(ProjectItem protobufRef, bool removeFile)
+        {
+            Project.RemoveItem(protobufRef);
+
+            if (removeFile)
+            {
+                File.Delete(Path.IsPathRooted(protobufRef.UnevaluatedInclude) ? protobufRef.UnevaluatedInclude : Path.Combine(Project.DirectoryPath, protobufRef.UnevaluatedInclude));
```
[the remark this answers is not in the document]
JunTaoLuo: I'm going to remove this step. Just remove the reference and don't do anything to the actual file.

src/dotnet-grpc/Commands/CommandBase.cs
JunTaoLuo: HttpRequestException, we should catch that.

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,343 @@
...
+        internal static readonly string ProtobufElement = "Protobuf";
+        internal static readonly string GrpcServicesElement = "GrpcServices";
+        internal static readonly string AccessElement = "Access";
+        internal static readonly string AdditionalImportDirsElement = "AdditionalImportDirs";
+        internal static readonly string SourceUrlElement = "SourceUrl";
+        internal static readonly string LinkElement = "Link";
+        internal static readonly string ProtosFolder = "Protos";
+
+        // Internal for testing
+        internal static Func<string, Task<Stream>> GetStreamAsync { get; set; } = HttpClient.GetStreamAsync;
```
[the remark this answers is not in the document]
JunTaoLuo: I'll try this out.

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -0,0 +1,343 @@
...
+
+            if (projectFiles.Length > 1)
+            {
+                throw new CLIToolException(string.Format(CultureInfo.CurrentCulture, CoreStrings.ErrorMoreThanOneProjectFound, currentDirectory));
+            }
+
+            return new Project(projectFiles[0]);
+        }
+
+        public IEnumerable<ProjectItem> ResolveReferences(string[] references)
```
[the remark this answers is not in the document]
JunTaoLuo: Meh... I don't think that's necessary if all we do is iterate it.

shirhatti:
If you call `dotnet grpc add-url -o <PATH_TO_DIRECTORY> <URL>` it fails (I tried on Windows). You need to specify a file path. Can we change that behavior?

shirhatti:
Seems odd to mix both xml attributes and child elements.

```xml
<Protobuf Include=".\Protos\error_details.proto" GrpcServices="None">
 
 <SourceUrl>https://raw.githubusercontent.com/googleapis/googleapis/master/google/rpc/error_details.proto</SourceUrl>
</Protobuf>
```

Was this intentional?

JunTaoLuo:
If only a directory is specified. What would we use for the file name? It's not guaranteed that the url contains the filename right? Also I did intentionally use child element for SourceUrl since it might be quite long, whereas I feel the other options are concise enough to be included as an attribute.

shirhatti:
Removing references seems clunky since there no way to enumerate them. You'd have to know the path to the proto file if you want to remove it

shirhatti:
Playing around with this, I don't think the default value for services should be `Both`. Since `Both` means you end up adding `Grpc.AspNetCore.Server' which brings in a transitive framework reference. You could end up in a position where can't even build the project.

If possible I'd only want to default to `Both` if the `Microsoft.AspNetCore.App` framework reference is already present. If not, we should default to `Client`.

shirhatti:
Why is additional import dirs an option with arity one that takes an semi-colon separated list? I'd rather see it as an option with `ArgumentArity.ZeroOrMore`

JunTaoLuo:
Two TODOs that will be addressed in a follow up:

- [ ] Use https://webpifeed.blob.core.windows.net/webpifeed/Partners/PackageVersions.json as a source of truth when adding package references.
- [ ] Add a command to enumerate the list of Protobuf references in the project.

src/dotnet-grpc/Options/CommonOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public static Option AccessOption() =>
+            new Option(
+                aliases: new[] { "--access" },
+                description: CoreStrings.AccessOptionDescription,
+                argument: new Argument<Access> { Name = "access" });
+
+        public static Option AdditionalImportDirsOption() =>
+            new Option(
+                aliases: new[] { "-a", "--additional-import-dirs" },
```
jtattermusch: IMHO the short name of the flag isn't self-descriptive enough.
protoc uses `-I` as short form of `--proto-path=` because that's a de-facto standard flag name for includes in the compiler world, while `-a` reminds me of `--all` at best. WDYT?

The long name of the flag is fine because "AdditionalImportDirs is already provided by Grpc.Tools

JunTaoLuo: I'll use -i (lowercase) since that seems more standard.

PR #274, opened 2019-05-29
Use HttpContext grpc-encoding header for compression interop tests

JamesNK:


JamesNK:
gRPC C# doesn't run compression interop tests, probably because these headers are filtered out of the metadata.

https://github.com/grpc/grpc/blob/837a99e1d49a892e6f2c46ee09a1b6b8405571c6/tools/run_tests/run_interop_tests.py#L145-L153

Issue #15, opened 2019-01-07
Configure MinRequestBodyDataRate when using GRPC


shirhatti:
We need to disable MinRequestBodyDataRate rate limit checks to support streaming RPCs. We should determine what's the most natural way to do this for the Kestrel Server when GRPC services are being used.

JunTaoLuo:
For the first preview, we'll ensure that templates guard against this #8

JunTaoLuo:
We'll need https://github.com/aspnet/AspNetCore/issues/7635 to do this properly.

JunTaoLuo:
Let's leave this open to at least be reminded that the stackalloc changes need to be reverted.

PR #275, opened 2019-05-29
Update runtime and use stackalloc again

JunTaoLuo:
Revert the stackalloc workaround now that we have a new Roslyn compiler with the latest SDK. 

Addresses the removal of the workaround in https://github.com/grpc/grpc-dotnet/issues/15.

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
@@ -42,7 +42,7 @@ namespace Grpc.AspNetCore.Server.Internal.CallHandlers
         {
             if (GrpcProtocolHelpers.IsInvalidContentType(httpContext, out var error))
             {
-                GrpcProtocolHelpers.SendHttpError(httpContext.Response, StatusCodes.Status415UnsupportedMediaType, StatusCode.Internal, error);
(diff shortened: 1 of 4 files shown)


src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
```diff
@@ -42,7 +42,7 @@ public Task HandleCallAsync(HttpContext httpContext)
         {
             if (GrpcProtocolHelpers.IsInvalidContentType(httpContext, out var error))
             {
-                GrpcProtocolHelpers.SendHttpError(httpContext.Response, StatusCodes.Status415UnsupportedMediaType, StatusCode.Internal, error);
+                GrpcProtocolHelpers.SendHttpError(httpContext.Response, StatusCodes.Status415UnsupportedMediaType, StatusCode.Internal, error!);
```
JunTaoLuo: Without this I get a nullable error here. Not sure why the compiler doesn't figure out that if the IsInvalidContentType returns true, error is not null in this branch. Did the NotNullWhenTrueAttribute approach stop working?

JunTaoLuo:
> /usr/bin/env: ‘bash\r’: No such file or directory
The command "./build/get-dotnet.sh" failed and exited with 127 during .

Looks like line ending issues...

PR #278, opened 2019-05-30
Update roslyn compiler for VS workloads

JunTaoLuo:
Turns out my VS installation had some custom modifications to use the latest Roslyn compiler. This change should ensure VS uses an updated compiler and can therefore build and test.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -16,4 +16,8 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="$(MicrosoftNetCompilersToolsetPackageVersion)" PrivateAssets="All" />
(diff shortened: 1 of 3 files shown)


Directory.Build.props
```diff
@@ -16,4 +16,8 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="$(MicrosoftNetCompilersToolsetPackageVersion)" PrivateAssets="All" />
```
JamesNK: How long is this needed for? Add a TODO for removing it.

JunTaoLuo: Probably until the next VS release.

JamesNK: Ok. Add a todo and :shipit:

This is semi blocking me (and anyone else using VS) at the moment.

PR #281, opened 2019-06-02
React to new .NET Core SDK

JamesNK:
Lots of compile errors to fix because of https://github.com/aspnet/AspNetCore/pull/10154

A functional test is broken because of a regression in TestServer. Test is ignored until TestServer is fixed.

### DIFF
--- a/test/FunctionalTests/UnaryMethodTests.cs
+++ b/test/FunctionalTests/UnaryMethodTests.cs
@@ -337,10 +337,9 @@ namespace Grpc.AspNetCore.FunctionalTests
         }
 
         [Test]
-        [Ignore("Regression in TestServer - https://github.com/aspnet/AspNetCore/issues/10737")]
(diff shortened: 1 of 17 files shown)


test/FunctionalTests/UnaryMethodTests.cs
```diff
@@ -337,6 +337,7 @@ static Task<Empty> ReturnContextInfoInTrailers(Empty request, ServerCallContext
         }
 
         [Test]
+        [Ignore("Regression in TestServer - https://github.com/aspnet/AspNetCore/issues/10737")]
```
JunTaoLuo: Seems like the issue has been resolved so disposing the request body no longer closes the stream.

JamesNK: Fixing it by creating a new stream. This is more correct in line with how HttpClient+StreamContent is used.

Issue #276, opened 2019-05-29
Client API feedback


JunTaoLuo:
naming:
- [x] Grpc.NET.Client vs Grpc.Dotnet.Client ?
- [x] ClientFactory 

packaging:
- [x] Consider splitting into more packages (server and client ClientFactory)
- [ ] ~Use metapackage to reference all server, client and integration packages~ https://github.com/grpc/grpc-dotnet/issues/292

apis:
- [x] Builder vs `Action<Options>` on creation to configure options and HttpClient options.

samples:
- [x] using the new APIs factory/GrpcClient.Create
- [x] nested calls

- [ ] ~Schedule API review~ https://github.com/grpc/grpc-dotnet/issues/335

PR #282, opened 2019-06-02
Split non-server gRPC client factory into separate package

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/276


JamesNK:
Also has https://github.com/grpc/grpc-dotnet/pull/281

JamesNK:
Also adds AddInterceptor methods to client factory.

src/Grpc.Net.ClientFactory/Internal/GrpcHttpClientFactory.cs
```diff
@@ -73,34 +74,22 @@ public TClient CreateClient(HttpClient httpClient, string name)
...
+            // Interceptors wrap each other and execute in reverse order
+            for (var i = options.Interceptors.Count - 1; i >= 0; i--)
             {
-                if (serverCallContext == null)
-                {
-                    throw new InvalidOperationException("Cannot propagate the call deadline to the client. Cannot find the current gRPC ServerCallContext.");
-                }
-
-                callInvoker.Deadline = serverCallContext.Deadline;
+                resolvedCallInvoker = resolvedCallInvoker.Intercept(options.Interceptors[i]);
```
JunTaoLuo: Were we okay with this? I recall we were unsure if all the functionalities here can be reproduced with just the intercept extension call. Something about modifying hosts and that it would require using the delegating handlers of HttpClient?

JamesNK: You can add delegating handlers to the HttpClient alongside interceptors using the client factory.

Issue #237, opened 2019-05-07
Remove None itemgroup workaround when VS 2019 (16.2) is released


JunTaoLuo:

JunTaoLuo:
Revert the changes in https://github.com/grpc/grpc-dotnet/pull/231

PR #285, opened 2019-06-03
Remove workarounds for DesignTimeBuild

JunTaoLuo:
We can remove these workarounds now that VS 16.2 Preview1 is released.

Addresses: https://github.com/grpc/grpc-dotnet/issues/237


src/Grpc.AspNetCore.Server/build/netcoreapp3.0/Grpc.AspNetCore.Server.targets
```diff
@@ -1,9 +0,0 @@
-<Project>
```
JunTaoLuo: We could keep this until 16.2 is GA so users using 16.1 will have DesignTimeBuilds by default.

JunTaoLuo:
I think I'll hold off on this change until we have a 0.1.21 release since users of VS 16.1 need the workaround for DesignTimeBuild.

PR #287, opened 2019-06-03
Rename Grpc.NetCore.HttpClient to Grpc.Net.Client

JamesNK:
Issues with current Grpc.NetCore.HttpClient name:
* Package will eventually target netstandard2.1 (not yet available) so it will be more than just "NetCore".
* HttpClient in the namespace collides with HttpClient usages in code. Client avoids this.

Renaming to Grpc.Net.Client fixes these issues.


jtattermusch:
LGTM.

Issue #160, opened 2019-03-13
timeout_on_sleeping_server interop test is flaky


jtattermusch:
ruby client
https://source.cloud.google.com/results/invocations/1a3e71ed-19e3-4161-8a98-4d01e625289f/targets/github%2Fgrpc%2Finterop_test/tests
```
cloud_to_cloud:ruby:aspnetcore_server:timeout_on_sleeping_server:tls

/var/local/git/grpc/src/ruby/lib/grpc/generic/active_call.rb:31:in `check_status': 13:Received RST_STREAM with error code 2 (GRPC::Internal)
	from /var/local/git/grpc/src/ruby/lib/grpc/generic/bidi_call.rb:209:in `block in read_loop'
	from /var/local/git/grpc/src/ruby/lib/grpc/generic/bidi_call.rb:195:in `loop'
	from /var/local/git/grpc/src/ruby/lib/grpc/generic/bidi_call.rb:195:in `read_loop'
	from src/ruby/pb/test/client.rb:461:in `each'
	from src/ruby/pb/test/client.rb:461:in `timeout_on_sleeping_server'
	from src/ruby/pb/test/client.rb:763:in `call'
	from src/ruby/pb/test/client.rb:763:in `main'
	from src/ruby/pb/test/client.rb:768:in `<main>'
I, [2019-03-13T09:31:34.366230 #77]  INFO -- : ... connecting securely to localhost:32771
W, [2019-03-13T09:31:35.367384 #77]  WARN -- : bidi: read-loop failed
W, [2019-03-13T09:31:35.367455 #77]  WARN -- : 13:Received RST_STREAM with error code 2 (GRPC::Internal)
/var/local/git/grpc/src/ruby/lib/grpc/generic/active_call.rb:31:in `check_status'
/var/local/git/grpc/src/ruby/lib/grpc/generic/bidi_call.rb:209:in `block in read_loop'
/var/local/git/grpc/src/ruby/lib/grpc/generic/bidi_call.rb:195:in `loop'
/var/local/git/grpc/src/ruby/lib/grpc/generic/bidi_call.rb:195:in `read_loop'
src/ruby/pb/test/client.rb:461:in `each'
src/ruby/pb/test/client.rb:461:in `timeout_on_sleeping_server'
src/ruby/pb/test/client.rb:763:in `call'
src/ruby/pb/test/client.rb:763:in `main'
src/ruby/pb/test/client.rb:768:in `<main>'
```

And python client:
[...]

PR #299, opened 2019-06-06
 Ensure deadline timer doesn't abort after call is complete

JamesNK:
Addresses https://github.com/aspnet/AspNetCore/issues/9239

PR #301, opened 2019-06-06
Refactor deadline to apply more gracefully

JamesNK:
This should fix flaky test - https://github.com/grpc/grpc-dotnet/issues/160

This brings grpc-dotnet closer to what Grpc.Core does when a deadline is exceeded: returns trailers with status of DeadlineExceeded, and RST_STREAM (NO_ERROR)

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
@@ -16,6 +16,8 @@
 
 #endregion
 
+using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Grpc.AspNetCore.Server.Features;
 using Grpc.Core;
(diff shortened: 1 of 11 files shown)


JamesNK:
And will replace https://github.com/grpc/grpc-dotnet/pull/299

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
```diff
@@ -46,10 +48,115 @@ public Task HandleCallAsync(HttpContext httpContext)
...
+                // Or the deadline has been exceeded and we want to trigger ServerCallContext.CancellationToken
+                cts.Cancel();
+
+                if (completedTask != handleCallTask)
+                {
+                    serverCallContext.DeadlineExceeded();
+                    httpContext.Response.ConsolidateTrailers(serverCallContext);
+
+                    // Ensure any errors thrown by the dead call are handled
+                    _ = ObserveDeadCall(serverCallContext, handleCallTask);
```
davidfowl: This looks like a bad time. What's the expectation when user code hangs and you tear everything down? Just let it hang...

JamesNK: Is it worse or better than HttpContext.Abort? Is there a difference between:
1. HttpContext.Abort being called, the server returning a response, and the request delegate hanging
2. The deadline being hit, the request delegate sending deadline response and exiting, and the call logic hanging in its own task

Current behavior:

> In gRPC there is the concept of a deadline. The deadline is sent by the client to the server with a gRPC call in a header. It is the client saying to the server "you have X seconds to finish this call, after-which give up".
>
> * If the server finished the call in less than X sends then the response completes as usual.
> * If the server exceeds the X seconds then it immediately finishes the HTTP/2 response, sending a RST_STREAM. The user's gRPC method code will likely still be executing after this. If they are good then it will be using a cancellation token, and they stop whatever they're doing quickly. But they might not, and continue doing stuff after the response is complete. This is fine. It is also fine if an error eventually gets thrown because they tried to do an HTTP thing.
>
> Today the gRPC server is finishing the connection when the deadline is hit by calling `HttpContext.Abort()`. This immediately sends RST_STREAM with an INTERNAL_ERROR error code. The request delegate with the user's code is still running. Using `Abort` is not ideal, but its response seems to be acceptable to the various gRPC clients of the world.

JamesNK:
This is slower by the way.

Before:

```cs
------------------------ |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 DeadlineHandleCallAsync | 2.877 us | 0.0131 us | 0.0123 us | 347,638.3 |      0.1221 |           - |           - |               520 B |
```

After:

```cs
------------------------ |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 DeadlineHandleCallAsync | 3.548 us | 0.0095 us | 0.0085 us | 281,833.5 |      0.2670 |           - |           - |             1.09 KB |
```

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
halter73: I don't think swallowing and logging the errors is enough considering our plans for more HTTP/2 pooling in the future. If we ever want to support pooling HTTP/2 streams without the extra allocations required for poisoning the old streams, we'll want to ensure any calls that could be referencing the HttpContext complete before the middleware does. We don't want our servers to think they can pool the HttpContext until all calls associated with the request complete.

JamesNK:
Will probably go with a different solution to this. I'm going to close for now.

PR #304, opened 2019-06-08
Change samples to use new gRPC client

JamesNK:
@JunTaoLuo you have a macbook right? Could you test that the samples continue to run on MacOS.


JunTaoLuo:
Yup it works! Gonna merge now so I can make a PR on top of this one.

Issue #251, opened 2019-05-15
Issues with client certificate


lean-bristleback:
Two issues:
1. When running an example project (/examples/Clients/Greeter/) on Windows and using a client certificate for authentication, the context.GetHttpContext().Connection.ClientCertificate property is still null. 
2. When setting HttpsConnectionAdapterOptions.ClientCertificateMode to ClientCertificateMode.RequireCertificate, the client fails to establish connection ("Error: HTTP/2 over TLS was not negotiated on an HTTP/2-only endpoint").

JamesNK:
In preview 6 there is a new gRPC client that uses HttpClient internally.

I have added a example of sending a client certificate with gRPC calls and using them on the server. It is available here - https://github.com/grpc/grpc-dotnet/pull/306

Issue #188, opened 2019-03-27
Feature request: Improve gRPC client Certificate handling


damienbod:
It would help, if it was possible to use a pfx file with password, or a thumbprint from the client. This would make it easy to setup HTTPS with the client for different deployment types.

something like this:

example 1:
```
cert = new X509Certificate2(Path.Combine(_environment.ContentRootPath, "sts_dev_cert.pfx"), "1234");
```

example 2:
```
using (X509Store store = new X509Store(StoreName.My, StoreLocation.LocalMachine))
{
  store.Open(OpenFlags.ReadOnly);
  var certs = store.Certificates.Find(X509FindType.FindByThumbprint, certificateThumbprint, false);
  cert = certs[0];
  store.Close();
}
```

example 3:
```
var vaultConfigSection = Configuration.GetSection("Vault");
var keyVaultService = new KeyVaultCertificateService(vaultConfigSection["Url"], vaultConfigSection["ClientId"], vaultConfigSection["ClientSecret"]);
cert = keyVaultService.GetCertificateFromKeyVault(vaultConfigSection["CertificateName"]);
```

vincentparrett:
+100 for this. The current SSLCredentials setup is not great. I'm deciding whether to use grpc or halibut (to replace wcf) at the moment and halibuts uses x509 certs with thumbprints and it's a lot easier to manage.

JunTaoLuo:
This is a good ask, we'll investigate further in the next release.

JamesNK:
From preview 6 you will be able to use a new client that uses HttpClient to make calls internally. You can set the client certificate on the HttpClient like you do today.

JamesNK:
In preview 6 there is a new gRPC client that uses HttpClient internally.

I have added a example of sending a client certificate with gRPC calls and using them on the server. It is available here - https://github.com/grpc/grpc-dotnet/pull/306

wazzamatazz:
The new HttpClient-based client is great, but it's only an option if you're using .NET Core 3.0 on the client side. Improving certificate handling for the Grpc.Core-based client as described above would be a great help when working with a target framework that doesn't support .NET Standard 2.1.

JamesNK:
Raise Grpc.Core related improvements at https://github.com/grpc/grpc

PR #306, opened 2019-06-09
Add example of using client certificate with new client

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/188, https://github.com/grpc/grpc-dotnet/issues/251


testassets/InteropTestsWebsite/Program.cs
```diff
@@ -49,7 +49,7 @@ public static void Main(string[] args)
                         if (useTls)
                         {
                             var basePath = Path.GetDirectoryName(typeof(Program).Assembly.Location);
-                            var certPath = Path.Combine(basePath, "Certs/server1.pfx");
+                            var certPath = Path.Combine(basePath, @"Certs\server1.pfx");
```
JunTaoLuo: Split the string and let Path.Combine choose the separator? Occurs in the client sample as well.

JamesNK:
Will update this example to use new client certificate when preview 6 is published.

jtattermusch:
LGTM.

Issue #273, opened 2019-05-29
Change client target to netstandard2.1


JamesNK:
https://github.com/grpc/grpc-dotnet/blob/55a559499197c0e6f4934b7b89139d3365a8c09c/src/Grpc.NetCore.HttpClient/Grpc.NetCore.HttpClient.csproj#L16

Requires https://github.com/dotnet/standard/issues/1144

JamesNK:
Plus client factory.

PR #315, opened 2019-06-11
Change client to target netstandard2.1

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/273

There is a bug somewhere in the compiler/SDK that required a workaround.

https://github.com/dotnet/cli/issues/11378#issuecomment-499988244

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "3.0.100-preview6-012250"
+    "version": "3.0.100-preview7-012341"
   }
 }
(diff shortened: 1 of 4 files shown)


global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "3.0.100-preview6-012250"
+    "version": "3.0.100-preview7-012341"
```
JunTaoLuo: so close to 012345

PR #322, opened 2019-06-13
 Add certificate authentication to example

JamesNK:


JamesNK:
@blowdart @HaoK The new certificate auth works well. Here is an example of using it with gRPC in case you're looking for demos 😄

Issue #331, opened 2019-06-14
IServiceMethodProvider - make boolean to be less shouty


mgravell:
Currently, the default `IServiceMethodProvider<TService>` (`BinderServiceMethodProvider<TService>`) shouts when it doesn't like a type:

``` c#
Log.BindMethodNotFound(_logger, typeof(TService));
```

However, now that we can add providers, this means that the *default* provider complains even if a *later* provider succeeds.

Suggestion:

1. make `OnServiceMethodDiscovery` return `bool` to indicate whether it has "claimed" a type
2. instead of logging a failure, `BinderServiceMethodProvider<TService>` should just `return false`
3. the loop in `ServiceRouteBuilder<TService>.Build` could `break;` when the first item returns `true` (i.e. has claimed the type)
4. if nobody claims it, **then** log (from `Build`)

or something similar...?

JamesNK:
I'd like to leave the door open to multiple providers returning services for a type.

Some options:
* The message level could be reduced to info or debug. A warning could be raised if no methods are resolved after all providers have run.
* You could remove the default provider from DI

JamesNK:
I like option 1.

JamesNK:
https://github.com/grpc/grpc-dotnet/pull/333

PR #333, opened 2019-06-16
Improve method discovery logging

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/331

Also fixes a bunch of nullable warnings. These are showing up because BCL is adding nullable metadata.


JamesNK:
@JunTaoLuo 🙏

jtattermusch:
LGTM.

PR #334, opened 2019-06-17
Send status on deadline with CompleteAsync

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/160

// @Tratcher @halter73

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,7 +6,7 @@
   <PropertyGroup>
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)keys\Grpc.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
 
     <!-- Don't make missing XML docs a fatal build error, but still surface so we have visibility into undocumented APIs. -->
     <WarningsNotAsErrors>$(WarningsNotAsErrors);CS1591</WarningsNotAsErrors>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,7 +6,7 @@
   <PropertyGroup>
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)keys\Grpc.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
-    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
 
     <!-- Don't make missing XML docs a fatal build error, but still surface so we have visibility into undocumented APIs. -->
     <WarningsNotAsErrors>$(WarningsNotAsErrors);CS1591</WarningsNotAsErrors>
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
@@ -25,11 +25,12 @@ using System.Threading.Tasks;
 using Grpc.AspNetCore.Server.Features;
 using Grpc.Core;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.Extensions.Logging;
 
 namespace Grpc.AspNetCore.Server.Internal
 {
-    internal sealed partial class HttpContextServerCallContext : ServerCallContext, IDisposable, IServerCallContextFeature
+    internal sealed partial class HttpContextServerCallContext : ServerCallContext, IAsyncDisposable, IServerCallContextFeature
     {
         private static readonly AuthContext UnauthenticatedContext = new AuthContext(null, new Dictionary<string, List<AuthProperty>>());
         private readonly ILogger _logger;
@@ -41,9 +42,11 @@ namespace Grpc.AspNetCore.Server.Internal
         private Metadata? _requestHeaders;
         private Metadata? _responseTrailers;
         private DateTime _deadline;
-        private Timer? _deadlineTimer;
+        private CancellationTokenSource? _deadlineCts;
+        private SemaphoreSlim? _deadlineLock;
         private Status _status;
         private AuthContext? _authContext;
+        private CancellationTokenRegistration _requestAbortedRegistration;
         // Internal for tests
         internal bool _disposed;
 
@@ -143,7 +146,7 @@ namespace Grpc.AspNetCore.Server.Internal
             HttpContext.Response.ConsolidateTrailers(this);
         }
 
-        protected override CancellationToken CancellationTokenCore => HttpContext.RequestAborted;
+        protected override CancellationToken CancellationTokenCore => _deadlineCts?.Token ?? HttpContext.RequestAborted;
 
         protected override Metadata ResponseTrailersCore
         {
@@ -166,16 +169,46 @@ namespace Grpc.AspNetCore.Server.Internal
 
         internal Task EndCallAsync()
         {
-            HttpContext.Response.ConsolidateTrailers(this);
-
-            if (HasBufferedMessage)
+            if (_deadlineLock == null)
             {
-                // Flush any buffered content
-                return HttpContext.Response.BodyWriter.FlushAsync().GetAsTask();
+                return EndCallAsyncCore();
             }
-            else
+
+            return DeadlineEndCallAsync();
+
+            async Task DeadlineEndCallAsync()
             {
-                return Task.CompletedTask;
+                await _deadlineLock!.WaitAsync();
+                try
+                {
+                    // Have to await EndCallAsyncCore inside the try/finally
+                    await EndCallAsyncCore();
+                }
+                finally
+                {
+                    _deadlineLock.Release();
+                }
+            }
+
+            Task EndCallAsyncCore()
+            {
+                // Don't set trailers if deadline exceeded or request aborted
+                if (CancellationToken.IsCancellationRequested)
+                {
+                    return Task.CompletedTask;
+                }
+
+                HttpContext.Response.ConsolidateTrailers(this);
+
+                if (HasBufferedMessage)
+                {
+                    // Flush any buffered content
+                    return HttpContext.Response.BodyWriter.FlushAsync().GetAsTask();
+                }
+                else
+                {
+                    return Task.CompletedTask;
+                }
             }
         }
 
@@ -249,11 +282,13 @@ namespace Grpc.AspNetCore.Server.Internal
             {
                 _deadline = Clock.UtcNow.Add(timeout);
 
-                // Set timer field first and then set the timeout.
-                // The method uses the timer to lock and we want to avoid the timer immediately
-                // running before the field is set.
-                _deadlineTimer = new Timer(DeadlineExceeded);
-                _deadlineTimer.Change(timeout, Timeout.InfiniteTimeSpan);
+                // Create lock before setting up deadline event
+                _deadlineLock = new SemaphoreSlim(1, 1);
+
+                _deadlineCts = new CancellationTokenSource(timeout);
+                _deadlineCts.Token.Register(DeadlineExceeded);
+
+                _requestAbortedRegistration = HttpContext.RequestAborted.Register(RequestAborted);
             }
             else
             {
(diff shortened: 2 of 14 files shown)


Directory.Build.props
```diff
@@ -6,7 +6,7 @@
   <PropertyGroup>
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)keys\Grpc.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
JamesNK: Temporarily disabled. Nullable warnings fixes in https://github.com/grpc/grpc-dotnet/pull/333

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -114,36 +117,89 @@ protected override Metadata RequestHeadersCore
...
+                else
+                {
+                    Log.ErrorExecutingServiceMethod(_logger, method, ex);
+
+                    var message = ErrorMessageHelper.BuildErrorMessage("Exception was thrown by handler.", ex, ServiceOptions.EnableDetailedErrors);
+                    _status = new Status(StatusCode.Unknown, message);
+                }
+
+                // Don't update trailers if request has exceeded deadline/aborted
+                if (!CancellationToken.IsCancellationRequested)
```
Tratcher: This is not sufficient thread safety, you'd need a lock here.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -249,11 +346,13 @@ public void Initialize()
...
-                // running before the field is set.
-                _deadlineTimer = new Timer(DeadlineExceeded);
-                _deadlineTimer.Change(timeout, Timeout.InfiniteTimeSpan);
+                // Create lock before setting up deadline event
+                _deadlineLock = new SemaphoreSlim(1, 1);
+
+                _deadlineCts = new CancellationTokenSource(timeout);
+                _deadlineCts.Token.Register(DeadlineExceeded);
+
+                _requestAbortedRegistration = HttpContext.RequestAborted.Register(RequestAborted);
```
Tratcher: Rather than register against RequestAborted here, I'd expected you to call CTS.CreateLinkedTokenSource to combine _deadlineCts and RequestAborted and then replace HttpContext.RequestAborted with the combine token. That way the app would be notified if the deadline expired.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -275,6 +374,11 @@ public void Initialize()
             HttpContext.Response.Headers.Append(GrpcProtocolConstants.MessageEncodingHeader, ResponseGrpcEncoding);
         }
 
+        private void RequestAborted()
+        {
+            _deadlineCts?.Cancel();
```
Tratcher: Not much point in calling Cancel on the deadline cts if the request was already aborted. As long as Dispose gets called eventually.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -294,56 +398,70 @@ private TimeSpan GetTimeout()
...
+                if (_disposed)
+                {
+                    return;
+                }
+
+                Log.DeadlineExceeded(_logger, GetTimeout());
+
+                var status = new Status(StatusCode.DeadlineExceeded, "Deadline Exceeded");
+
+                var trailersDestination = GrpcProtocolHelpers.GetTrailersDestination(HttpContext.Response);
```
Tratcher: GetStatusDestination?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -166,17 +222,58 @@ protected override Status StatusCore
...
+            // Don't set trailers if deadline exceeded or request aborted
+            if (CancellationToken.IsCancellationRequested)
+            {
+                return;
+            }
+
+            HttpContext.Response.ConsolidateTrailers(this);
+
+            // Clean up deadline
+            Dispose();
```
halter73: Does it still make sense for HttpContextServerCallContext to implement `IDisposable` if the user of is expected to call either EndCallAsync or ProcessHandlerError instead?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
halter73: What race are you worried about? It looks like we should have the _deadlineLock here in ProcessHandlerError and in DeadlineExceeded where HttpContext.Abort() is called.

Tratcher: Ah, nevermind if this is always called within the _deadlineLock.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: What do you mean by the app? ServerCallContext exposes its own token which is what users should use in their gRPC services

https://github.com/grpc/grpc-dotnet/blob/f721fbef75e485b252588002da4c7b919666f9c8/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs#L202

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: It is canceled on abort so the token on ServerCallContext is raised both for aborts and deadlines

https://github.com/grpc/grpc-dotnet/blob/f721fbef75e485b252588002da4c7b919666f9c8/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs#L202

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: Hmm, probably not. Also now that I think about it instances of this type are passed to the user. Someone could be completely crazy and cast to IDisposable and dispose it.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
halter73: It should be possible to use CreateLinkedTokenSource only when there are at least two cancellable source tokens and otherwise save the allocations, but it probably ends up not looking any better than what you already have.

E.g. https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs#L1502-L1520

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -294,56 +398,70 @@ private TimeSpan GetTimeout()
...
-                    Monitor.Exit(_deadlineTimer);
-                }
+                _deadlineLock.Release();
             }
         }
 
+        public void Dispose()
+        {
+            _disposed = true;
+            _deadlineCts?.Dispose();
```
halter73: This is always called with _deadlineLock, right?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
Tratcher: CancellationTokenCore gives them the result of one token or the other, not both. E.g. if you're using deadlines then the app doesn't know about RequestAborted. That should be a linked source so they can can be triggered by either.

JamesNK: > CancellationTokenCore gives them the result of one token or the other, not both. E.g. if you're using deadlines then the app doesn't know about RequestAborted. That should be a linked source so they can can be triggered by either.

If there is no deadline then only HttpContext.RequestAborted can raise the event.

If there is a deadline then _deadlineCts gets raised when HttpContext.RequestAborted is raised.

I'll investigate replacing the registration with a linked cancellation token but I suspect it will allocate more and not improve code readability much.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: Yes

JamesNK: Hmmm, _deadlineLock isn't being disposed. Do you need to dispose SemaphoreSlim?

JamesNK:
https://github.com/JamesNK/grpc-dotnet/commit/e8ae3f7e9a879298f83544282553eebc26e65477

Before:
```
------------------------ |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 DeadlineHandleCallAsync | 3.006 us | 0.0172 us | 0.0161 us | 332,665.5 |      0.2327 |           - |           - |               976 B |
```

After:
```
------------------------ |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 DeadlineHandleCallAsync | 3.191 us | 0.0110 us | 0.0097 us | 313,425.8 |      0.2518 |           - |           - |             1.04 KB |
```

It's slower. I don't see a benefit over what it was previously.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: It is used for some other trailers. It is odd in this context but it makes sense elsewhere.

JamesNK:
Refactored for disposing of SemaphoreSlim. Context is using Timer again. Use Timer.DisposeAsync after the lock has been exited to ensure the deadline is no longer running, then dispose the lock.

JamesNK:
@Tratcher @halter73 @davidfowl @JunTaoLuo Review please 🙏

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -294,56 +372,77 @@ private TimeSpan GetTimeout()
...
 
-        public void Dispose()
-        {
-            var lockTaken = false;
             try
             {
-                // Only take a lock if there is a deadline
-                if (_deadlineTimer != null)
+                // Re-check after lock
+                if (_deadlineDisposed)
```
Tratcher: But you just acquired the _deadlineLock, how could it be disposed?
Does _deadlineLock.WaitAsync throw if disposed?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
halter73: >  Note
> 
> Always call Dispose before you release your last reference to the SemaphoreSlim. Otherwise, the resources it is using will not be freed until the garbage collector calls the SemaphoreSlim object's Finalize method.

https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim.dispose?view=netcore-3.0

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
halter73: Do you have to release the SemaphoreSlim before disposing it?

I assume calls to SemaphoreSlim.WaitAsync() after disposal would throw an ODE, but do the ongoing calls awaiting SemaphoreSlim.WaitAsync() also see an ODE when it gets disposed?

I wonder if we need to add a Monitor now in order to avoid calling to SemaphoreSlim.WaitAsync() after we've disposed it.

JamesNK: > But you just acquired the _deadlineLock, how could it be disposed?

Yeah, this line should be removed. It will never do anything. It should be impossible to be here after dispose because the timer is disposed of first with DisposeAsync.

JamesNK: > I wonder if we need to add a Monitor now in order to avoid calling to SemaphoreSlim.WaitAsync() after we've disposed it.

I think I've already handled that situation. WaitAsync should never be called here because the timer is disposed of first with DisposeAsync.

JamesNK: https://github.com/grpc/grpc-dotnet/blob/55d1d3f370b230c3bd38b05777d040b0eafa3c80/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs#L425-L444

JamesNK:
Replaced dispose flag with callComplete flag. Added additional comments to make it clearer what is happening

PR #339, opened 2019-06-19
Allow mix background worker, grpc service and asp.net core application

pengweiqhca:


JamesNK:
I don't understand the reason behind this PR. If you're not running in an ASP.NET Core request then you shouldn't enable propagation with `EnableCallContextPropagation()`. Silently not propagating is not a good solution.

If you want to call the same client in the same app and only sometimes have propagation then you can register it multiple times with different names:

```cs
services.AddGrpcClient<GreeterClient>("Greeter", o => o.BaseAddress = url);

services.AddGrpcClient<GreeterClient>("GreeterWithPropagation", o => o.BaseAddress = url)
    .EnableCallContextPropagation();
```

```cs
public MyService(GrpcClientFactory clientFactory)
{
    _greeterClient = clientFactory.CreateClient<GreeterClient>("Greeter");
    _greeterClientWithPropagation = clientFactory.CreateClient<GreeterClient>("GreeterWithPropagation");
}
```

pengweiqhca:
In my code, I just want to use`GreeterClient` , not care my code to running in `background worker` or `grpc service`.
If use `different names`, I need to know where is my code to running.
`EnableCallContextPropagation()` should be as far as possible to propagate CallContext, rather than having to propagate CallContext.

JamesNK:
Your situation sounds pretty unusual, and it's not how propagation is indented to work.

Is there anything stopping you from having a copy of `EnableCallContextPropagation` in your own application. Everything it uses is public. You can modify it to suit your own needs.

Issue #318, opened 2019-06-11
Tracing


JamesNK:
Investigate tracing with gRPC, particularly OpenCensus.

JamesNK:
SO!

Talking with @davidfowl and the tracing groundwork has been done for us in ASP.NET Core and HttpClient. By using those we get tracing for free.

I experimented hooking up OpenTelemetrySDK - https://github.com/open-telemetry/opentelemetry-dotnet - and exporting to Zipkin and got promising results:

![image](https://user-images.githubusercontent.com/303201/59488134-cae82a80-8ed2-11e9-9e0e-f60685941a6d.png)

![image](https://user-images.githubusercontent.com/303201/59488148-d6d3ec80-8ed2-11e9-9969-83b5ba696b73.png)

Source: https://github.com/JamesNK/grpc-dotnet/commit/9831bfb2d7a9b56be3c2c9e4361140b75790e926

JamesNK:
I think perhaps the only remaining work here is potentially adding gRPC specific metadata to these events. For example, gRPC requests will always return 200 OK. Can we include grpc-status in these results?

JamesNK:
I don't believe there are any product changes required here.

In the future we could add docs/examples for tracing.

davidfowl:
If it is warranted, we could add more diagnostic listener events for GRPC invocations that may contain more information (like service name and method name and grpc status for the response).

davidfowl:
I think there's some value in doing this for both GRPC and Signal

Issue #283, opened 2019-06-03
Add performance counters


JamesNK:

JunTaoLuo:
Details: https://github.com/aspnet/AspNetCore/issues/9486

JunTaoLuo:
- Errors

JunTaoLuo:
ping @shirhatti.

JamesNK:
* Status
  * Errors
  * Deadlines exceeded
  * Operations cancelled
* MessagesReceived
* MessagesSent

PR #341, opened 2019-06-19
Add EventSource and activity tags to server

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/318, https://github.com/grpc/grpc-dotnet/issues/283

~Built on top of https://github.com/grpc/grpc-dotnet/pull/334~

~Changes in https://github.com/grpc/grpc-dotnet/commit/c986480bb06c969f267f9c673703a253dde6bc80~

### DIFF
--- a/examples/Server/Startup.cs
+++ b/examples/Server/Startup.cs
@@ -20,10 +20,8 @@ using System;
 using System.IdentityModel.Tokens.Jwt;
 using System.Security.Claims;
 using System.Security.Cryptography.X509Certificates;
-using System.Threading.Tasks;
 using Count;
 using Greet;
-using Microsoft.AspNetCore.Authentication.Certificate;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
@@ -44,14 +42,14 @@ namespace GRPCServer
                 {
                     // This registers a global interceptor with a Singleton lifetime. The interceptor must be added to the service collection in addition to being registered here.
                     options.Interceptors.Add<MaxConcurrentCallsInterceptor>();
-                    // This registers a global interceptor with a Scoped lifetime.
-                    options.Interceptors.Add<MaxStreamingRequestTimeoutInterceptor>(TimeSpan.FromSeconds(30));
                 })
                 .AddServiceOptions<GreeterService>(options =>
                 {
                     // This registers an interceptor for the Greeter service with a Singleton lifetime.
                     // NOTE: Not all calls should be cached. Since the response of this service only depends on the request and no other state, adding caching here is acceptable.
                     options.Interceptors.Add<UnaryCachingInterceptor>();
+                    // This registers an interceptor for the Greeter service with a Scoped lifetime.
+                    options.Interceptors.Add<MaxStreamingRequestTimeoutInterceptor>(TimeSpan.FromSeconds(30));
                 });
             services.AddGrpcReflection();
             services.AddSingleton(new MaxConcurrentCallsInterceptor(200));
--- a/src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
@@ -1,0 +1,159 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System.Diagnostics.Tracing;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using Grpc.Core;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal class GrpcEventSource : EventSource
+    {
+        public static readonly GrpcEventSource Log = new GrpcEventSource();
+
+        private PollingCounter? _totalCallsCounter;
+        private PollingCounter? _currentCallsCounter;
+        private PollingCounter? _messagesSentCounter;
+        private PollingCounter? _messagesReceivedCounter;
+        private PollingCounter? _callsFailedCounter;
+        private PollingCounter? _callsDeadlineExceededCounter;
+        private PollingCounter? _callsUnimplementedCounter;
+
+        private long _totalCalls;
+        private long _currentCalls;
+        private long _messageSent;
+        private long _messageReceived;
+        private long _callsFailed;
+        private long _callsDeadlineExceeded;
+        private long _callsUnimplemented;
+
+        internal GrpcEventSource()
+            : base("Grpc.AspNetCore.Server")
+        {
+        }
+
+        // Used for testing
+        internal GrpcEventSource(string eventSourceName)
+            : base(eventSourceName)
+        {
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [Event(eventId: 1, Level = EventLevel.Verbose)]
+        public void CallStart(string method)
+        {
+            Interlocked.Increment(ref _totalCalls);
+            Interlocked.Increment(ref _currentCalls);
+
+            WriteEvent(1, method);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [Event(eventId: 2, Level = EventLevel.Verbose)]
+        public void CallStop()
+        {
+            Interlocked.Decrement(ref _currentCalls);
+
+            WriteEvent(2);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [Event(eventId: 3, Level = EventLevel.Error)]
+        public void CallFailed(StatusCode statusCode)
+        {
+            Interlocked.Increment(ref _callsFailed);
+
+            WriteEvent(3, (int)statusCode);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [Event(eventId: 4, Level = EventLevel.Error)]
+        public void CallDeadlineExceeded()
+        {
+            Interlocked.Increment(ref _callsDeadlineExceeded);
+
+            WriteEvent(4);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [Event(eventId: 5, Level = EventLevel.Verbose)]
+        public void MessageSent()
+        {
+            Interlocked.Increment(ref _messageSent);
+
+            WriteEvent(5);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [Event(eventId: 6, Level = EventLevel.Verbose)]
+        public void MessageReceived()
+        {
+            Interlocked.Increment(ref _messageReceived);
+
+            WriteEvent(6);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [Event(eventId: 7, Level = EventLevel.Verbose)]
+        public void CallUnimplemented(string method)
+        {
+            Interlocked.Increment(ref _callsUnimplemented);
+
+            WriteEvent(7, method);
+        }
+
+        protected override void OnEventCommand(EventCommandEventArgs command)
+        {
+            if (command.Command == EventCommand.Enable)
+            {
+                // This is the convention for initializing counters in the RuntimeEventSource (lazily on the first enable command).
+                // They aren't disabled afterwards...
+
+                _totalCallsCounter ??= new PollingCounter("total-calls", this, () => _totalCalls)
+                {
+                    DisplayName = "Total Calls",
+                };
+                _currentCallsCounter ??= new PollingCounter("current-calls", this, () => _currentCalls)
+                {
+                    DisplayName = "Current Calls"
+                };
+                _callsFailedCounter ??= new PollingCounter("calls-failed", this, () => _callsFailed)
+                {
+                    DisplayName = "Total Calls Failed",
+                };
+                _callsDeadlineExceededCounter ??= new PollingCounter("calls-deadline-exceeded", this, () => _callsDeadlineExceeded)
+                {
+                    DisplayName = "Total Calls Deadline Exceeded",
+                };
+                _messagesSentCounter ??= new PollingCounter("messages-sent", this, () => _messageSent)
+                {
+                    DisplayName = "Total Messages Sent",
+                };
+                _messagesReceivedCounter ??= new PollingCounter("messages-received", this, () => _messageReceived)
+                {
+                    DisplayName = "Total Messages Received",
+                };
+                _callsUnimplementedCounter ??= new PollingCounter("calls-unimplemented", this, () => _callsUnimplemented)
+                {
+                    DisplayName = "Total Calls Unimplemented",
+                };
+            }
+        }
+    }
+}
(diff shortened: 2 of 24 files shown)


src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -257,9 +259,25 @@ private void EndCallCore()
...
             }
 
+            LogCallEnd();
+
             _callComplete = true;
         }
 
+        private void LogCallEnd()
+        {
+            var currentActivity = Activity.Current;
```
lmolkova: Activity for HTTP request is stopped when HtpClientHandler returns, and it depends on HttpCompletionOptions (i.e. when headers are received r when response is received and buffered).
And I thought per our email convresation that at the moment grpc call ends and trailer are received, http request is already finished. Did I get it wrong?

I.e. do we know for sure this is Activity created for HTTP outgoing request? Otherwise, this Activity is tracking something else (e.g. incoming request).

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -257,9 +259,25 @@ private void EndCallCore()
...
+            if (currentActivity != null)
+            {
+                currentActivity.AddTag("GrpcStatus", _status.StatusCode.ToTrailerString());
```
lmolkova: I suggest `grpc.status_code` similarly to `http.status_code` which is commonly used in OpenCeneus/Tracing/Telemetry as status tag on HTTP

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -324,6 +342,14 @@ protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
 
         public void Initialize()
         {
+            var currentActivity = Activity.Current;
+            if (currentActivity != null)
+            {
+                Activity.Current.AddTag("GrpcMethod", MethodCore);
```
lmolkova: similar to the above 

1. Do we know this is the right activity? 
2. `grpc.method`

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: This is the activity on the server. All this logic happens inside the request pipeline so the server activity should always be here.

Client is work in progress.

lmolkova: oh, I see, thanks for the info

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: Kestrel will start an activity, but there is the potential for someone else to start a different activity in middleware, which this logic would add a tag to. Is that a problem? Should this code check what the current activities name is? Instead of appending tags, should gRPC on the server always start a new activity just to be safe?

lmolkova: It's possible to traverse back to the first Activity that does not have Parent and it will be the one we care about. In most cases, it will be the Current one and won't create any overhead.

Creating new Activity means we'll need new DiagnosticsSource events, tracing tools will need to do some work to onboard, and it seems to create a lot of work and overhead without providing visible value. 
I guess if we'll need to host gRPC server without AspNetCore Hosting, new Activities it will be needed

src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
```diff
@@ -0,0 +1,159 @@
...
+using System.Diagnostics.Tracing;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using Grpc.Core;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal class GrpcEventSource : EventSource
+    {
+        public static readonly GrpcEventSource Log = new GrpcEventSource();
```
davidfowl: I'm not a fan of the static (even though it's the event source pattern), hosting adds the event source to the container.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -324,6 +342,14 @@ protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
 
         public void Initialize()
         {
+            var currentActivity = Activity.Current;
+            if (currentActivity != null)
+            {
+                Activity.Current.AddTag("grpc.method", MethodCore);
```
davidfowl: Who will see this? This happens within the GRPC pipeline itself and since no event is being written to the diagnostic source how will consumers get notified cc @lmolkova

JamesNK: When the hosting activity is stopped I presume that OpenTelemetry can get the tag values off the activity and record them.

src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
JamesNK: https://github.com/aspnet/AspNetCore/blob/3ce6c1dc5b022116ec9b13f9158e092323ac1158/src/Hosting/Hosting/src/Internal/HostingEventSource.cs#L13

No? Are you thinking of something else.

JamesNK: https://github.com/aspnet/AspNetCore/blob/3ce6c1dc5b022116ec9b13f9158e092323ac1158/src/Hosting/Hosting/src/GenericHost/GenericWebHostBuilder.cs#L80-L84

There is a DiagnosticsSource injected via DI. gRPC isn't creating one of those on the server.

davidfowl: Oh wait, that was another branch. Ignore me... 😄

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
davidfowl: If you have a request that starts out http then becomes GRPC because of the tag and is hanging for 1 minute, it won't be GRPC until that minute is up.

JamesNK: Yes. Do you want a new activity for gRPC? It feels like overkill for this situation to me.

Or do activities something like `ActivityName.Update` to raise on the diagnostic source? That could be used to tell the subscriber that the metadata on the activity has changed.

src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
```diff
@@ -0,0 +1,159 @@
...
+            Interlocked.Increment(ref _callsFailed);
+
+            WriteEvent(3, (int)statusCode);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [Event(eventId: 4, Level = EventLevel.Error)]
+        public void CallDeadlineExceeded()
+        {
+            Interlocked.Increment(ref _callsDeadlineExceeded);
```
JunTaoLuo: Should this increase the number of callsFailed?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JunTaoLuo: I'd lean towards a new gRPC activity. I don't think it would be too much overhead.

examples/Server/Startup.cs
```diff
@@ -44,14 +42,14 @@ public void ConfigureServices(IServiceCollection services)
...
-                    // This registers a global interceptor with a Scoped lifetime.
-                    options.Interceptors.Add<MaxStreamingRequestTimeoutInterceptor>(TimeSpan.FromSeconds(30));
                 })
                 .AddServiceOptions<GreeterService>(options =>
                 {
                     // This registers an interceptor for the Greeter service with a Singleton lifetime.
                     // NOTE: Not all calls should be cached. Since the response of this service only depends on the request and no other state, adding caching here is acceptable.
                     options.Interceptors.Add<UnaryCachingInterceptor>();
+                    // This registers an interceptor for the Greeter service with a Scoped lifetime.
+                    options.Interceptors.Add<MaxStreamingRequestTimeoutInterceptor>(TimeSpan.FromSeconds(30));
```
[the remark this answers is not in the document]
JamesNK: It was breaking examples that used streaming.

src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
JamesNK: It will be increased when the call ends.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
davidfowl: A new activity seems unfortunate as well especially because it's not exactly free. @lmolkova can we abuse OnActivityImport for this?

lmolkova: From the consumer side: AI or OT we'll use `OnActivityImport` to make sampling decision and it's hard to distinguish the first call that is intended for sampling from subsequent calls to update Activity state. It has to be done by using `OnActivityImport` object arg which is yet another contract between grpc and listener. 

Also for both AI and OT - we don't care much what happens with ASP.NET Core Activity until it is stopped - this is when the recording of activity happens. 

So I suggest that for now 
- we can be fine with the same Activity and just the Tag - it will be respected even though it appears after Activity starts at an unpredictable time. 
- using OnActivityImport will make default http case harder. 
- if we don't want to introduce new activity (and it's good), but has to notify someone about changes - just send another DS event like "Microsoft.AspNetCore.HttpIn.Change".

JamesNK: I've updated the server to write to the diagnostic source when the activity changes:

`_diagnosticSource.Write(currentActivity.OperationName + ".Change", null);`

JamesNK:
@lmolkova @davidfowl All feedback applied.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
davidfowl: Yea but don’t allocate , this does a concat every request

davidfowl: I just went digging around in the open telemetry source and found this:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/3538e05f0eed9df008d30e82a70bef11bd579bb5/src/OpenTelemetry.Collector.AspNetCore/Implementation/HttpInListener.cs#L108-L137

This has the exact same problem as GRPC so we should solve these similarly. 

- The request begins
- It becomes MVC after routing much like GRPC is decided after routing runs
- An event is fired that upgrades the current span

davidfowl: I think we should just add an event to the route matching to handle these types of cases. The endpoint should have enough metadata to determine what type of request it is as a result.

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -322,6 +342,15 @@ protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
         // Clock is for testing
         public void Initialize(ISystemClock? clock = null)
         {
+            var currentActivity = Activity.Current;
+            if (currentActivity != null)
+            {
+                Activity.Current.AddTag("grpc.method", MethodCore);
+                RaiseActivityChanged(currentActivity);
```
lmolkova: BTW, how do you know this is the activity ASP.NET Core created to track requests and not something created by user/other instrumentation site? Is it guaranteed?

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -322,6 +342,15 @@ protected override Task WriteResponseHeadersAsyncCore(Metadata responseHeaders)
         // Clock is for testing
         public void Initialize(ISystemClock? clock = null)
         {
+            var currentActivity = Activity.Current;
+            if (currentActivity != null)
+            {
+                Activity.Current.AddTag("grpc.method", MethodCore);
```
lmolkova:
```suggestion
                currentActivity .AddTag("grpc.method", MethodCore);
```

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -344,6 +373,22 @@ public void Initialize(ISystemClock? clock = null)
...
+        private void RaiseActivityChanged(Activity currentActivity)
+        {
+            const string ChangeSuffix = ".Change";
+            const string AspNetCoreHostingActivityName = "Microsoft.AspNetCore.Hosting.HttpRequestIn";
+            const string AspNetCoreHostingActivityChange = "Microsoft.AspNetCore.Hosting.HttpRequestIn" + ChangeSuffix;
+
+            // If the current activity is the ASP.NET Core hosting activity then use const string.
+            // This optimization avoids allocation is the most common use case.
+            var name = string.Equals(currentActivity.OperationName, AspNetCoreHostingActivityName, StringComparison.Ordinal)
+                ? AspNetCoreHostingActivityChange
```
lmolkova: this scares me a bit. so we only want to update activity that was created by ASP.NET Core. We shouldn't raise event in the context of child Activity created to track some user stuff.
I.e. if it is possible that `currentActivity.OperapertionName != AspNetCoreHostingActivityName`, we shouldn't add tags to it and should find another way to stamp tags on ASP.NET Core activity only.

JamesNK: I think the code that interacts with DiagnosticSource will go away with this PR - https://github.com/aspnet/AspNetCore/pull/11685

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
JamesNK: I think the code that interacts with DiagnosticSource will go away with this PR - https://github.com/aspnet/AspNetCore/pull/11685

lmolkova: ok, what about AddTag? is it guaranteed to be called on proper Activity.Current?

JamesNK: I don't know. What even would be the proper activity?

In my head I am imaging this will add tags to the request activity that is created by Kestrel host. Do you think it is better to check for the Kestrel host activity by name (also looking through parent activities)? The tags would then be added specifically to it rather than whatever the current activity happens to be. In that case, what should happen if Kestrel host activity is not present?

lmolkova: I mean imagine the following
1. Kestrel (Hosting) start activity A. it becomes Current
2. User starts activity in the middleware B. It becomes Current (it may be a child of A or user may break it and it won't be a child of A)
3. We are here in this code now, grpc.method is set on activity B.
4. User stops B
5. Kestrel stops A: AppInsights/OpenTelemetry records Activity A - it never understands that it was grpc Activity because tags were set on B. This is not what we want, we always want tags on A. 
Without these tags on A AppInsights will show it as usual HTTP span/telemetry and it will be confusing.

Is there any place where user code executes and can create another Activity between Hosting and HttpContextServerCallContext.Initialize()? If there is, we should better create another grpc activity. 

BTW, If user code executes in async method - they will not modify this Activity.Current - only in their execution context, so may be we are good.

davidfowl: The activity API is not easy enough for mere mortals to use. It’s actually insanely hard for us to get it right in the framework. Right now we have a single activity per request, are you saying we need one for each framework as well? Mvc, Grpc etc?

lmolkova: I agree it's extremely hard to use and having two per request is not good. If there is a way to add tag on initial activity created by Kestrel - do it.

If there is a way to get reference to the instance of that activity without checking Current - this is the best option. 

If  Current is the only way - check if it's one from Kestrel and do not set tag if it's not.

JamesNK: I've updated gRPC to only set tags on the host activity.

PR #342, opened 2019-06-19
Add Grpc.AspNetCore metapackage

JunTaoLuo:
This is the last package we intend to ship in 3.0 based on the diagram at https://github.com/grpc/grpc-dotnet/blob/master/doc/packages.md.

I've done some preliminary tests using a slightly modified grpc template, which ran correctly. I've also verified that the build and publish output contains only the dlls we expect so adding a dependency on Grpc.Tools with PrivateAssets=None is safe.

image

I think as another method of validation, I can convert our samples to now use the metapackage

### DIFF
--- a/src/Grpc.AspNetCore/Grpc.AspNetCore.csproj
+++ b/src/Grpc.AspNetCore/Grpc.AspNetCore.csproj
@@ -1,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Authors>The gRPC Authors</Authors>
+    <Copyright>Copyright 2019 The gRPC Authors</Copyright>
+    <Description>gRPC meta-package for ASP.NET Core</Description>
+    <PackageIconUrl>https://github.com/grpc/grpc.github.io/raw/master/img/grpc_square_reverse_4x.png</PackageIconUrl>
+    <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
+    <PackageProjectUrl>https://github.com/grpc/grpc-dotnet</PackageProjectUrl>
+    <PackageTags>gRPC RPC HTTP/2 aspnetcore</PackageTags>
+    <VersionPrefix>$(GrpcDotnetVersion)</VersionPrefix>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+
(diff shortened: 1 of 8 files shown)


JunTaoLuo:
We should create this package so we can update templates and the CLI with the newest packages.

src/Grpc.AspNetCore/Grpc.AspNetCore.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <VersionPrefix>$(GrpcDotnetVersion)</VersionPrefix>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
JamesNK: What does this do to the package?

src/Grpc.AspNetCore/Grpc.AspNetCore.csproj
```diff
@@ -0,0 +1,28 @@
...
+
+    <ProjectReference Include="..\Grpc.AspNetCore.Server.ClientFactory\Grpc.AspNetCore.Server.ClientFactory.csproj" />
+
+    <PackageReference Include="Google.Protobuf" Version="$(GoogleProtobufPackageVersion)" />
+    <PackageReference Include="Grpc.Tools" Version="$(GrpcPackageVersion)" PrivateAssets="None" />
```
JamesNK: What does `PrivateAssets="None"` do?

JamesNK: If you compile your website to run on a platform that Grpc.Tools doesn't support, what happens?

I worry about having this reference in the metapackage. Could it cause problems?

JamesNK: The scenario I was worried about was a compile error caused by a missing runtime. e.g. https://github.com/dotnet/corefx/issues/10709. I don't believe this is a problem for us.

src/Grpc.AspNetCore/Grpc.AspNetCore.csproj
JunTaoLuo: Copy pasta, I don't think this is needed.

src/Grpc.AspNetCore/Grpc.AspNetCore.csproj
JunTaoLuo: This ensures the `build/` assets are propagated to the parent (i.e. the app), otherwise the targets required for code gen is not propagated by default.

JamesNK: Add that info as a comment

JunTaoLuo:
I'll merge this now to unblock other work, but feel free to provide feedback @jtattermusch and I'll address them separately.

jtattermusch:
LGTM.

PR #348, opened 2019-06-22
Add EventSource and new activity to client

JamesNK:
@lmolkova

### DIFF
--- a/src/Grpc.Net.Client/Internal/GrpcCall.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCall.cs
@@ -160,6 +160,10 @@ namespace Grpc.Net.Client.Internal
         {
             if (!Disposed)
             {
+                // Locking on the call because:
+                // 1. Its not exposed publically
+                // 2. Nothing else locks on call
+                // 3. We want to avoid allocating a private lock object
                 lock (this)
                 {
                     if (!Disposed)
--- a/src/Grpc.Net.Client/Internal/GrpcCall.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCall.cs
@@ -36,6 +36,7 @@ namespace Grpc.Net.Client.Internal
         where TResponse : class
     {
         private readonly CancellationTokenSource _callCts;
+        private readonly TaskCompletionSource<StatusCode> _callTcs;
         private readonly CancellationTokenRegistration? _ctsRegistration;
         private readonly TimeSpan? _timeout;
         private readonly Uri _uri;
@@ -45,7 +46,7 @@ namespace Grpc.Net.Client.Internal
         private Metadata? _trailers;
         private string? _headerValidationError;
         private TaskCompletionSource<Stream>? _writeStreamTcs;
-        private TaskCompletionSource<bool>? _completeTcs;
+        private TaskCompletionSource<bool>? _writeCompleteTcs;
 
         public bool DeadlineReached { get; private set; }
         public bool Disposed { get; private set; }
@@ -66,6 +67,7 @@ namespace Grpc.Net.Client.Internal
             ValidateDeadline(options.Deadline);
 
             _callCts = new CancellationTokenSource();
+            _callTcs = new TaskCompletionSource<StatusCode>(TaskContinuationOptions.RunContinuationsAsynchronously);
             Method = method;
             _uri = new Uri(method.FullName, UriKind.Relative);
             _logScope = new GrpcCallScope(method.Type, _uri);
@@ -114,21 +116,21 @@ namespace Grpc.Net.Client.Internal
         {
             var message = CreateHttpRequestMessage();
             SetMessageContent(request, message);
-            StartSend(client, message);
+            _ = StartAsync(client, message);
         }
 
         public void StartClientStreaming(HttpClient client)
         {
             var message = CreateHttpRequestMessage();
             ClientStreamWriter = CreateWriter(message);
-            StartSend(client, message);
+            _ = StartAsync(client, message);
         }
 
         public void StartServerStreaming(HttpClient client, TRequest request)
         {
             var message = CreateHttpRequestMessage();
             SetMessageContent(request, message);
-            StartSend(client, message);
+            _ = StartAsync(client, message);
             ClientStreamReader = new HttpContentClientStreamReader<TRequest, TResponse>(this);
         }
 
@@ -136,7 +138,7 @@ namespace Grpc.Net.Client.Internal
         {
             var message = CreateHttpRequestMessage();
             ClientStreamWriter = CreateWriter(message);
-            StartSend(client, message);
+            _ = StartAsync(client, message);
             ClientStreamReader = new HttpContentClientStreamReader<TRequest, TResponse>(this);
         }
 
@@ -158,35 +160,45 @@ namespace Grpc.Net.Client.Internal
         {
             if (!Disposed)
             {
-                Disposed = true;
-
-                if (!ResponseFinished)
+                lock (this)
                 {
-                    // If the response is not finished then cancel any pending actions:
-                    // 1. Call HttpClient.SendAsync
-                    // 2. Response Stream.ReadAsync
-                    // 3. Client stream
-                    //    - Getting the Stream from the Request.HttpContent
-                    //    - Holding the Request.HttpContent.SerializeToStream open
-                    //    - Writing to the client stream
-                    CancelCall();
-                }
-                else
-                {
-                    _writeStreamTcs?.TrySetCanceled();
-                    _completeTcs?.TrySetCanceled();
-                }
+                    if (!Disposed)
+                    {
+                        Disposed = true;
 
-                _ctsRegistration?.Dispose();
-                _deadlineTimer?.Dispose();
-                HttpResponse?.Dispose();
-                ClientStreamReader?.Dispose();
-                ClientStreamWriter?.Dispose();
+                        if (!ResponseFinished)
+                        {
+                            // If the response is not finished then cancel any pending actions:
+                            // 1. Call HttpClient.SendAsync
+                            // 2. Response Stream.ReadAsync
+                            // 3. Client stream
+                            //    - Getting the Stream from the Request.HttpContent
+                            //    - Holding the Request.HttpContent.SerializeToStream open
+                            //    - Writing to the client stream
+                            CancelCall();
+                        }
+                        else
+                        {
+                            _writeStreamTcs?.TrySetCanceled();
+                            _writeCompleteTcs?.TrySetCanceled();
+                        }
 
-                // To avoid racing with Dispose, skip disposing the call CTS
-                // This avoid Dispose potentially calling cancel on a disposed CTS
-                // The call CTS is not exposed externally and all dependent registrations
-                // are cleaned up
+                        // If response has successfully finished then the status will come from the trailers
+                        // If it didn't finish then complete with a Cancelled status
+                        _callTcs.TrySetResult(StatusCode.Cancelled);
+
+                        _ctsRegistration?.Dispose();
+                        _deadlineTimer?.Dispose();
+                        HttpResponse?.Dispose();
+                        ClientStreamReader?.Dispose();
+                        ClientStreamWriter?.Dispose();
+
+                        // To avoid racing with Dispose, skip disposing the call CTS
+                        // This avoid Dispose potentially calling cancel on a disposed CTS
+                        // The call CTS is not exposed externally and all dependent registrations
+                        // are cleaned up
+                    }
+                }
             }
         }
 
@@ -312,6 +324,8 @@ namespace Grpc.Net.Client.Internal
                         throw new InvalidOperationException("Call did not return a response message");
                     }
 
+                    GrpcEventSource.Log.MessageReceived();
+
                     // The task of this method is cached so there is no need to cache the message here
                     return message;
                 }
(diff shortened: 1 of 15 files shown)


src/Grpc.Net.Client/Internal/GrpcEventSource.cs
```diff
@@ -0,0 +1,162 @@
...
+using Grpc.Core;
+
+namespace Grpc.Net.Client.Internal
+{
+    internal class GrpcEventSource : EventSource
+    {
+        public static readonly GrpcEventSource Log = new GrpcEventSource();
+
+#if HAVE_DIAGNOSTICS_COUNTERS
+        private PollingCounter? _totalCallsCounter;
```
JamesNK: Is PollingCounter available outside of .NET Core 3.0? Current this package only targets netstandard2.1

davidfowl: Nope. Counters are not in netstandard (at least not yet) cc @terrajobst

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -157,35 +159,61 @@ private void DisposeCore()
...
+                        Disposed = true;
+
+                        var resolvedStatus = _responseStatus ?? StatusCode.Cancelled;
+                        if (resolvedStatus != StatusCode.OK)
+                        {
+                            GrpcEventSource.Log.CallFailed(resolvedStatus);
+                        }
+                        GrpcEventSource.Log.CallStop();
+
+                        if (_activity != null)
```
lmolkova: do you know if  Dispose (i.e. Activity.Stop) happens in the same execution context as Start?

Basically, if someone does this:

```csharp
  GrpcCall call = await Task.Run (() => return GrpcCall());
  await call .SendTask;
  call.Dispose();
```

will cause a bugs that are very hard to find:
- `Task.Run (() => return GrpcCall());` will create activity that will only exists inside this task
- `Dispose` will attempt to stop the instance of this activity, but listener (tracer) will attempt to get info from `Activity.Current` - it doesn't know which instance of activity is stopped and always deals with current.

another one:
if when grpc call is created, Current was A and when it was Disposed it was B, after dispose Current will turn to A
```csharp
  var a = new Actitvity("a").Start();
  GrpcCall call = await Task.Run (() => return GrpcCall());
  var res = await call.SendTask;
  var b = new Actitvity("b").Start();
  call.Dispose();

  Console.WriteLine(Activity.Current.OperationName); // you'd expect it to be 'b', but it's 'a'
```

It seems to affect logger scope in the same way (some loggers put scopes in AsyncLocal). I don't know if it matters or may be this API is intended to be used in only one way that never allows this. 
If not, I suggest to move activity stop to the end of `SendAsync` - this will guarantee that it's actually the same execution path or at least leave a comment somewhere that it's a danger zone.

src/Grpc.Net.Client/Internal/GrpcEventSource.cs
JamesNK: What about an external package? I tried to find a package that contained PollingCounter and couldn't find one.

src/Grpc.Net.Client/Internal/GrpcCall.cs
JamesNK: I don't think SendAsync will work. The lifetime of a gRPC isn't start request -> get response. A gRPC call ends in one of these situations:

1. The response is received and http status or headers are incorrect
2. Error when sending or receiving message
3. Deadline timer executes
4. User calls Dispose on call object
5. The response body has finished reading

The deadline timer is definitely on a different thread. All the others ends could happen on a different thread.

lmolkova: different threads and callbacks are not necessarily a problem as long as they are async-friendly. AsyncLocal flows through async calls and doesn't care about threads.
But it seems it's not the case here.

There may be some tricks with execution context (capture and restore it), but I'm no expert and don't know if it is a good idea. Maybe @davidfowl knows who can help with execution context?



The ~~hacky~~ ways I see here are:

In dispose, check if _activity == Current. If it is - we are good (but it seems not very likely- who knows where user disposes calls)
If it's not current, 
 * keep old current value in local var
 * set Activity.Current  to _activity (bummer, public Current setter comes with DiagnosticSource 4.6.0)
 * After Stop completes return old current back

Another way: Put _activity instance into Stop event payload and expect listener do a special treatment (know the activity.current does not work). I don't like it as it additional contract and if listener doesn't know about it, it leads to hard-to-debug issues. Also, in general, we want to have generic listeners that don't know anything about payloads and subscribe to all stuff in the same way.

JamesNK: I've refactored code to stop the activity via CancellationToken.Register. That will capture the execution context.

A test of your situation passes.

src/Grpc.Net.Client/Internal/GrpcDiagnostics.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Diagnostics;
+
+namespace Grpc.Net.Client.Internal
+{
+    internal static class GrpcDiagnostics
+    {
+        // This is a static on a non-generic class so it isn't re-created once for each type argument
+        public static readonly DiagnosticListener DiagnosticListener = new DiagnosticListener("GrpcCallDiagnosticListener");
+
+        public const string ActivityName = "Grpc.Net.Client.GrpcOut";
```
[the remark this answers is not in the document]
JamesNK: I followed what I saw here - https://github.com/aspnet/AspNetCore/blob/6ecfdf2c49ffe25ebc3e8872d186bb85fe8e26b9/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L18

So you think the name should just be `GrpcOut`?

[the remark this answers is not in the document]
JamesNK: Someone could look at the activity without using the DiagnosticSource. Wouldn't the namespace still be useful in that situation?

src/Grpc.Net.Client/Internal/GrpcCall.cs
JamesNK: Refactored to use a TCS instead of CT.Register

src/Grpc.Net.Client/Internal/GrpcDiagnostics.cs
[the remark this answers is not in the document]
lmolkova: yes, I meant `GrpcOut`.

you are right, someone can try creating/using activities without DiagnosticSource (e.g. logger) and name might be useful there in some cases. 

At the same time, Activity name so far is only used as a part of DiagnosticSource event (to subscribe, filter and apply custom payload parsing in ApplicationInsights or OpenTelemetry), so I doubt it will be useful without DS.

I'm ok keeping it to `Grpc.Net.Client.GrpcOut` if you think it may be useful.

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -419,22 +446,67 @@ private void StartSend(HttpClient client, HttpRequestMessage message)
...
+                {
+                    var statusText = statusCode.ToString("D");
+                    if (statusText != null)
+                    {
+                        activity.AddTag(GrpcDiagnostics.GrpcStatusCodeTagName, statusText);
+                    }
+
+                    if (diagnosticSourceEnabled)
+                    {
+                        GrpcDiagnostics.DiagnosticListener.Write(GrpcDiagnostics.ActivityStopKey, HttpResponse);
```
lmolkova: if you don't want to use StopActivity(), please set end time before calling Write (check how StopActivity does it internally)

lmolkova: also typically DS events are used with payload in an anonymous type , check out here https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/DiagnosticSourceUsersGuide.md#payloads

this is needed to allow non-breaking addition of new payload property.

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -419,22 +446,67 @@ private void StartSend(HttpClient client, HttpRequestMessage message)
...
+                // 3. There is an existing activity (to enable activity propagation)
+                if (diagnosticSourceEnabled || Logger.IsEnabled(LogLevel.Critical) || Activity.Current != null)
+                {
+                    activity = new Activity(GrpcDiagnostics.ActivityName);
+                    activity.AddTag(GrpcDiagnostics.GrpcMethodTagName, Method.FullName);
+                    activity.Start();
+
+                    if (diagnosticSourceEnabled)
+                    {
+                        GrpcDiagnostics.DiagnosticListener.Write(GrpcDiagnostics.ActivityStartKey, request);
```
lmolkova: please use anonymous  type  (check out comment for Stop())

lmolkova: it could make sense to check if somebody wants Start event - they don't necessarily do as everything can be logged when Stop is called.

src/Grpc.Net.Client/Internal/GrpcCall.cs
davidfowl: Perf! 😄

lmolkova: BTW, perf impact likely comes from the allocation, our reflection tricks keep the impact of property fetching low. 

The chance  change payloads need to change is probably very low, so I'm ok with explicit request/response

cwe1ss: Adding my usual rant about the DiagnosticSource usages already being very inconsistent. This would be another deviation from the official docs/recommendations. I’m all about perf, so maybe consider changing those docs if they are not practical/followed anyway?!

lmolkova: nice to see you again @cwe1ss and looking forward working with you on OpenTelemetry!

there is no way to leverage payloads in a generic way - so there is always a very specific contract between instrumentation and listener. All listeners that want to stay agnostic will ignore payload.

I think docs cover it

```
CONSIDER - if you have an event that is so frequent that the performance of the logging is an important consideration, and you have only one data item and it is unlikely that you will ever have more data to pass to the event, and the data item is a normal class (not a value type) then you save some cost by simply by passing the data object directly without using an anonymous type wrapper
```

cwe1ss: I’ve just seen that passing in a single data object is a CONSIDER item in the docs, so they should be fine. I still don’t like the inconsistencies though 😄

JamesNK:
@lmolkova @davidfowl All feedback applied.

src/Grpc.Net.Client/Internal/GrpcEventSource.cs
JamesNK: They're in netstandard2.1 now 🎉

JamesNK:
@lmolkova @davidfowl Rebased on latest and PollingCounter is now in netstandard2.1. I consider this done. Final review/approval please 🙏

src/Grpc.Net.Client/Internal/GrpcEventSource.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+        }
+
+        // Used for testing
+        internal GrpcEventSource(string eventSourceName)
+            : base(eventSourceName)
+        {
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
BrennanConroy: Is this important?

src/Grpc.Net.Client/Internal/HttpContentClientStreamWriter.cs
```diff
@@ -126,13 +126,15 @@ private async Task WriteAsyncCore(TRequest message)
...
-                await writeStream.WriteMessage<TRequest>(
+                await writeStream.WriteMessageAsync<TRequest>(
                     _call.Logger,
                     message,
                     _call.Method.RequestMarshaller.ContextualSerializer,
                     _grpcEncoding,
                     _call.CallInvoker.SendMaxMessageSize,
                     _call.CancellationToken).ConfigureAwait(false);
+				
+				GrpcEventSource.Log.MessageSent();
```
BrennanConroy: nit: formatting

src/Grpc.Net.Client/Internal/GrpcEventSource.cs
JamesNK: I believe it is required because WriteEvent needs to capture metadata off the method attribute.

I copied them from https://github.com/aspnet/AspNetCore/blob/04bf1bf32e13d35d4d7f61f9cbeb9bfa22e0c617/src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelEventSource.cs#L41

BrennanConroy: That doesn't sound right, seems like that could easily break code if the compiler could remove attributes by optimizing a method.

I couldn't find any docs about this being required, and SignalR doesn't have it and our counters seem to be working...

test/Grpc.Net.Client.Tests/DiagnosticsTests.cs
```diff
@@ -0,0 +1,218 @@
...
+            Assert.AreEqual(2, result.Count);
+            Assert.AreEqual(GrpcDiagnostics.ActivityStartKey, result[0].Key);
+            Assert.AreEqual(requestMessage, GetValueFromAnonymousType<HttpRequestMessage>(result[0].Value, "Request"));
+            Assert.AreEqual(GrpcDiagnostics.ActivityStopKey, result[1].Key);
+            Assert.AreEqual(requestMessage, GetValueFromAnonymousType<HttpRequestMessage>(result[1].Value, "Request"));
+            Assert.AreEqual(responseMessage, GetValueFromAnonymousType<HttpResponseMessage>(result[1].Value, "Response"));
+        }
+
+        [Test]
+        public void DiagnosticListener_MakeCall_ActivityHasNameAndDuration()
```
JunTaoLuo: Hmm looks like some of the logic is shared across this test and the previous. Maybe combine them?

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -158,35 +160,45 @@ private void DisposeCore()
...
-                    // 1. Call HttpClient.SendAsync
-                    // 2. Response Stream.ReadAsync
-                    // 3. Client stream
-                    //    - Getting the Stream from the Request.HttpContent
-                    //    - Holding the Request.HttpContent.SerializeToStream open
-                    //    - Writing to the client stream
-                    CancelCall();
-                }
-                else
+                lock (this)
```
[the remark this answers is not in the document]
JamesNK: Avoids allocating. GrpcCall is never exposed publically. I'll add a comment

PR #352, opened 2019-06-25
Flow transitive build targets for project references

JunTaoLuo:
Right now the build targets that override LiteClientBase is only included in Grpc.Net.Client. The packages that transitively depend on this package (such as Grpc.Net.ClientFactory) do not flow those targets so referencing them would lead to type missing errors at build time (since the override did not occur and the base class resolves to ClientBase instead).

Though it's documented at https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#project-to-project-references, PrivateAssets=All on a ProjectReference did not translate to a PrivateAssets=All in the nuspec during pack. As a fallback, I'm adding the override target to all packages that transitively reference Grpc.Net.Client. I've tested that this resolves the issue even when multiple of these packages are referenced together. Though it can be argued that users shouldn't do this, at least this doesn't lead to catastrophic failure.

I need to check with the nuget folks to see if there's a way to propagate the PrivateAssets settings on a project to project reference to the nuspec at pack time.


JunTaoLuo:
Actually, it seems like the results of pack were being cached. PrivateAssets seems to work on a clean build. Will update PR.

src/Grpc.AspNetCore.Server.ClientFactory/Grpc.AspNetCore.Server.ClientFactory.csproj
```diff
@@ -26,7 +26,7 @@
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
 
     <ProjectReference Include="..\Grpc.AspNetCore.Server\Grpc.AspNetCore.Server.csproj" />
-    <ProjectReference Include="..\Grpc.Net.ClientFactory\Grpc.Net.ClientFactory.csproj" />
+    <ProjectReference Include="..\Grpc.Net.ClientFactory\Grpc.Net.ClientFactory.csproj" PrivateAssets="None" />
```
JamesNK: Add comment explaining why `PrivateAssets=false` to all the places

PR #356, opened 2019-06-26
prepare for the 0.1.22-pre2 release

jtattermusch:
I will merge myself just before making the release.


jtattermusch:
> When will the release be made? I would like to get #271 in for the release.

As soon as you let me know that preview7 is out ;-)

PR #359, opened 2019-06-27
Update to latest nightly .NET Core SDK

JamesNK:


JunTaoLuo:
haha I'm making the same changes locally.

JunTaoLuo:
How are you not running into the error ` MSB3822: Preserialized resources require the System.Resources.Extensions assembly at runtime, but it was not found in this project's references.`

JunTaoLuo:
Looks like you are

JamesNK:
> Looks like you are

Not locally 🤷‍♂

examples/Server/TicketRepository.cs
```diff
@@ -35,7 +35,7 @@ public int GetAvailableTickets()
             return _availableTickets;
         }
 
-        public bool BuyTickets(string user, int count)
+        public bool BuyTickets(string? user, int count)
```
JunTaoLuo: Looks like we resolved the nullable issue differently

JamesNK: Undo this

JamesNK: There will always be a user with a name because the method forces authorization. My fix is better 😋

Issue #232, opened 2019-05-06
Client compression


JamesNK:
Related https://github.com/grpc/grpc-dotnet/issues/43

PR #362, opened 2019-06-29
Client compression

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/232


JamesNK:
@jtattermusch Please review that this is doing client compression correctly.

src/Grpc.Net.Client/Internal/Compression/GzipCompressionProvider.cs
```diff
@@ -0,0 +1,66 @@
+﻿#region Copyright notice and license
```
JunTaoLuo: This is the same as the GzipCompressionProvider in Grpc.AspNetCore.Server. Can we create a shared folder and add it to the Compile itemgroup instead?

src/Grpc.Net.Client/Internal/Compression/ICompressionProvider.cs
```diff
@@ -0,0 +1,49 @@
+﻿#region Copyright notice and license
```
JunTaoLuo: Same here

src/Grpc.Net.Client/Internal/Compression/GzipCompressionProvider.cs
JamesNK: This is internal, the server is not. I might make the server compression API internal, but unifying code will bloat this PR. Will do it separately.

testassets/InteropTestsClient/RunTests.ps1
```diff
@@ -23,10 +23,9 @@ $allTests =
   "status_code_and_message",
   "special_status_message",
   "unimplemented_service",
-  "unimplemented_method"
-  #,
-  #"client_compressed_unary",
-  #"client_compressed_streaming"
+  "unimplemented_method",
+  "client_compressed_unary",
```
jtattermusch: I assume these are passing now?

JamesNK: Yes

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -506,18 +517,26 @@ private HttpRequestMessage CreateHttpRequestMessage()
...
                     {
+                        // grpc-timeout is set via CallOptions.Deadline
                         continue;
                     }
-
-                    AddHeader(message.Headers, entry);
+                    else if (entry.Key == GrpcProtocolConstants.CompressionRequestAlgorithmHeader)
+                    {
+                        // grpc-internal-encoding-request is used in the client to set message compression
+                        message.Headers.Add(GrpcProtocolConstants.MessageEncodingHeader, entry.Value);
```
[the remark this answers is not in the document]
JamesNK: That's what I thought.

When we have the concept of channel I'd like to add a strongly typed property for compression. I think it is fine to still have this special header then, it doesn't hurt anyone.

PR #363, opened 2019-06-29
Replace sync-over-async Assert.ThrowsAsync

JamesNK:
Build has started hanging since the SDK was updated. Trying to figure out what is going on.


JamesNK:
I don't know what is causing it but async-sync-async from NUnit's `Assert.ThrowsAsync` can't help.

JamesNK:
I think it is just a bug in .NET 🤷‍♂ 

I think these test changes are still worth merging.

jtattermusch:
> I don't know what is causing it - sometimes it passes, other times it hangs - but async-sync-async from NUnit's `Assert.ThrowsAsync` can't help. I've replaced it with a fully async ThrowsAsync.
> 
> Lets merge this and see whether the failures disappear.

I think NUnit's Assert.ThrowsAsync is basically broken by design
https://github.com/nunit/nunit/blob/a56e6859f3609128128ef87b0226472eb63ea5d4/src/NUnitFramework/framework/Assert.Exceptions.Async.cs#L48. I think it can be proven that in some scenarios it has the potential to deadlock. This might have to do with the way NUnit works execute test cases. I guess there's going to be two Wait() invocations somewhere blocking each other.

PR #364, opened 2019-07-01
Support CallOptions.CallCredentials

JamesNK:


src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -445,6 +451,32 @@ private async Task SendAsync(HttpClient client, HttpRequestMessage message)
...
+
+            if (configurator.Interceptor != null)
+            {
+                var authInterceptorContext = new AuthInterceptorContext(client.BaseAddress.OriginalString, Method.FullName);
+                var metadata = new Metadata();
+                await configurator.Interceptor(authInterceptorContext, metadata).ConfigureAwait(false);
+
+                foreach (var entry in metadata)
+                {
+                    AddHeader(message.Headers, entry);
```
jtattermusch: one feature of CallCredentials is that the auth metadata entry is only added if we're on a secure channel, otherwise it is stripped. This is to avoid sending sensitive tokens over plaintext connections. If possible, we should make the behavior of managed client consistent.
This check is actually provided by C-core, I will try to find the code that implements it.

JamesNK:
@jtattermusch Is this good to merge now?

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -425,6 +426,14 @@ private void StartSend(HttpClient client, HttpRequestMessage message)
 
         private async Task SendAsync(HttpClient client, HttpRequestMessage message)
         {
+            // In C-Core the call credential auth metadata is only applied if the channel is secure
+            // The equivalent in grpc-dotnet is only applying metadata if HttpClient is using TLS
+            // HttpClient scheme will be HTTP if it is using H2C (HTTP2 without TLS)
+            if (Options.Credentials != null && client.BaseAddress.Scheme == Uri.UriSchemeHttps)
```
jtattermusch: This seems good enough from the security perspective, but it might not be exactly what C-core does.
I'm actually not sure if C-core invokes the AuthInterceptor callback always and then strips away the metadata entries generated by the callback for insecure channels or if it just skips the call to the interceptor entirely.

If C-core always invokes the auth interceptor callback (even for insecure channels), we should probably do the same (and then not add the metadata to the call and perhaps print a debug log message).

JamesNK: I tested and the interceptor isn't run at all for unsecure.

PR #375, opened 2019-07-07
Update .NET Core SDK

JamesNK:


JunTaoLuo:
It seems like a hang somewhere. Let's see if it passes on a rerun

Issue #357, opened 2019-06-26
Support missing for contextual deserializer/serializer


mgravell:
Right now, the code directly hooks into the legacy `.Serializer` / `.Deserializer`, which provides access to the `Func<T, byte[]>` etc, however IMO this violates the API.

The *primary* API is now the contextual serializer, by which I mean:

- Grpc.Core[.Api] always uses the contextual serializer (except for some tests)
- the contextual serializer *always works* (it is emulated when initialized from the legacy API), where-as the legacy serializer **does not**; see [Marshaller.cs 44-47](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core.Api/Marshaller.cs#L44-L47)

``` c#
// gRPC only uses contextual serializer/deserializer internally, so emulating the legacy
// (de)serializer is not necessary.
this.serializer = (msg) => { throw new NotImplementedException(); };
this.deserializer = (payload) => { throw new NotImplementedException(); };
```

This has two significant problems:

1. any custom framework that tries to use the contextual serializer *will not work* from grpc-dotnet
2. any improvements made to the core implementation of `Marshaller` will not be automatically exposed

For context on 2, see [this PR](https://github.com/grpc/grpc/pull/19460), which does an in-place upgrade on how regular protoc-style protobuf deserializes, using the contextual API to significantly reduce buffer allocations. If grpc-dotnet *does not use* the contextual API: it will not benefit if/when those changes get merged.

I am aware of https://github.com/grpc/grpc-dotnet/issues/30, but IMO there are two different things:

- #30 is about making better use of these APIs in the future
- this report is about a fundamental bug that breaks the API **today**

You can also probably infer from this that I'm keen on using the contextual API, [and generally improving this space](https://github.com/grpc/grpc/issues/19447) - presumably looking into a new writer API too.

So, I guess my main point here is: grpc-dotnet should **not** use the legacy API - it should switch to the contextual API. I can offer some help with that if it would be welcome.

---

As a caveat / disclosure thing: I *will* acknowledge the remark:

``` c#
/// Note: experimental API that can change or be removed without any prior notice.
```

but... if the API surface changes, consumers need to update *anyway*. I don't personally see this as a reason not to consume this API.

mgravell:
Another option of course being to emulate it with a spoof ctx; that should be pretty easy too - happy to help

mgravell:
I've added a PR to the core that (if accepted) will fix "1" above; to fix "2" requires taking the experimental API

[...]

PR #376, opened 2019-07-09
Use contextual serialization

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/357

@mgravell

### DIFF
--- a/src/Grpc.Net.Client/Internal/StreamExtensions.cs
+++ b/src/Grpc.Net.Client/Internal/StreamExtensions.cs
@@ -26,6 +26,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Grpc.Core;
 using Grpc.Net.Client.Internal;
+using Grpc.Shared;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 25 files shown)


src/Grpc.Net.Client/Internal/StreamExtensions.cs
```diff
@@ -156,12 +157,18 @@ private static Status CreateUnknownMessageEncodingMessageStatus(string unsupport
                         throw new RpcException(CreateUnknownMessageEncodingMessageStatus(grpcEncoding, GrpcProtocolConstants.CompressionProviders.Select(c => c.EncodingName)));
                     }
 
+#pragma warning disable CS8600 // Converting null literal or possible null value to non-nullable type.
```
JamesNK: Nullable attribute isn't working here for some reason. Will clean it up when bug in compiler is fixed. Only effects designer.

mgravell:
Awesome. Sorry for bringing you a few hundred lines of work, but I think it moves things in the right direction. Appreciated!

JunTaoLuo:
Why is our tests hanging? I'm inclined to to think there's actually something wrong with them now.

PR #378, opened 2019-07-09
Building and packing Grpc.AspNetCore.Server.Reflection

gengle:
This should produce an assembly in the correct location.


thelinuxfoundation:

* User @gengle isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

JamesNK:
Not publishing it is intentional. We haven't decided if it is something we want to support officially.

Also the package currently depends on Grpc.Reflection which has a dependency on Grpc.Core, which we don't want. We would want to either remove the Grpc.Reflection dependency or change Grpc.Reflection to depend on Grpc.Core.Api.

JunTaoLuo:
We will publish this package when we are ready.

Issue #370, opened 2019-07-05
ClientBaseType not used?


JamesNK:
Grpc.Net.Client.Tests.csproj has a protobuf reference to a generated client that should inherit from `ClientBase`:

```
<Protobuf Include="Proto\core-greet.proto" GrpcServices="Client" ClientBaseType="ClientBase" />
```

But in generated code it is still inheriting from LiteClientBase:

![image](https://user-images.githubusercontent.com/303201/60691462-dbb11c80-9f23-11e9-8f05-90a612f23ef8.png)

1. Fix issue
2. Add code to unit test that asserts the client type inherits from `ClientBase`.

PR #379, opened 2019-07-09
Add tests for ClientBase type overrides

JunTaoLuo:
Addresses https://github.com/grpc/grpc-dotnet/issues/370.

In the meantime before decisions regarding ChannelBase are finalized, let's add tests for the override logic. The first commit is the one that adds tests. The second one is just clean up.

### DIFF
--- a/src/dotnet-grpc/Commands/CommandBase.cs
+++ b/src/dotnet-grpc/Commands/CommandBase.cs
@@ -161,7 +161,7 @@ namespace Grpc.Dotnet.Cli.Commands
                 // If file is outside of the project, display the file under Protos/ directory
                 if (!Path.GetFullPath(resolvedPath).StartsWith(Project.DirectoryPath, StringComparison.OrdinalIgnoreCase))
                 {
-                    newItem.Xml.AddMetadata(LinkElement, Path.Combine(ProtosFolder, Path.GetFileName(file)!));
+                    newItem.Xml.AddMetadata(LinkElement, Path.Combine(ProtosFolder, Path.GetFileName(file)!), expressAsAttribute: true);
                 }
             }
         }
--- a/test/Grpc.AspNetCore.Server.ClientFactory.Tests/ClientBaseTests.cs
+++ b/test/Grpc.AspNetCore.Server.ClientFactory.Tests/ClientBaseTests.cs
@@ -25,6 +25,7 @@ namespace Grpc.Net.Client.Tests
     [TestFixture]
     public class ClientBaseTests
     {
+        [Test]
         public void ClientBaseClass_Implicit_UsesLiteClientBase()
         {
             // Assert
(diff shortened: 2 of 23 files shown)


test/Grpc.Net.Client.Tests/Grpc.Net.Client.Tests.csproj
```diff
@@ -31,10 +32,4 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Testing" Version="$(MicrosoftExtensionsPackageVersion)" />
   </ItemGroup>
 
-  <ItemGroup>
```
JunTaoLuo: Not sure why I added this override but this was the root cause of the linked issue.

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -161,7 +161,7 @@ public void AddProtobufReference(Services services, string additionalImportDirs,
                 // If file is outside of the project, display the file under Protos/ directory
                 if (!Path.GetFullPath(resolvedPath).StartsWith(Project.DirectoryPath, StringComparison.OrdinalIgnoreCase))
                 {
-                    newItem.Xml.AddMetadata(LinkElement, Path.Combine(ProtosFolder, Path.GetFileName(file)!));
+                    newItem.Xml.AddMetadata(LinkElement, Path.Combine(ProtosFolder, Path.GetFileName(file)!), expressAsAttribute: true);
```
JunTaoLuo: Link attributes seem relatively short and it seems more compact to express it as an attribute instead.

test/Grpc.AspNetCore.Server.ClientFactory.Tests/ClientBaseTests.cs
```diff
@@ -0,0 +1,34 @@
...
+using Greet;
+using Grpc.Core;
+using NUnit.Framework;
+
+namespace Grpc.Net.Client.Tests
+{
+    [TestFixture]
+    public class ClientBaseTests
+    {
+        public void ClientBaseClass_Implicit_UsesLiteClientBase()
```
JamesNK: Add `[Test]`

test/Shared/ClientBaseTests.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using Grpc.Core;
+using NUnit.Framework;
+
+namespace Grpc.Net.Client.Tests
+{
+    [TestFixture]
+    public class ClientBaseTests
+    {
+        public void ClientBaseClass_ExplicitSetting_UsesLiteClientBase()
```
JamesNK: +Attribute

test/Shared/ClientBaseTests.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            // Assert
+            Assert.True(typeof(Greet.Greeter.GreeterClient).IsSubclassOf(typeof(LiteClientBase)));
+        }
+
+        public void ClientBaseClass_ExplicitSetting_UsesClientBase()
```
JamesNK: +Attribute

PR #380, opened 2019-07-09
Update test script

JunTaoLuo:
Run each test in series. Hopefully shedding more light into which one hangs.

### DIFF
--- a/build_and_test.sh
+++ b/build_and_test.sh
@@ -23,13 +23,13 @@ dotnet build
 
 echo "Testing solution"
 
-# Capturing test diagnostic logs because of hanging build
-# https://github.com/grpc/grpc-dotnet/pull/363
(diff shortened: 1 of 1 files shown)


build_and_test.sh
```diff
@@ -23,8 +23,13 @@ dotnet build
 
 echo "Testing solution"
 
-# Capturing test diagnostic logs because of hanging build
-# https://github.com/grpc/grpc-dotnet/pull/363
-dotnet test --no-build --diag:log.txt
+test_projects=( $( ls test/**/*Tests.csproj ) )
```
jtattermusch: sg, but we need to be careful to not accidentally stop running all tests after some rename/refactor in the future.
Should we add a line that smoke tests that test_project contains at least one of the test projects?

Issue #383, opened 2019-07-10
API review feedback


JunTaoLuo:
## Interceptor and service activation
- [x] use handler and make activator singletons
- [x] make service activators and interceptor activators internal
- [x] add AddRange(IEnumerable) to InterceptorCollection
- [x] InterceptorCollection inherit from `Collection<T>`
- [x] Expose arguments of IntercetptorRegistration
## Compression
- [x] ~Make CompressionProvider internal~
## Others
- [x] IServerCallContextFeature, move to root namespace
## Service method discovery
- [x] ~The current extensibility is too complicated~
- [x] ~Consider exposing the list of discovered methods~
## Routing
- [x] Strongly typed endpoint convention builder
## Client factory
- [x] ~AddGrpcClient return internal wrapper over IhttpClientBuilder~
- [x] AddInterceptor throws if builder is not internal wrapper
## Client
- [x] Investigate moving deadline/cancellation token off HttpClientCallInvoker and into interceptor
## Compression
- [ ] Move to shared package and reference from client and server

PR #384, opened 2019-07-11
Refactor activators to be singletons, make internal, use factory

JamesNK:

* Change to singletons
* Internalize
* Use factory when creating interceptors via args


src/Grpc.AspNetCore.Server/IGrpcServiceActivator.cs
```diff
@@ -27,13 +29,26 @@ public interface IGrpcServiceActivator<TGrpcService> where TGrpcService : class
...
 
         /// <summary>
         /// Releases the specified service.
         /// </summary>
         /// <param name="service">The service to release.</param>
-        void Release(TGrpcService service);
+        void Release(Handle<TGrpcService> service);
+    }
+
+    public readonly struct Handle<T>
```
davidfowl: Rename it to something grpc specific?

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
```diff
@@ -48,18 +48,18 @@ internal class ClientStreamingServerCallHandler<TService, TRequest, TResponse> :
                 ClientStreamingServerMethod<TRequest, TResponse> resolvedInvoker = async (resolvedRequestStream, resolvedContext) =>
                 {
                     var activator = resolvedContext.GetHttpContext().RequestServices.GetRequiredService<IGrpcServiceActivator<TService>>();
```
davidfowl: Get the services once?

src/Grpc.AspNetCore.Server/Internal/CallHandlers/ClientStreamingServerCallHandler.cs
```diff
@@ -48,18 +48,18 @@ internal class ClientStreamingServerCallHandler<TService, TRequest, TResponse> :
...
-                    TService? service = null;
+                    Handle<TService> service = default;
```
davidfowl: service -> handle

JamesNK:
🆙 📅

src/Grpc.AspNetCore.Server/IGrpcInterceptorActivator.cs
```diff
@@ -32,12 +32,12 @@ public interface IGrpcInterceptorActivator
...
         /// <returns>The created interceptor.</returns>
-        Handle<Interceptor> Create(IServiceProvider serviceProvider, params object[] args);
+        GrpcActivatorHandle<Interceptor> Create(IServiceProvider serviceProvider, params object[] args);
 
         /// <summary>
         /// Releases the specified interceptor.
         /// </summary>
         /// <param name="interceptor">The interceptor to release.</param>
-        void Release(Handle<Interceptor> interceptor);
+        void Release(GrpcActivatorHandle<Interceptor> interceptor);
```
davidfowl: in*

perf/benchmarkapps/BenchmarkClient/Workers/GrpcRawUnaryWorker.cs
```diff
@@ -30,12 +30,14 @@ namespace BenchmarkClient.Workers
 {
     public class GrpcRawUnaryWorker : IWorker
     {
+        private readonly DateTime? _deadline;
         private HttpClient? _client;
 
-        public GrpcRawUnaryWorker(int id, string target)
+        public GrpcRawUnaryWorker(int id, string target, DateTime? deadline = null)
```
halter73: A quick search of the PR diff isn't showing me where the caller to GrpcRawUnaryWorker is. Did the call site not change? Is _deadline always null here?

JunTaoLuo:
What's the perf difference after switching to Singleton?

src/Grpc.AspNetCore.Server.Reflection/GrpcReflectionServiceExtensions.cs
```diff
@@ -41,9 +45,71 @@ public static IServiceCollection AddGrpcReflection(this IServiceCollection servi
...
+                        {
+                            serviceDescriptors.Add(serviceDescriptor);
+                            continue;
+                        }
+                    }
+
+                    Log.ServiceDescriptorNotResolved(logger, serviceType);
+                }
+
+                return new ReflectionServiceImpl(serviceDescriptors);
```
halter73: Any reason not to add a ReflectionServiceImpl ctor that takes an ILoggerFactory and a EndpointDataSource as parameters and does all this stuff in the ctor? Is this closing over any variables? Putting this in a ctor seems better to me than putting this logic in a service factory lambda.

src/Grpc.AspNetCore.Server/IGrpcInterceptorActivator.cs
```diff
@@ -16,6 +16,7 @@
 
 #endregion
 
+using System;
 using Grpc.Core.Interceptors;
 
 namespace Grpc.AspNetCore.Server
```
JunTaoLuo: I believe we wanted to make these activators internal. I see that it's being shared with the Reflection package though.

src/Grpc.AspNetCore.Server/IGrpcServiceActivator.cs
```diff
@@ -27,13 +29,14 @@ public interface IGrpcServiceActivator<TGrpcService> where TGrpcService : class
...
         /// <returns>The created service.</returns>
-        TGrpcService Create();
+        GrpcActivatorHandle<TGrpcService> Create(IServiceProvider serviceProvider);
 
         /// <summary>
         /// Releases the specified service.
         /// </summary>
         /// <param name="service">The service to release.</param>
-        void Release(TGrpcService service);
+        void Release(in GrpcActivatorHandle<TGrpcService> service);
```
halter73: Would `ref` be better than `in` so we can set the handle back to `default`?

I think it also makes more sense semantically, since `in` generally means that the Handle is unmodified, but really the handle isn't usable after you call this.

src/Grpc.AspNetCore.Server/InterceptorRegistration.cs
```diff
@@ -17,6 +17,7 @@
 #endregion
 
 using System;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Grpc.AspNetCore.Server
```
JunTaoLuo: Did we want to expose args? I see that as a item of the linked issue.

JunTaoLuo:
Ah I see, this is only addressing some of #383. I assume we'll do the other parts in a separate PR?

src/Grpc.AspNetCore.Server.Reflection/GrpcReflectionServiceExtensions.cs
davidfowl: Yes this is kinda ugly.

JamesNK: `ReflectionServiceImpl` is from an external package.

src/Grpc.AspNetCore.Server/InterceptorRegistration.cs
JamesNK: Future PR

src/Grpc.AspNetCore.Server/IGrpcServiceActivator.cs
JamesNK: Fair, but we're going to make this internal so API shape isn't SUPER important right now.

src/Grpc.AspNetCore.Server/IGrpcInterceptorActivator.cs
JamesNK: TODO

JamesNK: TO**DONE**

perf/benchmarkapps/BenchmarkClient/Workers/GrpcRawUnaryWorker.cs
JamesNK: It's a minor benchmark improvement I didn't want to throw away so I snuck it into this PR. It's not used yet.

JamesNK:
🆙 📅

PR #390, opened 2019-07-14
Verify AddInterceptor and EnableCallContextPropagation is used with AddGrpcClient

JamesNK:

@rynowak I went with looking at the service collection instead of an internal wrapper type over IHttpClientBuilder. This works for EnableCallContextPropagation which is in another assembly and won't be able to check for the internal type.

### DIFF
--- a/src/Grpc.AspNetCore.Server.ClientFactory/GrpcServerHttpClientBuilderExtensions.cs
+++ b/src/Grpc.AspNetCore.Server.ClientFactory/GrpcServerHttpClientBuilderExtensions.cs
@@ -43,6 +43,8 @@ namespace Microsoft.Extensions.DependencyInjection
                 throw new ArgumentNullException(nameof(builder));
             }
 
+            ValidateGrpcClient(builder);
+
             builder.Services.TryAddScoped<ContextPropagationInterceptor>();
             builder.Services.AddHttpContextAccessor();
             builder.Services.AddTransient<IConfigureOptions<GrpcClientFactoryOptions>>(services =>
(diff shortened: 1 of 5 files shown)


src/Grpc.AspNetCore.Server.ClientFactory/GrpcServerHttpClientBuilderExtensions.cs
```diff
@@ -73,5 +75,24 @@ public static IHttpClientBuilder EnableCallContextPropagation(this IHttpClientBu
 
             return builder;
         }
+
+        private static void ValidateGrpcClient(IHttpClientBuilder builder)
+        {
+            // Validate the builder is for a gRPC client
+            foreach (var service in builder.Services)
+            {
+                if (service.ServiceType == typeof(IConfigureOptions<GrpcClientFactoryOptions>))
```
rynowak: Wow, this is some fragile. This seems pretty far to go to provide a nice error message for people. 

Other *cool* ideas:
- look at the type name
- add a marker interface
- make the type public
- use conditional weak table
- use your own data structure in services to track this instead of sniffing something with another purpose

JamesNK: This feels less fragile to me. An extension method might return the builder wrapped in a type which would break validate that looks at the type name or marker interface. The services collection meanwhile will always be passed around.

GrpcClientFactoryOptions is a public type where all the gRPC configuration hangs off. It has to be there. If a major refactor ever changes that then unit tests will immediately fail. What are you worried about breaking this?

JamesNK: Discussed with Ryan offline. Neither approach is perfect. Will go with this one for now, if it becomes a problem later then we can easily remove the validation.

PR #391, opened 2019-07-14
 Move context propagation to an interceptor

JamesNK:


src/Grpc.AspNetCore.Server.ClientFactory/ContextPropagationInterceptor.cs
```diff
@@ -0,0 +1,172 @@
...
+            var options = context.Options;
+            var serverCallContext = GetServerCallContext();
+
+            // Use propagated deadline if it is smaller than the specified deadline
+            if (serverCallContext.Deadline < context.Options.Deadline.GetValueOrDefault(DateTime.MaxValue))
+            {
+                options = options.WithDeadline(serverCallContext.Deadline);
+            }
+
+            if (serverCallContext.CancellationToken.CanBeCanceled)
```
jtattermusch: nit: the token will always be cancellable (as all calls can be cancelled by the client).

JamesNK: On Kestrel someone could set the request CancellationToken to CancellationToken.None. Not likely, but a simple if check doesn't really have any downsides 😄

PR #392, opened 2019-07-15
Update grpc-dotnet commandline dependency

JamesNK:
@JunTaoLuo after I made this change I ran `dotnet run add-file` and got a null reference exception. Because there were no file arguments. Should the argument have prevented the handler being called? I'm not familiar enough to say one way or another.

### DIFF
--- a/src/dotnet-grpc/Commands/AddFileCommand.cs
+++ b/src/dotnet-grpc/Commands/AddFileCommand.cs
@@ -35,12 +35,12 @@ namespace Grpc.Dotnet.Cli.Commands
         {
             var command = new Command(
                 name: "add-file",
-                description: CoreStrings.AddFileCommandDescription,
-                argument: new Argument<string[]>
-                {
-                    Name = "files",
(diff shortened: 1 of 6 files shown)


JamesNK:
Fixed it with an arity https://github.com/grpc/grpc-dotnet/pull/392/commits/b8925f995772b610b9ea803d446eb3c559bad5c3

src/dotnet-grpc/Commands/AddFileCommand.cs
```diff
@@ -35,12 +35,13 @@ public static Command Create()
...
-                {
-                    Name = "files",
-                    Description = CoreStrings.AddFileCommandArgumentDescription,
-                });
+                description: CoreStrings.AddFileCommandDescription);
+            command.AddArgument(new Argument<string[]>
+            {
+                Name = "files",
+                Description = CoreStrings.AddFileCommandArgumentDescription,
+                Arity = ArgumentArity.OneOrMore
```
JunTaoLuo: 👍

PR #395, opened 2019-07-15
Fix contents of AuthInterceptorContext

jtattermusch:
Use the same values that C-core would produce.
The serviceUrl is used as "aud" field when constructing a JWT auth token in the auth library,
so if it's malformed, authentication will fail.

This partially answers https://github.com/grpc/grpc/pull/19512#issuecomment-511380720 (one more fix coming to grpc/grpc)

### DIFF
--- a/src/Grpc.Net.Client/Internal/GrpcCall.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCall.cs
@@ -502,16 +502,21 @@ namespace Grpc.Net.Client.Internal
 
         private static AuthInterceptorContext CreateAuthInterceptorContext(Uri baseAddress, IMethod method)
         {
-            var serviceUrl = baseAddress.OriginalString;
-            if (baseAddress.Scheme == Uri.UriSchemeHttps && serviceUrl.EndsWith(":443", StringComparison.Ordinal))
+            var authority = baseAddress.Authority;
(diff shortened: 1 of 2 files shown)


src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -500,6 +500,21 @@ private async Task ReadCredentialMetadata(HttpClient client, HttpRequestMessage
             }
         }
 
+        private static AuthInterceptorContext CreateAuthInterceptorContext(Uri baseAddress, IMethod method)
+        {
+            string serviceUrl = baseAddress.OriginalString;
```
JamesNK: var

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -500,6 +500,21 @@ private async Task ReadCredentialMetadata(HttpClient client, HttpRequestMessage
...
+            if (baseAddress.Scheme == Uri.UriSchemeHttps && serviceUrl.EndsWith(":443", StringComparison.InvariantCulture))
```
JamesNK: OrdinalCulture

jtattermusch:
Added a few test cases and updated the logic slightly. I'm going to merge now to unblock https://github.com/grpc/grpc/pull/19648 but I can address extra comments (if any) as a followup.

PR #396, opened 2019-07-16
Update proto files to have semicolon after methods

JamesNK:
Semicolon is optional but its absence breaks some syntax highlighters. Lets use it everywhere.

### DIFF
--- a/examples/Proto/aggregate.proto
+++ b/examples/Proto/aggregate.proto
@@ -20,6 +20,6 @@ import "greet.proto";
 package Aggregate;
 
 service Aggregator {
-  rpc AccumulateCount (stream Count.CounterRequest) returns (Count.CounterReply) {}
-  rpc SayHellos (Greet.HelloRequest) returns (stream Greet.HelloReply) {}
+  rpc AccumulateCount (stream Count.CounterRequest) returns (Count.CounterReply) {};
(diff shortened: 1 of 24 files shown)


JamesNK:
@jtattermusch I will leave it with you if you want to consistently have semicolons in grpc/grpc proto files.

jtattermusch:
Which syntax highlighters are broken by the absence of semicolons?

My personal opinion is that  having both `{}` and `;`  is weird because I've mostly seen  these two formats in the past:
https://github.com/grpc/grpc/blob/cbc455ecfae634a0ace6f24c156516ca10be167d/src/proto/grpc/testing/worker_service.proto#L30
or 
https://github.com/grpc/grpc/blob/cbc455ecfae634a0ace6f24c156516ca10be167d/src/proto/math/math.proto#L45
I think having either `;` or `{}` but not both feels more natural because it's a C++ inspired syntax.
That said, I don't really feel strongly about the exact style being used at all, but it would be good to know which highlighters are broken with what's one of the two most common conventions.

JamesNK:
https://github.com/highlightjs

I'm guessing this is the offending bit:
https://github.com/highlightjs/highlight.js/blob/0594c3a244adda42a21164986841c28a78963ed5/src/languages/protobuf.js#L33

JamesNK:
Now that you've pointed it out to me I agree that both looks odd. I'll remove `{}` from our proto files.

PR #398, opened 2019-07-17
Log functional tests to test output

JamesNK:
Make it easier to see what is going on in a functional test:

image

PR #402, opened 2019-07-20
 Log compression and fix compression header bug

JamesNK:


JamesNK:
Also includes https://github.com/grpc/grpc-dotnet/pull/398

src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -476,19 +476,28 @@ internal bool IsEncodingInRequestAcceptEncoding(string encoding)
                 while (true)
                 {
                     var separatorIndex = acceptEncoding.IndexOf(',');
-                    if (separatorIndex == -1)
```
davidfowl: Use the typed header? cc @Tratcher

JamesNK: The header is custom to gRPC: `grpc-accept-encoding`. However if there is a better way of reading a comma delimited list of values from a header then that could replace this code

davidfowl: Like this? https://github.com/aspnet/AspNetCore/blob/4ef204e13b88c0734e0e94a1cc4c0ef05f40849e/src/Http/Http.Abstractions/src/Extensions/HeaderDictionaryExtensions.cs#L39

JamesNK: Allocations 😲

davidfowl: I know right... 😄 We need to expose another API for the split. 

Unrelated: There's a new API designed by a corefx contributor that beautifully handles splitting things by returning an enumerator of ranges.

src/Grpc.Net.Client/Internal/StreamExtensions.cs
```diff
@@ -200,6 +198,27 @@ private static Status CreateUnknownMessageEncodingMessageStatus(string unsupport
...
+            {
+                if (string.Equals(compressionEncoding, compressionProvider.EncodingName, StringComparison.Ordinal))
+                {
+                    Log.DecompressingMessage(logger, compressionProvider.EncodingName);
+
+                    var output = new MemoryStream();
+                    var compressionStream = compressionProvider.CreateDecompressionStream(new MemoryStream(messageData));
+                    compressionStream.CopyTo(output);
+
+                    result = output.ToArray();
```
davidfowl: 😢

Issue #401, opened 2019-07-20
Client-streaming RPC backpressure is broken on Kestrel


Avilad:
Related to #104. Demo code: https://github.com/Avilad/grpc-dotnet-flow-control-test.

When a client streams many small messages to a grpc-dotnet server, and the server consumes these messages slower than they are produced, `WriteAsync()` calls on the client are blocked as expected, but for each call the server makes to `MoveNext()`, `WriteAsync()` accepts more than one call before being blocked. This (presumably) causes a buildup of messages in Kestrel's buffers, eventually leading to a "Request body too large" exception.

When the Grpc.Core server is used instead, the backpressure works as expected, and allows on average one `WriteAsync()` call for each `MoveNext()` call.

davidfowl:
@Avilad do you have a call stack?

davidfowl:
OK I took a quick look at a couple of things and there's a bunch of things to unpack:

> When a client streams many small messages to a grpc-dotnet server, and the server consumes these messages slower than they are produced, WriteAsync() calls on the client are blocked as expected, but for each call the server makes to MoveNext(), WriteAsync() accepts more than one call before being blocked.

The number of calls doesn't need to be 1:1 and it cannot exceed the agreed upon window size, that's a protocol violation. Accepting multiple write per unblocked read isn't a problem unless the HTTP/2 protocol is being violated (and I'm not convinced it is).

> This (presumably) causes a buildup of messages in Kestrel's buffers, eventually leading to a "Request body too large" exception.

No buffers are being built up. This message is happening because there's a default max body size that should be disabled for streaming requests (its just counting bytes) @JamesNK will fix that.

cc @halter73

Avilad:
At any rate, the behavior is pretty different between the gRPC core server and kestrel. And even though the number of calls doesn't need to be 1:1, in my demo code, it's almost exactly 10:1. The server reads one message from the stream every ~10ms, and the log from the client shows that it sends around 10 messages every 10ms (roughly 1/10 messages send in ~0.01s, and the remaining 9/10 send in ~0.0001s). Since the messages are of constant size, I can't see how this wouldn't result in messages building up in a buffer somewhere.

JamesNK:
I tested streaming 1 GB to the server in 64 KB chunks. Remove the default max body size fixes that error. However the server is using a lot of memory.

davidfowl:
> At any rate, the behavior is pretty different between the gRPC core server and kestrel.

[...]

PR #403, opened 2019-07-20
Remove streaming max request size

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/401

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
@@ -121,17 +121,22 @@ namespace Grpc.AspNetCore.Server.Internal.CallHandlers
                 }
                 else
                 {
-                    httpContext.Features.Set<IHttpMaxRequestBodySizeFeature>(UnlimitedHttpMaxRequestBodySizeFeature.Instance);
+                    // IsReadOnly could be true if middleware has already started reading the request body
+                    // In that case we can't disable the max request body size for the request stream
+                    Log.UnableToDisableMaxRequestBodySize(Logger);
                 }
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
```diff
@@ -103,13 +104,34 @@ static async Task AwaitHandleCall(HttpContextServerCallContext serverCallContext
...
+            var maxRequestBodySizeFeature = httpContext.Features.Get<IHttpMaxRequestBodySizeFeature>();
+            if (maxRequestBodySizeFeature != null)
+            {
+                if (!maxRequestBodySizeFeature.IsReadOnly)
+                {
+                    maxRequestBodySizeFeature.MaxRequestBodySize = null;
+                }
+                else
+                {
+                    httpContext.Features.Set<IHttpMaxRequestBodySizeFeature>(UnlimitedHttpMaxRequestBodySizeFeature.Instance);
```
davidfowl: You don't need to do this.

JamesNK: Why not? Won't MaxRequestBodySize throw an error is IsReadOnly is true.

JamesNK: I re-read the IsReadOnly docs. Does it mean that is too late in the request for changing the body size to have any effect?

davidfowl: > Does it mean that is too late in the request for changing the body size to have any effect?

Right.

davidfowl:
Nice!

PR #406, opened 2019-07-22
Fix incorrectly setting examined when message is read

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/401

When a complete message was parsed gRPC was incorrectly requesting more data. When given a constant stream of data from the client the server would eagerly grab it, creating a build up on data on the pipe.

Only request more data when a message is not finished.

### DIFF
--- a/test/Grpc.AspNetCore.Server.Tests/PipeExtensionsTests.cs
+++ b/test/Grpc.AspNetCore.Server.Tests/PipeExtensionsTests.cs
@@ -23,6 +23,7 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Google.Protobuf.WellKnownTypes;
 using Grpc.AspNetCore.Server.Compression;
 using Grpc.AspNetCore.Server.Internal;
 using Grpc.Core;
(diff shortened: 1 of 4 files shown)


JamesNK:
cc @Avilad

test/Grpc.AspNetCore.Server.Tests/PipeExtensionsTests.cs
```diff
@@ -640,5 +662,57 @@ public Stream CreateDecompressionStream(Stream stream)
...
+
+            public TestPipeReader(PipeReader pipeReader)
+            {
+                _pipeReader = pipeReader;
+            }
+
+            public override void AdvanceTo(SequencePosition consumed)
+            {
+                Consumed = consumed;
+                Examined = null;
```
davidfowl: This more asserts the calls made to the PipeReader than the behavior itself. That should be fine in this case though...

davidfowl: Maybe it would be better just to set a bool that determines when AdvanceTo overload was called. Though I don't like this test, we shouldn't be storing sequence positions outside of AdvanceTo (they basically go away after this call).

JamesNK: Updated to capture the underlying position index before AdvanceTo is called.

JamesNK:
> So fast 😱

Fixed quickly because you provided a good description and reproduction of the issue. Thanks for your help. A potentially tricky bug got fixed very quickly 😄

Issue #393, opened 2019-07-15
dotnet-grpc will silently ignore some add-file paths


JamesNK:
`dotnet run add-file s`

The path `s` is not resolved to anything. I would expect that the tool would notify the user that the path didn't do anything.

PR #412, opened 2019-07-25
Warn if file reference not resolved

JunTaoLuo:
Addresses https://github.com/grpc/grpc-dotnet/issues/393


src/dotnet-grpc/Properties/CoreStrings.resx
```diff
@@ -207,4 +207,7 @@
   <data name="ErrorOutputMustBeFilePath" xml:space="preserve">
     <value>Output path `{0}` is invalid. The path cannot be a directory path and must be a file path.</value>
   </data>
+  <data name="LogWarningNoReferenceResolved" xml:space="preserve">
+    <value>No file reference was found for pattern `{0}`.</value>
```
JamesNK: What about:

> No file(s) found matching file argument `{0}`.

I prefer this because pattern isn't used in any other messages. Also this message is more clear about what the pattern was used for.

PR #413, opened 2019-07-26
Add functional tests for hangs in client

JamesNK:
WIP because the SDK containing the fix has a problem. Will update when the SDK with fix is fixed.


JamesNK:
Replaced by https://github.com/grpc/grpc-dotnet/pull/413

One of the unit tests was removed because its fix isn't in the SDK yet.

PR #418, opened 2019-07-30
Flush request body stream at start of client and duplex streaming calls

JamesNK:
Addresses part of https://github.com/dotnet/corefx/issues/39586

### DIFF
--- a/build_and_test.sh
+++ b/build_and_test.sh
@@ -19,7 +19,7 @@ source activate.sh
 
 echo "Building solution"
 
-dotnet build
+dotnet build -c Release
 
 echo "Testing solution"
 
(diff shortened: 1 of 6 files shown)


src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -609,10 +609,17 @@ private static AuthInterceptorContext CreateAuthInterceptorContext(Uri baseAddre
             _writeCompleteTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
 
             message.Content = new PushStreamContent(
-                (stream) =>
+                async stream =>
                 {
+                    // Immediately flush request stream to send headers
+                    // https://github.com/dotnet/corefx/issues/39586#issuecomment-516210081
+                    await stream.FlushAsync().ConfigureAwait(false);
```
davidfowl: Is this why it hangs?!

JamesNK: If there is no HTTP traffic being sent from the HttpClient at that time, yes.

JamesNK:
I finally figure out what was causing the tests to hang. Something about dotnet test writing to the console causes it.

I updated the build script to write to a file and then write the file to console.

build_and_test.sh
```diff
@@ -19,17 +19,25 @@ source activate.sh
...
 echo "Testing solution"
 
 test_projects=( $( ls test/**/*Tests.csproj ) )
 
 for test_project in "${test_projects[@]}"
 do
-    # Capturing test diagnostic logs because of hanging build
-    # https://github.com/grpc/grpc-dotnet/pull/363
-    dotnet test $test_project --no-build --diag:artifacts/${test_project##*/}.log.txt
+    # "dotnet test" is hanging when it writes to console for an unknown reason
```
JunTaoLuo: Hmm weird but okay.

jtattermusch: do we believe this because of a bug in the preview version of .NET core or because of a bug in grpc-dotnet?

JamesNK: I believe it is an issue in .NET Core test. Issue here - https://github.com/microsoft/vstest/issues/2080

I've added a link to it in the build script

PR #423, opened 2019-07-31
Fix hang in client writer when call is started canceled

JamesNK:
Race condition in client/duplex streaming when call is passed a canceled cancellation token.


JunTaoLuo:
Ugh... hanging template tests. Retrying.

Issue #420, opened 2019-07-30
Move configuration from HttpClientCallInvoker to Channel implementation


JamesNK:
ChannelBase is now in Grpc.Core.Api. We can move configuration that belongs on the channel to a grpc-dotnet implementation of it.

PR #428, opened 2019-08-02
Add Channel implementation and move config from invoker to channel

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/420


src/Grpc.Net.Client/ChannelBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            ChannelBuilder channelBuilder = new ChannelBuilder();
+            channelBuilder.ConfigureOptions(o =>
+            {
+                o.HttpClient = httpClient;
+            });
+            return channelBuilder;
+        }
+
+        public static ChannelBuilder ForOptions(GrpcChannelOptions channelOptions)
```
JamesNK: I don't like this. It is used in HttpClientFactory to "apply" settings set with it to ChannelBuilder.

src/Grpc.Net.Client/ChannelBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+        /// </summary>
+        public int? SendMaxMessageSize { get; set; }
+
+        /// <summary>
+        /// Gets or sets the maximum message size in bytes that can be received by the client.
+        /// </summary>
+        public int? ReceiveMaxMessageSize { get; set; }
+    }
+
+    public class GrpcChannel : ChannelBase
```
JamesNK: GrpcChannel doesn't add anything public that ChannelBase doesn't already have. It is public in case it might need new API in the future.

JamesNK: Should be sealed.

src/Grpc.Net.Client/ChannelBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Net.Http;
+using Grpc.Core;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+
+namespace Grpc.Net.Client
+{
+    public class ChannelBuilder
```
JamesNK: Should be sealed.

src/Grpc.Net.Client/GrpcClient.cs
```diff
@@ -39,7 +39,33 @@ public static class GrpcClient
         /// <returns>A gRPC client.</returns>
         public static TClient Create<TClient>(HttpClient httpClient, ILoggerFactory? loggerFactory = null) where TClient : ClientBase
```
JamesNK: Remove?

src/Grpc.Net.Client/GrpcClient.cs
```diff
@@ -39,7 +39,33 @@ public static class GrpcClient
...
+            return CreateCore<TClient>(new HttpClientCallInvoker(httpClient, loggerFactory));
+        }
+
+        /// <summary>
+        /// Creates a gRPC client using the specified <see cref="ChannelBase"/>.
+        /// </summary>
+        /// <typeparam name="TClient">The type of the gRPC client. This type will typically be defined using generated code from a *.proto file.</typeparam>
+        /// <param name="channel">The <see cref="ChannelBase"/>.</param>
+        /// <returns>A gRPC client.</returns>
+        public static TClient Create<TClient>(ChannelBase channel) where TClient : ClientBase
```
JamesNK: Consider removing?

Codegen clients will have a constructor that takes ChannelBase. This method is slightly useful as it provides a simple fast way to use a generic argument to create a strongly typed client from a channel.

src/Grpc.Net.ClientFactory/Internal/GrpcHttpClientFactory.cs
```diff
@@ -35,13 +35,15 @@ internal class GrpcHttpClientFactory<TClient> : INamedTypedHttpClientFactory<TCl
         private readonly Cache _cache;
         private readonly IServiceProvider _services;
         private readonly ILoggerFactory _loggerFactory;
-        private readonly IOptionsMonitor<GrpcClientFactoryOptions> _optionsMonitor;
+        private readonly IOptionsMonitor<GrpcClientFactoryOptions> _clientFactoryOptionsMonitor;
+        private readonly IOptionsMonitor<GrpcChannelOptions> _channelOptionsMonitor;
```
JamesNK: `GrpcChannelOptions` are currently separate from `GrpcClientFactoryOptions`. Could refactor to make `ChannelOptions` a property on `GrpcClientFactoryOptions`.

src/Grpc.Net.ClientFactory/Internal/GrpcHttpClientFactory.cs
```diff
@@ -76,17 +84,22 @@ public TClient CreateClient(HttpClient httpClient, string name)
...
+            var channelOptions = _channelOptionsMonitor.Get(name);
 
-            var options = _optionsMonitor.Get(name);
-            for (var i = 0; i < options.CallInvokerActions.Count; i++)
-            {
-                options.CallInvokerActions[i](httpClientCallInvoker);
-            }
+            channelOptions.HttpClient = httpClient;
+
+            var channelBuilder = ChannelBuilder.ForOptions(channelOptions);
```
JamesNK: Not a fan of this.

src/Grpc.Net.Client/ChannelBuilder.cs
```diff
@@ -0,0 +1,152 @@
...
+#endregion
+
+using System;
+using System.Diagnostics;
+using System.Net.Http;
+using Grpc.Core;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
JunTaoLuo: Why is this disabled? Also one more spot.

src/Grpc.Net.Client/ChannelBuilder.cs
```diff
@@ -0,0 +1,152 @@
...
+        private ChannelBuilder() : this(new GrpcChannelOptions())
+        {
+        }
+
+        private ChannelBuilder(GrpcChannelOptions options)
+        {
+            _options = options;
+        }
+
+        public static ChannelBuilder ForUrl(Uri uri)
```
JunTaoLuo: I wonder if we want something we do for WebHost along the lines of WebHost.CreateDefaultBuilder(...). I can see GrpcChannel.CreateDefaultBuilder(...) or GrpcChannel.CreateBuilder(...).

src/Grpc.Net.Client/GrpcClient.cs
JunTaoLuo: This would be slightly more concise right? I don't mind this syntax sugar.

src/Grpc.Net.Client/GrpcClient.cs
JunTaoLuo: This one seems a little more strenous though. I don't think this is simplifying anything?

`GrpcClient.Create<GreeterClient>(channelBase)` vs `new GreeterClient(channelBase)` doesn't seem like a significant improvement.

src/Grpc.Net.ClientFactory/GrpcHttpClientBuilderExtensions.cs
```diff
@@ -28,6 +29,55 @@ namespace Microsoft.Extensions.DependencyInjection
...
+            if (configureChannel == null)
+            {
+                throw new ArgumentNullException(nameof(configureChannel));
+            }
+
+            ValidateGrpcClient(builder);
+
+            builder.Services.AddTransient<IConfigureOptions<GrpcClientFactoryOptions>>(services =>
+            {
+                return new ConfigureNamedOptions<GrpcClientFactoryOptions>(builder.Name, (options) =>
```
JunTaoLuo: nit: no need for parenthesis on options.

src/Grpc.Net.Client/ChannelBuilder.cs
JamesNK: I don't want to write docs for a design that isn't final.

src/Grpc.Net.Client/GrpcClient.cs
JunTaoLuo: Fair point in terms of simplicity. Let's remove

src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -0,0 +1,129 @@
...
+        internal List<CallCredentials>? CallCredentials { get; }
+
+        // Timing related options that are set in unit tests
+        internal ISystemClock Clock = SystemClock.Instance;
+        internal bool DisableClientDeadlineTimer { get; set; }
+
+        internal GrpcChannel(GrpcChannelOptions channelOptions, GrpcTransportOptions transportOptions, ILoggerFactory loggerFactory) : base(transportOptions.Target)
+        {
+            var httpClient = (transportOptions as HttpClientTransportOptions)?.HttpClient;
+            Debug.Assert(httpClient != null, "HttpClientTransportOptions is the only options and have already used HttpClient on options.");
```
JunTaoLuo: This error message isn't very clear.

JamesNK:
@jtattermusch This is a major refactor of how clients are created using a new ChannelBuilder type. It would be great if you could take a look.

I'd like to merge it now so people can try it out in nightly builds. Leave feedback if this is merged and I can do a followup PR.

src/Grpc.Net.Client/GrpcChannelOptions.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// An options class for configuring a <see cref="GrpcChannel"/>.
+    /// </summary>
+    public sealed class GrpcChannelOptions
+    {
+        /// <summary>
+        /// Gets or sets the transport options for the channel. Transport options are required
+        /// to successfully create a channel.
+        /// </summary>
+        public GrpcTransportOptions? TransportOptions { get; set; }
```
jtattermusch: can the user currently use the TransportOptions to actually configure something? It looks like the transport options are only set when the channel builder is created and I don't see a situation where they'd need to be updated.
Perhaps this doesn't have to be a public property?

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -531,14 +531,26 @@ private async Task SendAsync(HttpClient client, HttpRequestMessage request)
...
             {
                 // In C-Core the call credential auth metadata is only applied if the channel is secure
                 // The equivalent in grpc-dotnet is only applying metadata if HttpClient is using TLS
                 // HttpClient scheme will be HTTP if it is using H2C (HTTP2 without TLS)
                 if (client.BaseAddress.Scheme == Uri.UriSchemeHttps)
                 {
-                    await ReadCredentialMetadata(client, request, Options.Credentials).ConfigureAwait(false);
+                    var configurator = new DefaultCallCredentialsConfigurator();
+
+                    if (Options.Credentials != null)
```
jtattermusch: if both callCredentials are compositeChannel credentials are provided, will the order of invocation match to Grpc.Core behavior? If we checked that, it might be good to add a comment that explicitly mentions that.

src/Grpc.Net.Client/ChannelBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+using Grpc.Core;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Grpc.Net.Client
+{
+    /// <summary>
+    /// A builder for configuring <see cref="GrpcChannel"/> instances.
+    /// </summary>
+    public sealed class ChannelBuilder
```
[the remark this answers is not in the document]
JamesNK: I'm not sure if ASP.NET Core style configuration adds much. DI allows multiple lambdas to be registered which doesn't apply in this situation.

If we didn't have a builder then it would be a choice between something like:

`GrpcChannel.Create(new GrpcChannelOptions { })` vs `GrpcChannel.Create(options => { })`

The nice thing about the `ChannelBuilder.ForHttpClient(client)` is it forces the transport to be set and provides a nice helper method to do so.

The alternative is something like:

```csharp
GrpcChannel.Create(new GrpcChannelOptions
{
    TransportOptions = new HttpClientTransportOptions { HttpClient = httpClient }
});
```

Or the builder could be the argument to create and it could have some helper methods for setting some values:

```csharp
GrpcChannel.Create(builder =>
{
    builder.SetHttpClientTransport(httpClient);
    builder.SendMaxMessage = 100;
});
```

The downside is if/when we support multiple transports then people could incorrectly do something like:

```csharp
GrpcChannel.Create(builder =>
{
    // Which is used?
    builder.SetHttpClientTransport(httpClient);
    builder.SetSocketTransport(socket);
    builder.SendMaxMessage = 100;
});
```

src/Grpc.Net.Client/ChannelBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+            configure(_options);
+            return this;
+        }
+
+        /// <summary>
+        /// Sets the maximum message size in bytes that can be sent from the client.
+        /// </summary>
+        /// <param name="sendMaxMessageSize">The maximum message size in bytes.</param>
+        /// <returns>The same instance of the <see cref="ChannelBuilder"/> for chaining.</returns>
+        public ChannelBuilder SetSendMaxMessageSize(int? sendMaxMessageSize)
```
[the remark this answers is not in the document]
JamesNK: It is on both because the builder just sets values on the options.

src/Grpc.Net.Client/GrpcChannelOptions.cs
JamesNK: They could do this: `TransportOptions = new HttpClientTransportOptions { HttpClient = httpClient }`

src/Grpc.Net.Client/Internal/GrpcCall.cs
JamesNK: I didn't look into it. Do you know the order off the top of your head?

jtattermusch: I'd expect that the call credentials from composite channel credentials get invoked first, but I don't know for sure, we'd need to try.

Issue #430, opened 2019-08-04
Validate names and types with gRPC + HttpClientFactory


JamesNK:
See https://github.com/aspnet/Extensions/pull/1563

PR #434, opened 2019-08-05
Add name validation when configuring gRPC clients with factory

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/430


src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
```diff
@@ -23,6 +23,8 @@
 using Grpc.Core;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Options;
+using System.Linq;
```
JunTaoLuo: nit: sort

Issue #433, opened 2019-08-04
Move compression to common package and add properties on channel


JamesNK:
From https://github.com/grpc/grpc-dotnet/issues/383

PR #436, opened 2019-08-05
Move compression provider to new common package

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/433


JamesNK:
@jtattermusch Is the new package setup correctly?

PR #438, opened 2019-08-07
Add ReadAllAsync extension method to common

JamesNK:
Adds ReadAllAsync to Common package.

If we like this then I'll add additional unit tests.

### DIFF
--- a/src/Grpc.Net.Common/AsyncStreamReaderExtensions.cs
+++ b/src/Grpc.Net.Common/AsyncStreamReaderExtensions.cs
@@ -1,0 +1,50 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


src/Grpc.Net.Common/AsyncStreamReaderExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Grpc.Core
+{
+    /// <summary>
+    /// Extension methods for <see cref="IAsyncStreamReader{T}"/>.
+    /// </summary>
+    public static class AsyncStreamReaderExtensions
```
JamesNK: Could be moved to Grpc.Core.Api in the future and type forwarded.

JamesNK:
Rebased to include only ReadAllAsync changes.

JamesNK:
Added tests. @jtattermusch could you take a look. It's ready to merge.

PR #442, opened 2019-08-10
Update SDK and enable fixed tests

JamesNK:
SDK contains HTTP/2 fix for some disabled tests.


test/FunctionalTests/Server/ClientStreamingMethodTests.cs
```diff
@@ -275,8 +275,6 @@ async Task<CounterReply> AccumulateCount(IAsyncStreamReader<CounterRequest> requ
             var responseTask = Fixture.Client.SendAsync(httpRequest, HttpCompletionOption.ResponseHeadersRead);
 
             // Assert
-            Assert.IsFalse(responseTask.IsCompleted, "Server should wait for client to finish streaming");
```
JunTaoLuo: Why is this removed?

JamesNK: It's a bad assert. The client stream is being sent immediately.

Issue #353, opened 2019-06-25
Service reference CLI improvements


JunTaoLuo:
- [x] Use https://webpifeed.blob.core.windows.net/webpifeed/Partners/PackageVersions.json as a source of truth when adding package references.
- [x] Add a command to enumerate the list of Protobuf references in the project.
- [x] Conditionally pick Grpc.Net.ClientFactory and Grpc.AspNetCore.Server.ClientFactory

https://github.com/grpc/grpc-dotnet/pull/271#issuecomment-505354900

PR #443, opened 2019-08-10
ServiceReference CLI improvements

JunTaoLuo:
Addresses #353 WIP

### DIFF
--- a/src/dotnet-grpc/Commands/CommandBase.cs
+++ b/src/dotnet-grpc/Commands/CommandBase.cs
@@ -70,6 +70,8 @@ namespace Grpc.Dotnet.Cli.Commands
 
         internal IConsole Console { get; set; }
         internal Project Project { get; set; }
+        private bool IsUsingWebSdk => Project.AllEvaluatedProperties.Any(p => string.Equals(WebSDKProperty, p.Name, StringComparison.OrdinalIgnoreCase)
+                    && string.Equals("true", p.UnevaluatedValue, StringComparison.OrdinalIgnoreCase));
 
         public Services ResolveServices(Services services)
         {
@@ -80,8 +82,7 @@ namespace Grpc.Dotnet.Cli.Commands
             }
 
             // If UsingMicrosoftNETSdkWeb is true, generate Client and Server services
-            if (Project.AllEvaluatedProperties.Any(p => string.Equals(WebSDKProperty, p.Name, StringComparison.OrdinalIgnoreCase)
-                    && string.Equals("true", p.UnevaluatedValue, StringComparison.OrdinalIgnoreCase)))
+            if (IsUsingWebSdk)
             {
                 return Services.Both;
             }
@@ -98,7 +99,10 @@ namespace Grpc.Dotnet.Cli.Commands
             {
                 if (dependency.ApplicableServices.Split(';').Any(s => string.Equals(s, services.ToString(), StringComparison.OrdinalIgnoreCase)))
                 {
-                    AddNugetPackage(dependency.Name, dependency.Version, dependency.PrivateAssets);
+                    if (dependency.ApplicableToWeb == null || string.Equals(dependency.ApplicableToWeb, IsUsingWebSdk.ToString(), StringComparison.OrdinalIgnoreCase))
+                    {
+                        AddNugetPackage(dependency.Name, dependency.Version, dependency.PrivateAssets);
+                    }
                 }
             }
         }
--- a/src/dotnet-grpc/Commands/CommandBase.cs
+++ b/src/dotnet-grpc/Commands/CommandBase.cs
@@ -147,6 +147,8 @@ namespace Grpc.Dotnet.Cli.Commands
             }
             catch
             {
+                // TODO (johluo): Consider logging a message indicating what went wrong and actions, if any, to be taken to resolve possible issues.
+                // Currently not logging anything since the fwlink is not published yet.
                 return null;
             }
         }
--- a/src/dotnet-grpc/Commands/CommandBase.cs
+++ b/src/dotnet-grpc/Commands/CommandBase.cs
@@ -72,11 +72,8 @@ namespace Grpc.Dotnet.Cli.Commands
 
         internal IConsole Console { get; set; }
         internal Project Project { get; set; }
-        private string UsingWebSdk => Project.AllEvaluatedProperties
-            .SingleOrDefault(p =>
-                string.Equals(UsingWebSDKPropertyName, p.Name, StringComparison.OrdinalIgnoreCase))
-            ?.UnevaluatedValue
-            ?? "false";
+        private bool IsUsingWebSdk => Project.AllEvaluatedProperties.Any(p => string.Equals(UsingWebSDKPropertyName, p.Name, StringComparison.OrdinalIgnoreCase)
+            && string.Equals("true", p.UnevaluatedValue, StringComparison.OrdinalIgnoreCase));
 
         public Services ResolveServices(Services services)
         {
@@ -87,7 +84,7 @@ namespace Grpc.Dotnet.Cli.Commands
             }
 
             // If UsingMicrosoftNETSdkWeb is true, generate Client and Server services
-            if (string.Equals("true", UsingWebSdk, StringComparison.OrdinalIgnoreCase))
+            if (IsUsingWebSdk)
             {
                 return Services.Both;
             }
@@ -108,7 +105,7 @@ namespace Grpc.Dotnet.Cli.Commands
                 if (dependency.ApplicableServices.Split(';').Any(s => string.Equals(s, services.ToString(), StringComparison.OrdinalIgnoreCase)))
                 {
                     // Check if the dependency is applicable to this SDK type
-                    if (dependency.ApplicableToWeb == null || string.Equals(dependency.ApplicableToWeb, UsingWebSdk, StringComparison.OrdinalIgnoreCase))
+                    if (dependency.ApplicableToWeb == null || string.Equals(dependency.ApplicableToWeb, IsUsingWebSdk.ToString(), StringComparison.OrdinalIgnoreCase))
                     {
                         // Use the version specified from the remote file before falling back the packaged versions
                         var packageVersion = packageVersions?.GetValueOrDefault(dependency.Name) ?? dependency.Version;
@@ -139,7 +136,7 @@ namespace Grpc.Dotnet.Cli.Commands
                 using var packageVersionStream = await _httpClient.GetStreamAsync(PackageVersionUrl);
                 using var packageVersionDocument = await JsonDocument.ParseAsync(packageVersionStream);
                 var packageVersionsElement = packageVersionDocument.RootElement.GetProperty("Packages");
-                var packageVersionsDictionary = new Dictionary<string, string>();
+                var packageVersionsDictionary = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
                 foreach (var packageVersion in packageVersionsElement.EnumerateObject())
                 {
--- a/src/dotnet-grpc/Commands/CommandBase.cs
+++ b/src/dotnet-grpc/Commands/CommandBase.cs
@@ -26,7 +26,6 @@ using System.Linq;
 using System.Net.Http;
 using System.Reflection;
 using System.Security.Cryptography;
-using System.Text;
 using System.Text.Json;
 using System.Threading.Tasks;
 using Grpc.Dotnet.Cli.Internal;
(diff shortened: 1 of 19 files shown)


src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -70,6 +72,11 @@ internal CommandBase(IConsole console, Project project, HttpClient httpClient)
 
         internal IConsole Console { get; set; }
         internal Project Project { get; set; }
+        private string UsingWebSdk => Project.AllEvaluatedProperties
```
JunTaoLuo: Eh... maybe a bool is better.

JamesNK: Change to method that returns bool: IsUsingWebSdk

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -90,19 +96,64 @@ public Services ResolveServices(Services services)
...
+                foreach (var packageVersion in packageVersionsElement.EnumerateObject())
+                {
+                    packageVersionsDictionary[packageVersion.Name] = packageVersion.Value.GetString();
+                }
+
+                return packageVersionsDictionary;
+            }
+            catch
+            {
+                return null;
```
JamesNK: Write error. Explain consequences of error (an older version could be referenced?) and what the user should do (update package version manually afterwards?)

src/dotnet-grpc/Commands/ListCommand.cs
```diff
@@ -0,0 +1,102 @@
...
+                        console.LogError(e);
+
+                        return -1;
+                    }
+                });
+
+            return command;
+        }
+
+        public  void List()
```
JamesNK:
```suggestion
        public void List()
```

image

// @rynowak

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -90,19 +96,64 @@ public Services ResolveServices(Services services)
...
+                "Microsoft.Extensions.ApiDescription.Client": "0.3.0-preview7.19365.7",
+                "Newtonsoft.Json": "12.0.2"
+              }
+            }*/
+            try
+            {
+                using var packageVersionStream = await _httpClient.GetStreamAsync(PackageVersionUrl);
+                using var packageVersionDocument = await JsonDocument.ParseAsync(packageVersionStream);
+                var packageVersionsElement = packageVersionDocument.RootElement.GetProperty("Packages");
+                var packageVersionsDictionary = new Dictionary<string, string>();
```
JamesNK: StringComparer.Ordinal

src/dotnet-grpc/Commands/ListCommand.cs
```diff
@@ -0,0 +1,102 @@
...
+                table.AddColumn(r => r.GetMetadataValue(AccessElement), "Access");
+            }
+
+            if (protobufElements.Any(r => !string.IsNullOrEmpty(r.GetMetadataValue(AdditionalImportDirsElement))))
+            {
+                table.AddColumn(r => r.GetMetadataValue(AdditionalImportDirsElement), "Additional Imports");
+            }
+
+            var screen = new ScreenView(consoleRenderer, Console) { Child = table };
+            screen.Render(new Region(0, 0, System.Console.WindowWidth, System.Console.WindowWidth));
```
JamesNK: Fancy. Have you tested out what happens when file names/URLs are very long? Does it still look ok?

test/dotnet-grpc.Tests/TestBase.cs
```diff
@@ -16,21 +16,27 @@
...
 
 namespace Grpc.Dotnet.Cli.Tests
 {
     public class TestBase
     {
+
         internal static readonly string SourceUrl = "https://contoso.com/greet.proto";
 
-        internal static readonly string ProtoContent = @"// Copyright 2019 The gRPC Authors
+        internal static Dictionary<string, string> ContentDictionary = new Dictionary<string, string>()
```
JamesNK: I'm not a fan of a mutable static dictionary that is shared between tests. Keep tests isolated.

Create and configure TestMessageHandler setup as needed inside unit tests that use it.

test/dotnet-grpc.Tests/ListCommandTests.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.IO;
+using Grpc.Dotnet.Cli.Commands;
+using NUnit.Framework;
+
+namespace Grpc.Dotnet.Cli.Tests
+{
+    [TestFixture]
+    public class ListCommandTests : TestBase
+    {
+        [NonParallelizable]
```
JamesNK: Why is this test not parallelizable?

test/dotnet-grpc.Tests/ListCommandTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            //</Protobuf>
+            Assert.AreEqual(new string[] { "Proto/a.proto", "Both", "https://contoso.com/greet.proto" }, lines[1].Split(' ', StringSplitOptions.RemoveEmptyEntries));
+
+            // Third line is the reference to
+            //<Protobuf Include="Proto/b.proto" Access="Internal"/>
+            Assert.AreEqual(new string[] { "Proto/b.proto", "Both", "Internal" }, lines[2].Split(' ', StringSplitOptions.RemoveEmptyEntries));
+
+            // Cleanup
+            Directory.SetCurrentDirectory(currentDir);
+            Directory.Delete(tempDir, true);
```
JamesNK: Put this in a finally.

test/dotnet-grpc.Tests/CommandBaseTests.cs
```diff
@@ -110,6 +130,48 @@ public void EnsureNugetPackages_DoesNotOverwriteExistingPackageReferences()
             Assert.NotNull(packageRefs.SingleOrDefault(r => r.UnevaluatedInclude == "Grpc.Tools" && !r.HasMetadata(CommandBase.PrivateAssetsElement)));
         }
 
+        [Test]
+        public async Task EnsureNugetPackages_UsesVersionsFromRemoteFile_IfAvailable()
```
JamesNK: Add failure tests

src/dotnet-grpc/Commands/ListCommand.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        public  void List()
+        {
+            var consoleRenderer = new ConsoleRenderer(Console);
+            var protobufElements = Project.GetItems(ProtobufElement).ToList();
+            var table = new TableView<ProjectItem> { Items = protobufElements};
+
+            // Required columns (always displayed)
+            table.AddColumn(r => r.UnevaluatedInclude, "Protobuf Reference");
```
JamesNK: Should this text, and other table titles, come from a resource file?

src/dotnet-grpc/Commands/CommandBase.cs
JunTaoLuo: Yea I flip flopped on this a few times. It's odd that you'd need to convert the string to a bool back to a string here: https://github.com/grpc/grpc-dotnet/pull/443/files#diff-d2f510a9a2d14310feef2ce1dd1284e8R111. But bool looks cleaner by a little.

JunTaoLuo: Also, why a method and not a property?

src/dotnet-grpc/Commands/CommandBase.cs
JunTaoLuo: I thought about that, but right now, that fwlink remains unpublished so adding an error/warning message seems premature since it will always show up.

src/dotnet-grpc/Commands/ListCommand.cs
JunTaoLuo: image

src/dotnet-grpc/Commands/CommandBase.cs
JunTaoLuo: IgnoreCase?

test/dotnet-grpc.Tests/CommandBaseTests.cs
JunTaoLuo: Every other test is a failure test. All other EnsureNugetPackages test gets an empty response when hitting the fwlink and falls back to the bundled versions. Also, this test specifically omits a version for Grpc.Net.ClientFactory for testing fallback logic.

src/dotnet-grpc/Commands/ListCommand.cs
JunTaoLuo: Not great. This is mostly dependent on the TableView API which allows size to content (default), fixed, and star options. The other option is to use star and specify the weight (ratio based, e.g. 1:0.5:1) but the size is no longer contextual to the content and will cut off the content if it's too long:
```Protobuf Reference                                                                         Service Type                                                                               Source URL                                                                                 Access
..\Proto\aggregate.proto                                                                   Server                                                                                     Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed gravida odio nisi, sed laoreetInternal
..\Proto\certify.proto                                                                     Server                                                                                                                                                                                Public
..\Proto\count.proto                                                                       Both                                                                                                                                                                                  Public
..\Proto\greet.proto                                                                       Both                                                                                                                                                                                  Public
..\Proto\mail.proto                                                                        Server                                                                                                                                                                                Public
..\Proto\ticket.proto                                                                      Server                                                                                                                                                                                Public
```

Alternative is to calculate the column ourselves  and use ellipses when content is too long (which I didn't want to do). Could also consider asking for new features to make this better but I'm not sure about the turnaround time.

src/dotnet-grpc/Commands/CommandBase.cs
JamesNK: If a property is heavy then a method indicates to callers that if you need to use it multiple times then the result should be set to a local variable

A property that uses LINQ feels heavy to me.

It's not that important.

src/dotnet-grpc/Commands/CommandBase.cs
JamesNK: Sure, I don't believe package IDs are case-sensitive

test/dotnet-grpc.Tests/CommandBaseTests.cs
```diff
@@ -450,4 +470,57 @@ public void IsUrl_ChecksUrlValidity(string url, bool isUrl)
         private Project CreateIsolatedProject(string path)
             => Project.FromFile(path, new ProjectOptions { ProjectCollection = new ProjectCollection() });
     }
+
+    [TestFixture]
+    public class CommandBaseRemoteFileTests : TestBase
+    {
+        [OneTimeSetUp]
+        public void InitializeRemoteFile()
```
JamesNK: Why have any of this as fields on the test class? Just have a helper method that someone can use to create an HttpClient with the expected content. That is what all the gRPC client tests do.

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -90,19 +93,64 @@ public Services ResolveServices(Services services)
...
             Debug.Assert(services != Services.Default);
 
             foreach (var dependency in GetType().Assembly.GetCustomAttributes<GrpcDependencyAttribute>())
             {
+                // Check if the dependency is applicable for this service type
                 if (dependency.ApplicableServices.Split(';').Any(s => string.Equals(s, services.ToString(), StringComparison.OrdinalIgnoreCase)))
                 {
-                    AddNugetPackage(dependency.Name, dependency.Version, dependency.PrivateAssets);
+                    // Check if the dependency is applicable to this SDK type
+                    if (dependency.ApplicableToWeb == null || string.Equals(dependency.ApplicableToWeb, IsUsingWebSdk.ToString(), StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
JunTaoLuo: I don't like it but I feel there isn't much difference either way. Comparing strings vs parsing as bool and comparing bools.

src/dotnet-grpc/Commands/CommandBase.cs
JamesNK: And create an issue for adding an error message when this is fully functional.

test/dotnet-grpc.Tests/CommandBaseTests.cs
JunTaoLuo: I was trying to minimize code duplication.

src/dotnet-grpc/Commands/CommandBase.cs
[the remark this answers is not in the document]
JamesNK: `string -> bool -> compare bools` feels cleaner than `bool -> string -> compare strings`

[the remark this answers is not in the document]
JunTaoLuo: Which is why I also tried strings everywhere in the original version of this PR. I feel like this is a minutiae that doesn't matter too much.

src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -90,19 +93,66 @@ public Services ResolveServices(Services services)
...
+                var packageVersionsDictionary = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+                foreach (var packageVersion in packageVersionsElement.EnumerateObject())
+                {
+                    packageVersionsDictionary[packageVersion.Name] = packageVersion.Value.GetString();
+                }
+
+                return packageVersionsDictionary;
+            }
+            catch
```
ryanbrandenburg: Should probably limit the type of exception here to `JSONParseException` or whatever it's called if the intent is to fail gracefully if the link doesn't point to json.

JunTaoLuo:
cc @shirhatti to take a final look before I merge.

PR #444, opened 2019-08-10
Clean up samples

JamesNK:


examples/Clients/Aggregator/Program.cs
```diff
@@ -32,11 +32,9 @@ class Program
     {
         static Random RNG = new Random();
 
-        private const string Address = "localhost:50051";
-
         static async Task Main(string[] args)
         {
-            var httpClient = ClientResources.CreateHttpClient(Address);
+            var httpClient = new HttpClient { BaseAddress = new Uri("https://localhost:50051") };
```
JunTaoLuo: nit: could consider moving to 5001 instead of 50051 to better align with our templates.

PR #445, opened 2019-08-10
Fix EnableCallContextPropagation scoping issue

JamesNK:


src/Grpc.AspNetCore.Server.ClientFactory/ContextPropagationInterceptor.cs
```diff
@@ -24,10 +24,14 @@
 
 namespace Grpc.AspNetCore.Server.ClientFactory
 {
+    /// <summary>
+    /// Interceptor that will set the current request's cancellation token and deadline onto CallOptions.
+    /// This interceptor is registered with a singleton lifetime. IHttpContextAccessor picks the request up base
```
JunTaoLuo:
```suggestion
    /// This interceptor is registered with a singleton lifetime. IHttpContextAccessor picks up the request base
```

PR #446, opened 2019-08-11
Refactor channel creation

JamesNK:
Removed ChannelBuilder. It was basically just setting values on an options class. Options + class initializer syntax feels like more idomatic .NET


src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
```diff
@@ -33,6 +33,41 @@ namespace Microsoft.Extensions.DependencyInjection
...
+        /// </remarks>
+        public static IHttpClientBuilder AddGrpcClient<TClient>(this IServiceCollection services)
+            where TClient : ClientBase
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            var name = TypeNameHelper.GetTypeDisplayName(typeof(TClient), fullName: false);
```
JunTaoLuo: Is there a mismatch in the doc comment and the value of fullName here?

codymullins:
Any idea when the next release will go out with these changes?

JamesNK:
You can use the nightly builds today or wait for preview 9.

test/Grpc.Net.Client.Tests/GrpcChannelTests.cs
```diff
@@ -61,14 +55,11 @@ public void Build_SslCredentialsWithHttp_ThrowsError()
...
-            // Assert
-            var ex = Assert.Throws<InvalidOperationException>(() => channelBuilder.Build());
+            // Arrange & Assert
+            var ex = Assert.Throws<InvalidOperationException>(() => GrpcChannel.ForAddress("https://localhost", new GrpcChannelOptions
+            {
+                Credentials = new SslCredentials("rootCertificates!!!")
+            }));
 
             // Act
             Assert.AreEqual("Using SslCredentials with non-null arguments is not supported by GrpcChannel.", ex.Message);
```
mustakimali: @JamesNK can you explain why is it not supported? how can one pass a ca-certificate? does it automatically looks at the system ca store? Thanks.

JamesNK: Correct. Specifying manually isn’t supported or needed with http client

mustakimali: Thanks for your reply. Last question,

When creating instances of `Channel` I could pass [`ChannelOption`](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core/ChannelOptions.cs#L60) and specify load-balancing policy.

```
...
var channelOptions = new List<ChannelOption>
{
    new ChannelOption("grpc.lb_policy_name", "round_robin")
};
var channel = new Channel("https://example.com", chCreds, channelOptions);
```

but now when registering gRPC client using GrpcClientFactory I don't see any option to pass these to [`GrpcChannelOption`](https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.Net.Client/GrpcChannelOptions.cs#L26). Is it not supported?

JamesNK: Those options? No. That's a different gRPC client implementation.

PR #450, opened 2019-08-12
Respect NoCompress flag in the client

JamesNK:


JamesNK:
Fix client_compressed_streaming in https://github.com/grpc/grpc/pull/19694

PR #451, opened 2019-08-12
Add server compression interop tests

JamesNK:
Fix server compression interop tests in https://github.com/grpc/grpc/pull/19694


JamesNK:
Note that client doesn't verify responses are compressed. There is no API for it and that test is optional - https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#server_compressed_unary

Issue #454, opened 2019-08-14
Dispose HttpClient created by channel


JamesNK:
Channel can be configured with an HttpClient, or create one if no client is specified. Need to consider disposing the client when GrpcChannel creates it.

PR #458, opened 2019-08-15
Add channel setting for disposing HttpClient

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/454

### DIFF
--- a/test/Grpc.Net.Client.Tests/GrpcChannelTests.cs
+++ b/test/Grpc.Net.Client.Tests/GrpcChannelTests.cs
@@ -177,7 +177,7 @@ namespace Grpc.Net.Client.Tests
         }
 
         [Test]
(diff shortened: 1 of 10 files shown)


test/Grpc.Net.Client.Tests/GrpcChannelTests.cs
```diff
@@ -29,66 +32,187 @@ public class GrpcChannelTests
...
+            }
+
+            protected override void Dispose(bool disposing)
+            {
+                Disposed = true;
+            }
+        }
+
+        [Test]
+        public void Dispose_CalledWhenHttpClientSpecified_Disposed()
```
JunTaoLuo: nit: shouldn't the name be `NotCalled`. This test is verifying that an HttpClient is specified, but DisposeHttpClient is not set then dispose is not called?

PR #462, opened 2019-08-16
Seal metadata type and cleanup InterceptorCollection

JamesNK:


src/Grpc.AspNetCore.Server/InterceptorCollection.cs
```diff
@@ -28,21 +28,44 @@ namespace Grpc.AspNetCore.Server
...
+
+            Add(new InterceptorRegistration(interceptorType, args));
         }
 
         /// <summary>
         /// Append a set of interceptor registrations to the end of the pipeline.
         /// </summary>
         /// <param name="registrations">The set of interceptor registrations to add.</param>
-        public void AddRange(IEnumerable<InterceptorRegistration> registrations)
+        internal void AddRange(IEnumerable<InterceptorRegistration> registrations)
```
JamesNK: InterceptorRegistration can only be create by Add. As far as I can tell this is never useful.

JunTaoLuo: The original idea for adding this originally was that you can create a collection of interceptors that can be added to different services.

PR #469, opened 2019-08-20
Fix package versioning

JamesNK:
Directory.Build.props was resolved before IsGrpcPublishedPackage was set. Fix by moving dependent config to targets

### DIFF
--- a/build/version.props
+++ b/build/version.props
@@ -1,6 +1,14 @@
 <Project>
   <PropertyGroup>
-    <!-- version of all grpc-dotnet packages -->
+
+    <!-- package version of grpc-dotnet -->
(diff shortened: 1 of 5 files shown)


build/version.props
```diff
@@ -1,6 +1,14 @@
 <Project>
   <PropertyGroup>
-    <!-- version of all grpc-dotnet packages -->
+
+    <!-- package version of grpc-dotnet -->
```
JunTaoLuo: Can we do something like https://github.com/aspnet/AspNetCore/blob/master/eng/Versions.props#L9-L13

Let's separate out major, minor, patch and pre-release label and use those to construct the GrpcDotnet*Version properties.

JamesNK: Maybe, but something to think about in the future.

PR #471, opened 2019-08-22
Remove git hash from client user agent

JamesNK:


JunTaoLuo:
Is the test failure flakiness?

JamesNK:
Yes

Issue #474, opened 2019-08-23
Validate request is using HTTP/2


JamesNK:
The Kestrel endpoint might not be configured to use HTTP/2. Validate requests are using HTTP/2 to avoid this error - https://github.com/grpc/grpc-dotnet/issues/241#issuecomment-522800925

PR #475, opened 2019-08-24
Validate request protocol

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/474

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
@@ -23,6 +23,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
 using Microsoft.Extensions.Logging;
+using Microsoft.Net.Http.Headers;
 
 namespace Grpc.AspNetCore.Server.Internal.CallHandlers
 {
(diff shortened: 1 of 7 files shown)


src/Grpc.AspNetCore.Server/Internal/CallHandlers/ServerCallHandlerBase.cs
```diff
@@ -55,7 +56,14 @@ public Task HandleCallAsync(HttpContext httpContext)
...
             {
-                GrpcProtocolHelpers.SendHttpError(httpContext.Response, StatusCodes.Status415UnsupportedMediaType, StatusCode.Internal, error!);
+                GrpcProtocolHelpers.SendHttpError(httpContext.Response, StatusCodes.Status415UnsupportedMediaType, StatusCode.Internal, error);
+                return Task.CompletedTask;
+            }
+            if (httpContext.Request.Protocol != GrpcProtocolConstants.Http2Protocol)
+            {
+                var protocolError = $"Request protocol '{httpContext.Request.Protocol}' is not supported.";
+                GrpcProtocolHelpers.SendHttpError(httpContext.Response, StatusCodes.Status426UpgradeRequired, StatusCode.Internal, protocolError);
+                httpContext.Response.Headers[HeaderNames.Upgrade] = GrpcProtocolConstants.Http2Protocol;
```
JamesNK: @Tratcher @halter73 Is 426 the right status to return when a HTTP/1.1 request comes to a gRPC?

Tratcher: Anything you put down will be primarily for diagnostic purposes, there's no automatic transition support beyond ALPN.
426 is OK, 421 looks like the second best option.

JamesNK: Diagnostics and providing a clear error message is the goal here. Currently HTTP/1.1 requests get through to the gRPC request delegate and only fail when the framework tries to get or set trailers.

Will use 426.

test/FunctionalTests/Client/CompressionTests.cs
```diff
@@ -41,7 +41,7 @@ public async Task SendCompressedMessage_ServiceCompressionConfigured_ResponseGzi
 
             string? requestMessageEncoding = null;
             string? responseMessageEncoding = null;
-            using var httpClient = Fixture.CreateClient(new TestDelegateHandler(
+            using var httpClient = Fixture.CreateClient(messageHandler: new TestDelegateHandler(
```
davidfowl: ?

PR #484, opened 2019-09-03
Add client micro-benchmark

JamesNK:
Basic micro-benchmark of a unary call.


perf/Grpc.AspNetCore.Microbenchmarks/Program.cs
```diff
@@ -29,24 +29,24 @@ static void Main(string[] args)
         }
 #else
         // Profiling option. This will call methods explicitly, in-process
```
jtattermusch: nit: I think the profiling option will become hard to maintain as more benchmarks are added.

JamesNK: It's not meant to be an exhaustive list, it's just something to call a benchmark directly (which you update as needed).

The benchmark framework doesn't play nicely with profilers.

PR #503, opened 2019-09-10
Refactor examples into their own projects

JamesNK:
Each example now has its own server and solution. Self contained examples will be easier to understand.


examples/Aggregator/Server/Server.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Protobuf Include="..\Proto\aggregate.proto" GrpcServices="Server" Link="Protos\aggregate.proto" />
```
JunTaoLuo: We should remove GrpcServices attributes. The default here is Server.

examples/Greeter/Client/Client.csproj
```diff
@@ -6,12 +6,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Protobuf Include="..\..\Proto\greet.proto" GrpcServices="Client" Link="Protos\greet.proto" />
+    <Protobuf Include="..\Proto\greet.proto" GrpcServices="Client" Link="Protos\greet.proto" />
 
     <PackageReference Include="Google.Protobuf" Version="$(GoogleProtobufPackageVersion)" />
+    <PackageReference Include="Grpc.Net.Client" Version="$(GrpcDotNetPackageVersion)" />
```
JunTaoLuo: Curious why use the package instead of the project?

examples/Worker/Server/Startup.cs
```diff
@@ -0,0 +1,49 @@
...
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapGrpcService<CounterService>();
```
JunTaoLuo: Is this the same as Counter server? I understand the simplicity of one client corresponding to one server, but I wonder if it's better to add a Readme to specify which client works with with server. I think it would highlight the composability of gRPC clients and servers.

examples/Greeter/Client/Client.csproj
JamesNK: Because these samples are designed to be as independent as possible (yes the variable could also be hardcoded but I didn't want to duplicate it everywhere)

examples/Worker/Server/Startup.cs
JamesNK: This sample still has its own server. The contract just happens to be CounterService.

examples/Aggregator/Server/Startup.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapGrpcService<GreeterService>();
```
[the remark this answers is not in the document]
JamesNK: They are here because the aggregator service needs other services to call. That's the whole point of the sample.

examples/Aggregator/Server/Server.csproj
JamesNK: How is it defaulted? The template adds `GrpcServices="Server"`

JunTaoLuo: Oops, brainfart, GrpcServices="Server" is needed.

PR #509, opened 2019-09-11
Allow for a base URI that includes path components.

mvarblow:
Fixes handling of relative paths in gRPC calls to allow for a base URI that includes path components. Without this change, only 'bare' domains can be used for the base address on a gRPC client.

### DIFF
--- a/src/Grpc.Net.Client/Internal/GrpcCall.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCall.cs
@@ -68,7 +68,7 @@ namespace Grpc.Net.Client.Internal
             _callCts = new CancellationTokenSource();
             _callTcs = new TaskCompletionSource<Status>(TaskContinuationOptions.RunContinuationsAsynchronously);
             Method = method;
-            _uri = new Uri(method.FullName, UriKind.Relative);
+            _uri = new Uri(RemoveLeadingSlash(method.FullName), UriKind.Relative);
             _logScope = new GrpcCallScope(method.Type, _uri);
             Options = options;
             Channel = channel;
@@ -88,6 +88,8 @@ namespace Grpc.Net.Client.Internal
                 throw new InvalidOperationException("Deadline must have a kind DateTimeKind.Utc or be equal to DateTime.MaxValue or DateTime.MinValue.");
             }
         }
+
+        private string RemoveLeadingSlash(string uri) => uri.StartsWith('/') ? uri.Substring(1) : uri;
 
         public Task<Status> CallTask => _callTcs.Task;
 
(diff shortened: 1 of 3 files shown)


thelinuxfoundation:

* One or more commits are not associated with a GitHub user. The email used in the Git commit must match a verified email address in a GitHub user profile. For more information, please see https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/

_Regards,
CLA GitHub bot_

mvarblow:
Verified email address used for commit to satisfy the CLA GitHib bot.

mvarblow:
I signed the CLA

JamesNK:
I believe the current behavior is intentional. gRPC services are always located in the root of a server.

@jtattermusch I recall this coming up before and you felt that only root gRPC services should be supported on the server and be callable by the client.

jtattermusch:
> Why should gRPC services only be supported from the site root? That seems like a pretty arbitrary requirement.
> 
> We'd like to use gRPC to communicate with our custom developed Windows services. We currently have the services configured to use the HTTP.SYS hosting model under a base URL path (e.g. /my-service-name/), which allows for them to run under the default 443 port on the server and inherit their SSL settings from the IIS default web site. This is way more convenient to manage than using Kesterel hosting and having to configure SSL settings for each service/instance.

Prefixed URIs are not allowed intentionally. A few reasons why:

1. all other gRPC implementations only support the rooted model and maintaining interoperability with the other gRPC implementations is key. By choosing a prefixed URL, the user (often times unknowingly) sacrifices interoperability with other gRPC services and that's something we'd like to prevent

2. when using protobuf, the path always corresponds to fully-qualified name of the service so just knowing the namespaces in the .proto definitions also makes it clear how services are going to be exposed on a given server. (this is useful for e.g. tooling support and integrations as well as some grpc features).

3. prefixed URIs are actually strongly discouraged by the gRPC spec:

    https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
    > Some gRPC implementations may allow the Path format shown above to be overridden, but this functionality is strongly discouraged. gRPC does not go out of its way to break users that are using this kind of override, but we do not actively support it, and some functionality (e.g., service config support) will not work when the path is not of the form shown above.

Also see https://github.com/grpc/grpc-dotnet/issues/110, https://github.com/grpc/grpc/issues/14900

JamesNK:
I think unless gRPC guidance changes, or there is a compelling scenario, we won't do this. Closing.

mvarblow:
I would argue that our scenario is compelling :), but I see that it goes against the gRPC guidance. So I can understand rejecting the change. Thanks for taking the time to evaluate and respond to the pull request!

PR #511, opened 2019-09-12
Benchmarks: Add header for h2load client, use protocol on server

JamesNK:
Grpc.Core will fail if it receives a request without TE=trailers header.

Improve support for h2 and h2c benchmarks


JamesNK:
@jtattermusch @JunTaoLuo Please review so I can commit this. I want to test what happens to benchmarks and it takes some time for them to start and run.

perf/benchmarkapps/NativeServer/Program.cs
```diff
@@ -38,10 +33,16 @@ public static void Main(string[] args)
                .AddCommandLine(args)
                .Build();
 
+            var protocol = config["protocol"] ?? string.Empty;
+            if (!protocol.Equals("h2c", StringComparison.OrdinalIgnoreCase))
+            {
+                throw new InvalidOperationException("Only h2c is supported by C-core benchmark server.");
```
JunTaoLuo: Is this just how we're setting up the benchmarks server? Or is there another limitation that's causing this?

JamesNK: Getting TLS certificates to work in Grpc.Core is a pain in the butt so I just didn't do it. Can change in the future.

PR #515, opened 2019-09-16
Refactor perf benchmarks to use grpc.io scenarios

JamesNK:


JamesNK:
> It's not necessarily the goal to use the exact same scenario definitions, I just wanted to point it out.

I know, but consistency in benchmarks helps. And using the same tests in the future allows this server to be used in the grpc.io benchmark environment.

Issue #516, opened 2019-09-16
Methods with streaming outbound messages block forever on failure


Earthmark:
### EDIT: I originally hit this issue with pre1 of the nuget package, with pre2 it seems to no longer be the case. Please close this issue as I have yet to find that button on my end.

Issue:
When calling WriteAsync on a request stream of a streaming call where the client fails to connect to the server, the write call seems to block forever.

Expected: 
HttpRequestException to be raised after the expected request timeout.

Note: CompleteAsync doesn't explode either, but it returns instantly, and attempting to write after completing does raise the appropriate exception about the stream being completed already.

This does not seem to effect the Response pipe in either single value or streaming values modes, as they both raise HttpRequestException after the expected timeout. This also doesn't seem to effect sending a single value, as that also raises an HttpRequestException after the web timeout.

Code to replicate the issue:
This was minified, but the original code used a client inside a worker, it was modified to just run inside main so I could copy it in here easier, the same bug occurs inside a hosted service worker as happens inside main.
Program.cs
```
using System;
using System.Threading.Tasks;
using grpcTest;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;

namespace GrpcHangTest
{
  class Program
  {
    public static async Task Main()
    {
      var host = Host.CreateDefaultBuilder().ConfigureServices(services =>
      {
        services.AddGrpcClient<Greeter.GreeterClient>(o =>
        {
          // Configure this to any port that's not bound.
          o.Address = new Uri("http://localhost:8079");
        });
      }).Build();

      var client = host.Services.GetRequiredService<Greeter.GreeterClient>();

      using var request = client.SendStream();
      try
      {
        await request.RequestStream.WriteAsync(new Request
        {
          Payload = "Potato"
        });
      }
      finally
      {
        Console.WriteLine("I am never hit!");
      }
    }
  }
}
```
greet.proto
```
syntax = "proto3";

option csharp_namespace = "grpcTest";

package Greet;

service Greeter {
  rpc SendStream (stream Request) returns (Reply);
}

message Request {
  string payload = 1;
}

message Reply {
  string payload = 1;
}
```
Csproj (for completeness)
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
[...]

PR #517, opened 2019-09-17
Add client streaming failure test

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/516


Earthmark:
Thank you for adding the test, I ain't too familiar with the code base but looking over the testing infrastructure I do agree this re-creates and tests the issue I was observing.
Only suggestion I have is to possibly add a timeout, as the issue resulted in an endless wait and I'm not sure if this has a global timeout set up.

JamesNK:
> Only suggestion I have is to possibly add a timeout, as the issue resulted in an endless wait and I'm not sure if this has a global timeout set up.

That is what `DefaultTimeout()` is for.

PR #524, opened 2019-09-18
Fix generating XML documentation

JamesNK:
I noticed that the XML docs file is missing from the gRPC package 😭 Apparently Directory.Build.targets is too late to set the value.


chwarr:
Is `Directory.Build.props` also too late?

jtattermusch:
> Is `Directory.Build.props` also too late?

That's a good point. Let's try that.

JamesNK:
The reason it was in Directory.Build.targets is we only want to set it to true when `IsGrpcPublishedPackage` is true - otherwise unit tests and other projects that don't need XML docs start generating warnings about missing XML docs. But settings dependent on a flag set in csproj can only be placed in targets which is too late.

PR #531, opened 2019-09-19
Attempt to add scripts for generating API docs

jtattermusch:
This this the result of my first attempt. It is slightly hacky, but it seems to work.

- the API docs will be added to grpc.io  in a separate section (so there will be gRPC C# and grpc-dotnet section, to avoid mixing up two different API for creating servers and channels in the same generated docs)
- the generated docs include docs for Grpc.Core.Api  (the way I added it is somewhat hacky)
- there are some warnings when generating the docfx docs, but overall the docs look fine, so these can be fixed later.
- the scripts are heavily inspired by https://github.com/grpc/grpc/tree/master/src/csharp/docfx

### DIFF
--- a/docfx/.gitignore
+++ b/docfx/.gitignore
@@ -1,0 +1,2 @@
+html
+obj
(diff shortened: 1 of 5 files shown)


jtattermusch:
@JamesNK @shirhatti @JunTaoLuo

PR #538, opened 2019-09-21
Add's support for method level interceptors

cosullivan:
This PR adds support for method level interceptors that can be configured through an extension to the service options 

```
.AddServiceOptions<GreeterService>(
    options =>
    {
        options.AddMethodOptions(nameof(GreeterService.SayHello), 
            methodOptions =>
            {
                methodOptions.Interceptors.Add<TracingInterceptor>();
            });
      });
```

Or via the ServiceMethodProviderContext with an additional option/callback to configure and override the method level options.


JamesNK:
I want to support this but I have a different plan to do it. I want an approach that will work for all configuration settings.

Issue #506, opened 2019-09-11
dotnet-grpc should target the last stable shared framework and allow roll-forward to new stable version 


JunTaoLuo:
This is how global tools are built in dotnet/diagnostics for example and we should do this so that we don't break users when we are releasing new previews. Currently, the tool will use the shared framework we built with which for previews may not yet be released. This should be less of a problem once we release 3.0 but it's still good to future proof by using the last stable shared framework.

Issue #360, opened 2019-06-27
Fix build conflict warnings


JamesNK:
> C:\Program Files\dotnet\sdk\3.0.100-preview7-012687\Microsoft.Common.CurrentVersion.targets(2106,5): warning MSB3277: Found conflicts between different versions of "Microsoft.Win32.Registry" that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed. [C:\Development\Source\grpc-dotnet\src\dotnet-grpc\dotnet-grpc.csproj]
> C:\Program Files\dotnet\sdk\3.0.100-preview7-012687\Microsoft.Common.CurrentVersion.targets(2106,5): warning MSB3277: Found conflicts between different versions of "System.Security.Principal.Windows" that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed. [C:\Development\Source\grpc-dotnet\src\dotnet-grpc\dotnet-grpc.csproj]

JamesNK:
Can fix System.Security.Principal.Windows by explicitly referencing 4.5.1 from NuGet. Microsoft.Win32.Registry can only be fixed by referencing a 3.0 preview package.

We can suppress warning for now and fix post 3.0 when package is no longer preview.

PR #539, opened 2019-09-23
Update global tool and resolve build warnings.

JunTaoLuo:
Addresses https://github.com/grpc/grpc-dotnet/issues/360 and https://github.com/grpc/grpc-dotnet/issues/506. Also resolves a few pack warnings.


src/dotnet-grpc/runtimeconfig.template.json
```diff
@@ -0,0 +1,3 @@
+{
+  "rollForwardOnNoCandidateFx": 2
```
JamesNK: Can you add a comment of what this does

JunTaoLuo:
They show up when you pack, not during build.

src/dotnet-grpc/runtimeconfig.template.json
JunTaoLuo: Not in this file? I'll add it in the csproj.

Directory.Build.targets
```diff
@@ -23,6 +24,8 @@
 
   <!-- IsGrpcPublishedPackage is set in csproj so related config must be in targets file -->
   <ItemGroup Condition="'$(IsGrpcPublishedPackage)' == 'true'">
+    <None Include="$(PackageIconFullPath)" Pack="true" PackagePath="\"/>
```
JamesNK: What is this for?

JunTaoLuo: The docs recommend this: https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file

src/Grpc.AspNetCore/Grpc.AspNetCore.csproj
```diff
@@ -13,6 +13,12 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <!-- Avoid NU5128 warning. TFM in dependencies group of the nuspec must match lib/ref content. -->
+    <None Include="lib\**\*">
```
JamesNK: Does this include a DLL in the package? I explicitly excluded this with `IncludeBuildOutput` on line 11.

JunTaoLuo: No it does not, it includes just the dummy file.

image

Issue #410, opened 2019-07-23
Versioning strategy post 3.0


JunTaoLuo:
We need to consider how we line up releases between gRPC and AspNetCore. This will be important for plans after 3.0.

PR #542, opened 2019-09-24
Add documentation for versioning policy going forward

JunTaoLuo:
Capture the versioning policy that came out of discussions on https://github.com/grpc/grpc-dotnet/issues/410.

Also included:
- updated packages chart


JunTaoLuo:
As part of updating the dependency diagram, I noticed that there are some redundant edges. Grpc.AspNetCore.Server and Grpc.Net.Client get Grpc.Core.Api transitively through Grpc.Net.Common and therefore doesn't need a direct reference. This won't be a breaking change since the dependency won't be lost. Thoughts?

doc/versioning.md
```diff
@@ -0,0 +1,32 @@
...
+- Grpc.Net.Client
+- Grpc.Net.ClientFactory
+- Grpc.Net.Common
+- dotnet-grpc
+
+## Design considerations
+
+### Major version number
+
+The initial major version number was set to 2 to match the release of corresponding Grpc.Core packages. Going forward, the major version will be incremented when there's a need to ingest a major Grpc.Core or ASP.NET Core dependency for breaking changes.
```
jtattermusch: We still need to comply with https://github.com/grpc/grpc/blob/master/doc/versioning.md. It would be good to use wording that makes it clear we won't bump major version unless necessary.

doc/versioning.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Minor version number
+
+The initial minor version number was set to 23 to match the release of corresponding Grpc.Core packages. Going forward, we will be releasing on the Grpc.Core schedule and should match the minor version number of the corresponding release.
```
jtattermusch: nit: link to https://github.com/grpc/grpc/blob/master/doc/grpc_release_schedule.md?

doc/versioning.md
```diff
@@ -0,0 +1,32 @@
...
+- Grpc.Net.Client
+- Grpc.Net.ClientFactory
+- Grpc.Net.Common
+- dotnet-grpc
+
+## Design considerations
+
+### Major version number
+
+The initial major version number was set to 2 to match the release of corresponding Grpc.Core packages. Going forward, the major version will be incremented when there's a need to ingest a major Grpc.Core or ASP.NET Core dependency for breaking changes. Grpc.Core follows these [versioning policies](https://github.com/grpc/grpc/blob/master/doc/versioning.md) and major version is only incremented on rare occasions.
```
chwarr: Minor: Consider "Under [Grpc.Core's versioning policy](https://github.com/grpc/grpc/blob/master/doc/versioning.md) the major version is only..." so that the sentence makes sense with or without the link and the link adds additional value.

doc/versioning.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Minor version number
+
+The initial minor version number was set to 23 to match the release of corresponding Grpc.Core packages. Going forward, we will be releasing on the Grpc.Core schedule and should match the minor version number of the corresponding release. The Grpc.Core release schedule can be found [here](https://github.com/grpc/grpc/blob/master/doc/grpc_release_schedule.md).
```
chwarr: Minor: consider just hyperlinking the first mention of the schedule. "Going forward, we will be releasing on the [Grpc.Core schedule](https://github.com/grpc/grpc/blob/master/doc/grpc_release_schedule.md)..."

doc/versioning.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Patch version number
+
+The patch number will increment for patches of existing major.minor release. These patches are built out of the v{major}.{minor}.x branches on github, e.g. https://github.com/grpc/grpc-dotnet/tree/v2.23.x.
```
chwarr: Minor: "...branches on **G**it**H**ub..."

doc/versioning.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Pre-release and nightly suffixes
+
+The suffix -pre{x} will be added to pre-release packages released to nuget.org. These are non-stable packages intended for early ingestion of upcoming releases.
+The suffix -dev is added to packages produced by nightly builds available on https://grpc.jfrog.io/grpc/api/nuget/v3/grpc-nuget-dev.
```
chwarr: Minor: Consider "available from the NuGet feed https://grpc.jfrog.io/grpc/api/nuget/v3/grpc-nuget-dev"

doc/versioning.md
jtattermusch: The doc says "Grpc.Core follows the versioning policy .." , but the policy is gRPC-wide and applies to all languages (including grpc-java, grpc-go and grpc-dotnet). It might be simpler to just say that grpc-dotnet complies with the gRPC-wide versioning policy [link]

JunTaoLuo:
:up::date:

Issue #571, opened 2019-10-02
Authentication failed - wrong status code client side


AnthonyGiretti:
Hi!

I experienced an issue with grpc-dotnet client side (app console)

When the server side (grpc-dotnet too) authentication fails, I receive (with the HttpClient client side) the wrong grpc status code: CANCELLED. According to gRPC spec I should receive instead UNAUTHENTICATED.

Thanks

JamesNK:
You get a CANCELLED status code because the server returns a 400 (unauthorized) status code header. That is quite different from returning a 200 status code with an UNAUTHENTICATED grpc-status in the response trailer.

JamesNK:
@jtattermusch Do you know of any clients or servers that have any built in support for mapping 401 to a gRPC status?

AnthonyGiretti:
Thank you for the answer.
But allow me to disagree

The server returns 401 then it should be mapped to UNAUTHENTICATED, (Unauthenticated doesn't exist in http then it should be considered as Unauthorized).
In grpc specs UNAUTHENTICATED means: _The request does not have valid authentication credentials for the operation._

Then we cannot expect neither 200 (you are right) nor CANCELLED because in grpc world CANCELLED means: _The operation was cancelled, typically by the caller._

Here are GRPC statuses code: https://grpc.github.io/grpc/core/md_doc_statuscodes.html

Does it make sense or I'm totally wrong? 

Thanks!

JamesNK:
My interpretation is the server should return 200 OK with a status of UNAUTHENTICATED. A way to get the behavior you want is to have middleware on the server that changes 401 status responses to 200 with an UNAUTHENTICATED trailer status for gRPC endpoints.

JamesNK:
I found this. There maybe a mapping: https://grpc.github.io/grpc/cpp/md_doc_http-grpc-status-mapping.html

@jtattermusch Should the client map the gRPC status from these HTTP status codes?

AnthonyGiretti:
Yes!
That's what I was fighting for :)



401 Unauthorized | UNAUTHENTICATED
-- | --



401 Unauthorized | UNAUTHENTICATED
-- | --



401 Unauthorized | UNAUTHENTICATED
-- | --

chwarr:
[...]

PR #575, opened 2019-10-02
Map HttpStatusCode to gRPC status code

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/571

Implement https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md

### DIFF
--- a/src/Grpc.Net.Client/Internal/GrpcCall.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCall.cs
@@ -237,6 +237,13 @@ namespace Grpc.Net.Client.Internal
         {
             Log.ResponseHeadersReceived(Logger);
 
+            // gRPC status can be returned in the header when there is no message (e.g. unimplemented status)
+            // An explicitly specified status header has priority over other failing statuses
+            if (GrpcProtocolHelpers.TryGetStatusCore(httpResponse.Headers, out var status))
+            {
+                return status;
+            }
+
             if (httpResponse.StatusCode != HttpStatusCode.OK)
             {
                 var statusCode = MapHttpStatusToGrpcCode(httpResponse.StatusCode);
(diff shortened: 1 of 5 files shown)


src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -265,6 +266,35 @@ public Task<TResponse> GetResponseAsync()
             return null;
         }
 
+        private static StatusCode MapHttpStatusToGrpcCode(HttpStatusCode httpStatusCode)
```
chwarr: Sigh. This matches the mapping documentation and Java and Go. However, the C Core library maps differently. I assume this mapping is fine, but I don't know for sure.

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -239,7 +239,8 @@ public Task<TResponse> GetResponseAsync()
 
             if (httpResponse.StatusCode != HttpStatusCode.OK)
             {
-                return new Status(StatusCode.Cancelled, "Bad gRPC response. Expected HTTP status code 200. Got status code: " + (int)httpResponse.StatusCode);
+                var statusCode = MapHttpStatusToGrpcCode(httpResponse.StatusCode);
```
chwarr: This looks like it may be too early to perform this translation. The C code checks for a grpc-status value before mapping the HTTP2 status code, regardless of what the HTTP2 status code is--if I'm reading it correctly. [Java also looks for grpc-status first](https://github.com/grpc/grpc-java/blob/81da3eb95be37fa0647ce8da2e19de96ab84c600/core/src/main/java/io/grpc/internal/Http2ClientStreamTransportState.java#L181-L201). Not sure about Go...

Issue #580, opened 2019-10-03
In examples the proto package names should be all lowercase


Ciantic:
This is very minor thing, but package names in the protobuf example files such as:

https://github.com/grpc/grpc-dotnet/blob/master/examples/Greeter/Proto/greet.proto

Should be in lowercase like `package greet;` right now they are PascalCased e.g. `package Greet;` as if it's some .NET namespace. Same applies to all examples, the package names should be in lowercase.

JamesNK:
Is that a convention? Do you know if it is documented anywhere or is it sort of a general consensus?

Ciantic:
It comes from this:

https://developers.google.com/protocol-buffers/docs/style

> Files should be named lower_snake_case.pro
> Package name should be in lowercase, and should correspond to the directory hierarchy. e.g., if a file is in my/package/, then the package name should be my.package.

But generally all text is lower_snake_case in protobufs.

JamesNK:
Ok, I'll update samples, documentation and template.

JamesNK:
Done

PR #581, opened 2019-10-03
Change example and test package names to lowercase

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/580

### DIFF
--- a/test/FunctionalTests/Server/UnaryMethodTests.cs
+++ b/test/FunctionalTests/Server/UnaryMethodTests.cs
@@ -346,7 +346,7 @@ namespace Grpc.AspNetCore.FunctionalTests.Server
 
             // Act 1
             var response = await Fixture.Client.PostAsync(
-                "SingletonCount.Counter/IncrementCount",
+                "singleton_count.Counter/IncrementCount",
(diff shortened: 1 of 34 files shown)


test/FunctionalTests/Server/UnaryMethodTests.cs
```diff
@@ -346,7 +346,7 @@ public async Task SingletonService_PrivateFieldsPreservedBetweenCalls()
 
             // Act 1
             var response = await Fixture.Client.PostAsync(
-                "SingletonCount.Counter/IncrementCount",
+                "singleton_count.Counter/IncrementCount",
```
JunTaoLuo: I do find a mix of snake_case and PascalCase looks odd though. How do other languages that have different casing schemes handle this?

jtattermusch: I think package names should be  `lower_case_with_underscores` according to the .proto style guide.

Because `singleton_count.Counter/IncrementCount` is part of the method URL, it needs to contain a verbatim protobuf package, so that the URL is the same across languages. Note that under normal use, both client and server will generate this URL internally for you so you won't see it at all.

Issue #590, opened 2019-10-08
System.InvalidOperationException: Error binding gRPC service


khteh:
### What version of gRPC and what language are you using?
2.23.2, C#

### What operating system (Linux, Windows,...) and version?
Ubuntu 19.04

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
```
$ dn --info
.NET Core SDK (reflecting any global.json):
 Version:   3.0.100
 Commit:    04339c3a26

Runtime Environment:
 OS Name:     ubuntu
 OS Version:  19.04
 OS Platform: Linux
 RID:         ubuntu.19.04-x64
 Base Path:   /usr/share/dotnet-3.0.100/sdk/3.0.100/

Host (useful for support):
  Version: 3.0.0
  Commit:  7d57652f33

.NET Core SDKs installed:
  3.0.100 [/usr/share/dotnet-3.0.100/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.App 3.0.0 [/usr/share/dotnet-3.0.100/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.0.0 [/usr/share/dotnet-3.0.100/shared/Microsoft.NETCore.App]

To install additional .NET Core runtimes or SDKs:
  https://aka.ms/dotnet-download
```

### What did you do?
If possible, provide a recipe for reproducing the error. Try being specific and include code snippets if helpful.
I have the following .proto:
```
service Auth {
    rpc RefreshSignIn(RefreshSignInRequest) returns (RefreshSignInResponse);
}
message RefreshSignInRequest {
    Biz4x.Core.Grpc.User user = 1;
}
message RefreshSignInResponse {
    string id = 1;
    Biz4x.Core.Grpc.Response response = 2;
}
```
And I get the following runtime exception which I have no clue about:
```
15:18:33 [Fatal] () Application startup exception
System.InvalidOperationException: Error binding gRPC service 'AuthService'.
 ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.Reflection.AmbiguousMatchException: Ambiguous match found.
   at System.RuntimeType.GetMethodImplCommon(String name, Int32 genericParameterCount, BindingFlags bindingAttr, Binder binder, CallingConventions callConv, Type[] types, ParameterModifier[] modifiers)
[...]

PR #591, opened 2019-10-08
Fix method discovery when there are duplicate names

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/590


test/Grpc.AspNetCore.Server.Tests/TestObjects/Services/GreeterServiceWithDuplicateNames.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace Grpc.AspNetCore.Server.Tests.TestObjects
+{
+    public class GreeterServiceWithDuplicateNames : Greeter.GreeterBase
+    {
+        public override Task<HelloReply> SayHello(HelloRequest request, ServerCallContext context)
+        {
+            return base.SayHello(request, context);
+        }
+
+        public static Task<HelloReply> SayHello(ServerCallContext context)
```
JunTaoLuo: Maybe throw in a non-public method as well.

Issue #594, opened 2019-10-08
Add sample that uses IProgress<T>


JamesNK:
Add a [sample](https://github.com/grpc/grpc-dotnet/tree/master/examples) of the server reporting progress and the client using it to update an `IProgress<T>`.

PR #597, opened 2019-10-09
Add progessor example

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/594


examples/Progressor/Client/ResponseProgress/ResponseProgress.cs
```diff
@@ -0,0 +1,65 @@
...
+#endregion
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+using Grpc.Core;
+
+namespace Client.ResponseProgress
+{
+    public class ResponseProgress<TMessage, TResult, TProgress> : Progress<TProgress> where TMessage : IProgressMessage<TResult, TProgress>
```
JamesNK: `TMessage` is required because `IAsyncStreamReader<T>` is not covariant.

JamesNK: https://github.com/grpc/grpc/pull/20549

JunTaoLuo: There's a new Grpc.Core.Api release, I think we can update to that and make the change?

JamesNK: Still waiting on 2.25.0-pre1 of Grpc.AspNetCore here

PR #600, opened 2019-10-12
Validate gRPC response protocol is not downgraded

JamesNK:
HttpClient sends HTTP/1.1 and HTTP/2 with ALPN negotiation, so a server could potentially start a HTTP/1.1 connection if it doesn't support HTTP/2.

Validate response is always HTTP/2.

@jtattermusch What does Grpc.Core do here? I'm guessing it would only ever allow HTTP/2.


JunTaoLuo:
ping @jtattermusch what does C Core do here?

jtattermusch:
@yashkt do you know the answer? (if not you, who does?)

yashykt:
gRPC does not work with HTTP/1.1. It would need HTTP/2. I don't remember the exact behavior from the server in case the client sends HTTP/1.1 but I am guessing that it would be something like a bad response.

test/FunctionalTests/Infrastructure/InProcessTestServer.cs
```diff
@@ -84,7 +84,7 @@ public InProcessTestServer(Action<IServiceCollection>? initialConfigureServices)
         public override void StartServer()
         {
             // We're using 127.0.0.1 instead of localhost to ensure that we use IPV4 across different OSes
-            var url = "http://127.0.0.1";
+            //var url = "http://127.0.0.1";
```
JunTaoLuo: not needed?

Issue #609, opened 2019-10-18
implement SerializationContext.GetBufferWriter


mgravell:
the contextual marshaller API has two methods intended for the low-alloc buffer scenarios: `DeserializationContext.PayloadAsReadOnlySequence()` and `SerializationContext.GetBufferWriter()/Complete()`. At the current time, only the former (`DeserializationContext`) is implemented, with the latter (`SerializationContext`) throwing `NotImplementedException` for this call.

Unfortunately, there is no way to feature-detect this, other than catching the exception and falling back to the older `Complete(byte[])` API.

There are 3 ways this could be improved, in increasing difficulty and benefit:

1. make it possible to feature-detect without a `catch`; pragmatic suggestion: just return `null` from `GetBufferWriter()` if the API isn't supported; there are *very few consumers* of this API, so changing the semantic here should not be a big problem (/cc @jtattermusch )

2. implement it using a basic implementation, perhaps using `ArrayBufferWriter<byte>` internally

3. implement it "properly" using direct write and discontiguous buffers

Context: I've "lit up" the `ReadOnlySequence<byte>` and `IBufferWriter<byte>` APIs in protobuf-net v3, and I'm looking to integrate them with protobuf-net.Grpc; right now, I'm using the `catch` against `NotSupportedException` and `NotImplementedException`, but that is not without cost.

mgravell:
Note: this might need spare consideration on the client vs server paths.

PR #611, opened 2019-10-19
Implement GetBufferWriter() via ArrayBufferWriter<byte> (#609)

mgravell:
Addresses #609 

This changes `DefaultSerializationContext` to (internally) allow an array, or an `IBufferWriter<byte>`. A new `enum` is introduced to avoid confusion over the internal state.

The `IBufferWriter<byte>` is lazily implemented using `ArrayBufferWriter<byte>` the first time (if ever) it is requested, and is then reset and reused for subsequent operations (this means the internal buffer in `ArrayBufferWriter<byte>` can be reused).

The consumption API is changed from `byte[]` to `ReadOnlyMemory<byte>`. In most places this makes no change as the `byte[]` was already being cast to `ReadOnlyMemory<byte>`. A question should be asked, however, about what `GZipStream` (the only compression-stream I could find) does in the `byte[]/int/int` vs `ReadOnlySpan<byte>` use-cases; does it add a copy, or does `GZipStream` directly consume the span? (I will try to investigate). If it (`GZipStream`) does *not* do the right thing, then we might prefer to use `ArraySegment<byte>`, but unfortunately `ArrayBufferWriter<byte>` does not directly expose this (maybe it should?)

I considered - and dismissed - the idea of using a single `object` field that was *either* the array *or* the writer, but this would prohibit full buffer-writer re-use, as every time an array is involved, we'd lose the writer. Having separate fields avoids this. Given the mandated minimum object size, in practice the extra field is free.

Edit: update, [`GZipStream` does "the right thing", so this should not present any issue](https://github.com/dotnet/corefx/blob/967eb3c5cb5fd36c8bb56d06f4c840f6a7a5a754/src/System.IO.Compression/src/System/IO/Compression/GZipStream.cs#L115-L129)

Edit: one additional question might be: should we `Reset()` *after* fully consuming the payload? (currently it resets *before* asking the *next* payload to serialize). The difference is simply: if a `byte[]` is provided, when does it become collectable? We should not `Reset()` before we have finished with all the writes, however. I suspect we'd be *fine*, but ***in theory*** `ArrayBufferWriter<byte>` could hypothetically and legitimately do things like give the memory to the array-pool when `Clear()` is called.

---

Tests are added, but doing so uncovered a second problem (doubly-compiled types, i.e. same full-name in two different assemblies). I have proposed a fix in #612

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -4,7 +4,7 @@
     <CommandLineParserPackageVersion>2.3.0</CommandLineParserPackageVersion>
     <GoogleProtobufPackageVersion>3.10.0</GoogleProtobufPackageVersion>
     <GrpcDotNetPackageVersion>2.23.2</GrpcDotNetPackageVersion>
-    <GrpcPackageVersion>2.23.0</GrpcPackageVersion>
+    <GrpcPackageVersion>2.24.0</GrpcPackageVersion>
     <MicrosoftAspNetCorePackageVersion>3.0.0-preview8.19405.7</MicrosoftAspNetCorePackageVersion>
     <MicrosoftBuildLocatorPackageVersion>1.2.2</MicrosoftBuildLocatorPackageVersion>
     <MicrosoftBuildPackageVersion>16.0.461</MicrosoftBuildPackageVersion>
--- a/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
@@ -59,12 +59,13 @@ namespace Grpc.AspNetCore.Server.Internal
                 GrpcServerLog.SendingMessage(logger);
 
                 var serializationContext = serverCallContext.SerializationContext;
+                serializationContext.Reset();
                 serializer(response, serializationContext);
-
-                if (!serializationContext.TryConsumePayload(out var responsePayload))
+                if (!serializationContext.TryGetPayload(out var responsePayload))
                 {
                     throw new InvalidOperationException("Serialization did not return a payload.");
                 }
+
                 GrpcServerLog.SerializedMessage(serverCallContext.Logger, typeof(TResponse), responsePayload.Length);
 
                 // Must call StartAsync before the first pipeWriter.GetSpan() in WriteHeader
--- a/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
@@ -60,14 +60,11 @@ namespace Grpc.AspNetCore.Server.Internal
 
                 var serializationContext = serverCallContext.SerializationContext;
                 serializer(response, serializationContext);
-                var responsePayload = serializationContext.Payload;
-                serializationContext.Payload = null;
 
-                if (responsePayload == null)
+                if (!serializationContext.TryConsumePayload(out var responsePayload))
                 {
                     throw new InvalidOperationException("Serialization did not return a payload.");
                 }
-
                 GrpcServerLog.SerializedMessage(serverCallContext.Logger, typeof(TResponse), responsePayload.Length);
 
                 // Must call StartAsync before the first pipeWriter.GetSpan() in WriteHeader
@@ -107,7 +104,7 @@ namespace Grpc.AspNetCore.Server.Internal
                 }
 
                 WriteHeader(pipeWriter, responsePayload.Length, isCompressed);
-                pipeWriter.Write(responsePayload);
+                pipeWriter.Write(responsePayload.Span);
 
                 // Flush messages unless WriteOptions.Flags has BufferHint set
                 var flush = canFlush && ((serverCallContext.WriteOptions?.Flags ?? default) & WriteFlags.BufferHint) != WriteFlags.BufferHint;
@@ -480,7 +477,7 @@ namespace Grpc.AspNetCore.Server.Internal
             return false;
         }
 
-        private static bool TryCompressMessage(ILogger logger, string compressionEncoding, CompressionLevel? compressionLevel, Dictionary<string, ICompressionProvider> compressionProviders, byte[] messageData, [NotNullWhen(true)]out byte[]? result)
+        private static bool TryCompressMessage(ILogger logger, string compressionEncoding, CompressionLevel? compressionLevel, Dictionary<string, ICompressionProvider> compressionProviders, ReadOnlyMemory<byte> messageData, [NotNullWhen(true)]out byte[]? result)
         {
             if (compressionProviders.TryGetValue(compressionEncoding, out var compressionProvider))
             {
@@ -489,7 +486,7 @@ namespace Grpc.AspNetCore.Server.Internal
                 var output = new MemoryStream();
                 using (var compressionStream = compressionProvider.CreateCompressionStream(output, compressionLevel))
                 {
-                    compressionStream.Write(messageData, 0, messageData.Length);
+                    compressionStream.Write(messageData.Span);
                 }
 
                 result = output.ToArray();
--- a/src/Grpc.Net.Client/Internal/StreamExtensions.cs
+++ b/src/Grpc.Net.Client/Internal/StreamExtensions.cs
@@ -228,9 +228,7 @@ namespace Grpc.Net.Client
                 // Serialize message first. Need to know size to prefix the length in the header
                 var serializationContext = new DefaultSerializationContext();
                 serializer(message, serializationContext);
-                var data = serializationContext.Payload;
-
-                if (data == null)
+                if (!serializationContext.TryConsumePayload(out var data))
                 {
                     throw new InvalidOperationException("Serialization did not return a payload.");
                 }
@@ -269,7 +267,7 @@ namespace Grpc.Net.Client
             }
         }
 
-        private static byte[] CompressMessage(ILogger logger, string compressionEncoding, CompressionLevel? compressionLevel, Dictionary<string, ICompressionProvider> compressionProviders, byte[] messageData)
+        private static byte[] CompressMessage(ILogger logger, string compressionEncoding, CompressionLevel? compressionLevel, Dictionary<string, ICompressionProvider> compressionProviders, ReadOnlyMemory<byte> messageData)
         {
             if (compressionProviders.TryGetValue(compressionEncoding, out var compressionProvider))
             {
@@ -278,7 +276,7 @@ namespace Grpc.Net.Client
                 var output = new MemoryStream();
                 using (var compressionStream = compressionProvider.CreateCompressionStream(output, compressionLevel))
                 {
-                    compressionStream.Write(messageData, 0, messageData.Length);
+                    compressionStream.Write(messageData.Span);
                 }
 
                 return output.ToArray();
--- a/src/Grpc.Net.Client/Internal/StreamExtensions.cs
+++ b/src/Grpc.Net.Client/Internal/StreamExtensions.cs
@@ -227,8 +227,9 @@ namespace Grpc.Net.Client
 
                 // Serialize message first. Need to know size to prefix the length in the header
                 var serializationContext = new DefaultSerializationContext();
+                serializationContext.Reset();
                 serializer(message, serializationContext);
-                if (!serializationContext.TryConsumePayload(out var data))
+                if (!serializationContext.TryGetPayload(out var data))
                 {
                     throw new InvalidOperationException("Serialization did not return a payload.");
                 }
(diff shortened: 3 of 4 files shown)


JamesNK:
Hmmm, the goal I was aiming for with this was to be able to write directly to the response pipeline (unless compression is enabled).

Because `GetBufferWriter` is called without the message size then that isn't possible. I never had time to test this properly end to end ☹️

Before grpc-dotnet uses the buffer writer I think `SerializationContext.GetBufferWriter(int payloadLength)` is needed. It could replace the existing `GetBufferWriter()` overload if the break is ok. The writer would then ensure the message size is prefixed (in non-compression writes)

JamesNK:
Does protobuf-net have a way to calculate the message size without serialization? Grpc.Tools includes a CalculateSize method, which is why this can work when using Google.Protobuf.

You might need to continue to serialize to an array, pass the length to GetBufferWriter, then write the array to the returned writer.

mgravell:
V3 (the only version to support this API) has a measure method that does everything as cheaply as possible and returns the length and some transient state that makes serialization much cheaper when done "for real" (basically, it includes a lookup of all the object lengths it calculated the first time).

If I need to transition to an API change that includes the length in advance: not a huge problem.

We hit very similar problems with our raw http usage; still need that pesky `Content-Length` header.

mgravell:
> Grpc.Tools includes a CalculateSize method, which is why this can work when using Google.Protobuf.

The API for sending a length [doesn't exist](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core.Api/SerializationContext.cs).

This leads me to suspect one of two things:

1. It works for "regular" because the native version is just handing out unmanaged buffer rentals, then looking at the length *afterwards* (similar to what we're doing here, but zero copy)
2. It hasn't been tested in earnest on "regular" yet

FWIW, me/Tim/Ben kept telling folks that we wanted to be able to stitch external buffers into an outbound pipe for exactly this reason (to avoid the extra copy). I can't remember the exact words during the API review we were in, but it could be paraphrased:

> There are a lot of experts in this room, and we can't see why you'd need that.

mgravell:
> Unfortunately on the client the buffer write still won't be able to write directly to the request stream.

I'm not sure I fully understand the difference here client vs server. Can you elaborate on why the different rules?

> Stream.WriteAsync returns Task and serialization isn't async.

Would it help if it was? I've overhauled the internals specifically to allow for this possibility. I'm still waiting on Fowler to get a new API approved that allows me to target it as an abstraction, but: those wheels are in motion. I can pull out the API proposal if it would help, but the tl;dr would be

> make the pipelines shape (re `FlushAsync` etc) available alongside `IBufferWriter<byte>`, and have pipelines implement that API

JamesNK:
> The API for sending a length [doesn't exist](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core.Api/SerializationContext.cs).

Yes, it needs something like:

```cs
public virtual IBufferWriter<byte> GetBufferWriter(int payloadLength)
{
    throw new NotImplementedException();
}
```

Then Grpc.Tools generated code can look something like:

```cs
public void ContextualSerialize<T>(T message, SerializationContext context)
{
    var bufferWriter = context.GetBufferWriter(message.CalculateSize());
    var output = new CodedOutputWriter(bufferWriter);
    message.WriteTo(output);
}
```

mgravell:
My concern there is that Grpc.Core might say "I'm, no, *we* don't need that, and forcing the consumer to measure first adds overhead to our usage". So: should be a fun conversation (orders popcorn)

Side note: I already have one API change open on that API, re making it feaure-testable for consumers.

Here's the thing: the *abaolute worst outcome* (for me) would be if I had to measure to call that API, and then the API said "nope, we don't support that; use arrays".

JamesNK:
This is the serialization I mean that isn't async: https://github.com/grpc/grpc/blob/a1ad4800da2bb02c843bdaea8fb06518916b6fba/src/csharp/Grpc.Core.Api/Marshaller.cs#L58

mgravell:
> This is the serialization I mean that isn't async

Gotcha; I take it sync IO is disabled (Kestrel-style) here, then?

JamesNK:
This is fine on the server. The server writes the Protobuf data to a buffer writer, and once it is complete `Grpc.AspNetCore.Server` will flush it if necessary.

It is more complicated on the client because we're ultimately writing to the request stream via `Stream.WriteAsync`. Implementing an IBufferWriter<byte> that writes to a stream with async isn't possible because the serialization funcs aren't async.

The simple solution in the client is to continue to write to `ArrayBufferWriter<byte>`, then copy the data from it to the stream. Your PR is already doing this.

JamesNK:
> My concern there is that Grpc.Core might say "um, no, we don't need that, and forcing the consumer to measure first adds overhead to our usage". So: should be a fun conversation (orders popcorn)

I don't think this is a problem. `message.CalculateSize()` is already always used in gRPC to calculate the size of the array created: https://github.com/protocolbuffers/protobuf/blob/94cbf009e934d45ee6fc1441ff9b56df9b314b3b/csharp/src/Google.Protobuf/MessageExtensions.cs#L91-L104

mgravell:
Note : it is also pretty trivial to turn an output `Stream` into an `IBufferWriter<byte>`, *as long as* `Write` (rather than `WriteAsync`) works. This isn't zero copy, but it avoids having to serialize everything into one single block (essentially, you lease an array-pool buffer, and only write it to the stream whenever it gets too full; not on every `Advance`). Let me know if you'd be interested in that as a PR.

mgravell:
Observation, though: the only API you currently support is the `byte[]`, which forces buffering.

Hence, you lose nothing by taking this PR as a stop-gap, since it has the *exact same buffering intermediate*, but fills in the API hole. 

The code can always be updated to (whatever length-in-advance-API) *when that exists*.

JamesNK:
Your still serializing to an `ArrayBufferWriter<byte>` in Grpc.AspNetCore though. You don't see an array being allocated in your code, but it still happens.

And if we're thinking about changing the API in Grpc.Core.Api then it doesn't seem like a good idea to depend on it for no benefit. I will create a PR in Grpc.Core.Api to provide the length for the buffer. We'll see if that change can get included in the next version.

JamesNK:
> Note : it is also pretty trivial to turn an output `Stream` into an `IBufferWriter<byte>`, *as long as* `Write` (rather than `WriteAsync`) works. This isn't zero copy, but it avoids having to serialize everything into one single block (essentially, you lease an array-pool buffer, and only write it to the stream whenever it gets too full; not on every `Advance`). Let me know if you'd be interested in that as a PR.

I'd like to do that, but it relies on the `HttpClient` request stream from `SerializeToStreamAsync` allowing non-async writes.

https://github.com/grpc/grpc-dotnet/blob/b43a6fc37e22a726b9e64aab002603211f31c7dd/src/Grpc.Net.Client/Internal/PushUnaryContent.cs#L43

https://github.com/grpc/grpc-dotnet/blob/b43a6fc37e22a726b9e64aab002603211f31c7dd/src/Grpc.Net.Client/Internal/StreamExtensions.cs#L259-L261

JamesNK:
https://github.com/dotnet/corefx/blob/15abdfc3816352c64098735b79a98dd97524ee36/src/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpBaseStream.cs#L100-L108

no bueno

mgravell:
> You're still serializing to an `ArrayBufferWriter<byte>` in Grpc.AspNetCore though. You don't see an array being allocated in your code, but it still happens.

If you're talking about the one *in this code*, we must compare:

1. an `ArrayBufferWriter<byte>` that is allocated at most once per `DefaultSerializationContext` (which is recycled between uses), and which may hold an **oversized** array - so: actual array allocations will be very rare
2. an exact-sized `byte[]` that is allocated per-serialize

Then option 1 is **absolutely** much lower allocation than option 2.

If you're saying there's *another* writer somewhere that I haven't seen: that doesn't change a single thing about the allocations above :)

The point is that the PR allows us to avoid *almost all* array allocations associates with serialization, by amortizing the backing-buffer over many, many operations.

chwarr:
> My concern there is that Grpc.Core might say "um, no, we don't need that, and forcing the consumer to measure first adds overhead to our usage". So: should be a fun conversation (orders popcorn)

The [Bond-over-gRPC C# implementation in one such implementation that doesn't need to measure the size before starting serialization][bond-marshaller]. (Bond's Compact Binary format doesn't length prefix embedded message fields like Protobuf does, so its serialization is single-pass.)

We haven't had time to switch to the new C# serialization APIs from Grpc.Core. We're also waiting to see if some of the more advanced methods end up in the interface.

> FWIW, me/Tim/Ben kept telling folks that we wanted to be able to stitch external buffers into an outbound pipe for exactly this reason (to avoid the extra copy).

If there were a scatter/gather API on `Span<byte>`/`Memory<byte`> instances, then the Bond serialization code for gRPC could be made very efficient. It would accumulate a collection of spans over chunky buffers. When it encountered an embedded blob/pre-serialized message in its payload, it would just add that span without making a copy. Then, they can all get spliced into the output together.

[bond-marshaller]: https://github.com/microsoft/bond/blob/9c3d566032fbbfef835a7dc0c7ea015edebacdf5/cs/src/grpc/Marshaller.cs#L32-L48

JamesNK:
> Bond-over-gRPC C# implementation in one such implementation that doesn't need to measure the size before starting serialization

This length prefix for the overall gRPC message. How you serialize the message content is up to you.

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md - see `Length-Prefixed-Message`

JamesNK:
> The point is that the PR allows us to avoid almost all array allocations associates with serialization, by amortizing the backing-buffer over many, many operations.

Ok, but I want to wait a bit to figure out `SerializationContext` first. Patience 😄

mgravell:
> would accumulate a collection of spans over chunky buffers. 

It would have to be "memories" - spans are incredibly hard to accumulate, due to the ref-struct semantics. But I understand what you mean. No API for this exists (or is likely to ever exist, IMO) so : you would have to add a copy step.

chwarr:
> The length prefix is for the overall gRPC message.

Understood and agreed.

My [comment](https://github.com/grpc/grpc-dotnet/pull/611#issuecomment-544382614) was more about the ` IBufferWriter<byte> GetBufferWriter(int payloadLength)` proposal. If the `payloadLength` is required to be exactly correct, then it feels like this will penalize serializers that don't know their size ahead of time. If it were, instead more of a `sizeHint`, then a serializer that knows its size ahead of time cheaply (e.g., Protobuf) can use this to hint the right size. For other serializers, they'd call `GetBufferWriter(someGuessAtTheFinalSize)` and then call the usual `IBufferWriter<>.GetMemory()`/`IBufferWriter<>.GetSpan()` and `IBufferWriter<>.Advance()` while writing. This would need to buffer to know the payload size to write the length prefix.

This is perhaps getting a bit off topic for this PR. If you want to move this to an issue, let me know.

JamesNK:
Lets break this down:
1. There is value in serializing to a buffer writer instead of an array because the SerializeContext/BufferWriter can be reused (all serializers)
2. There is value in a serializer being able to specify the payload size before serializing, because then the writer could write directly to the gRPC response (Google.Protobuf, maybe protobuf-net)
3. Not all serializers will know the message size in advance. These serializers will always need to write their content to an in-memory buffer. Then the gRPC framework will need to get the length of content written, write it to the response, then copy the message to the response (Bond)

I think the best solution here is to leave `GetBufferWriter` as it is and add `SetPayloadLength` to the context:

```cs
public virtual void SetPayloadLength(int payloadLength)
{
}
```

Serializers who know their length prior to serialization can call it before calling GetBufferWriter. For them, Grpc.AspNetCore.Server can write the length to the response and return `Response.BodyWriter` as the buffer. For other serializers that don't set the length first (Bond), they can serialize to ArrayBufferWriter, like what this PR is doing currently.

Yay/nay?

mgravell:
> Yay/nay?

That would get a Yay from me. Additional observations:

- it doesn't shatter the existing API, which is always good; existing code that doesn't specify a length in advance continues to work in all scenarios
- I can potentially make protobuf-net v2 and v3 use different paths (they have different capabilities; I detect between them)

mgravell:
spiked it (commit above); viable for me...

chwarr:
LGTM.

Some alternative signatures I came up with as food for thought. None of these seem materially better than the `void SetPayloadLength(int payloadLength)`. As an analog, `HttpContext` uses a nullable ContentLength property.

* `IBufferWriter<> SetPayloadLength(int payloadLength)`: returned buffer writer is the one to use.
* `IBufferWriter<> GetNonBufferedWriter(int payloadLength)`: I don't like that this so explicitly calls out the buffered/non-buffered distinction, but that's a way of looking at the differences
* `void SwitchToNonBuffered(int payloadLength)`: explicit that there a buffered/non-buffered switch occuring.

JamesNK:
Replaced by https://github.com/grpc/grpc-dotnet/pull/629

PR #612, opened 2019-10-19
Add tests for #611

mgravell:
This is a sub-branch from #611; adds tests for `DefaultSerializationContext` (#611)

The reason this is split into a separate PR is that I discovered that `DefaultSerializationContext` and `DefaultDeserializationContext` were being compiled into two separate dlls; this ... is a fundamentally terrible idea, IMO, and it makes it a: incredibly hard to test (needs `extern alias` etc), and b: quite brittle (this can *really* upset the runtime in a few places).

My proposal here is to move these down to `Grpc.Net.Common`, fixing both problems.


JamesNK:
I don't want to use InternalsVisibleTo. I've had bad experiences with it in the past.

A simpler solution is to leave the duplication and test in a project that doesn't reference the client and server.

mgravell:
You call. FWIW, I've had problems with doubly-defined types , yet have never had any problems with `[Internals visible to]` :man_shrugging:

JamesNK:
Replaced by https://github.com/grpc/grpc-dotnet/pull/629

Issue #602, opened 2019-10-12
Investigate gRPC compression performance


JamesNK:
I think we might be able to optimize compression to avoid additional array allocations.

zsybupt:
```c#

                if (output.TryGetBuffer(out ArraySegment<byte> buffer))
                {
                    if (buffer.Offset == 0 && buffer.Count == buffer.Array?.Length)
                    {
                        result = buffer.Array;
                        return true;
                    }
                }

                result = output.ToArray();
                return true;
```

zsybupt:
if (output.TryGetBuffer(out ArraySegment<byte> buffer))
                {
                    if (buffer.Offset == 0 && buffer.Count == buffer.Array?.Length)
                    {
                        result = buffer.Array;
                        return true;
                    }
                }

                result = output.ToArray();
                return true;

PR #615, opened 2019-10-21
Add compression benchmark and refactor serialization

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/602

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/ReadOnlySequenceStream.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/ReadOnlySequenceStream.cs
@@ -1,0 +1,167 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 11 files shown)


zsybupt:
I think we can use the same implementation for compress message

JamesNK:
Sure. There are going to be some changes to serialization in the next day or two, I'll look at it at the same time.

src/Grpc.AspNetCore.Server/Internal/ReadOnlySequenceStream.cs
```diff
@@ -0,0 +1,167 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    // Potentially remove in the future when https://github.com/dotnet/corefx/issues/31804 is implemented
+    internal class ReadOnlySequenceStream : Stream
+    {
+        private static readonly Task<int> TaskOfZero = Task.FromResult(0);
```
JunTaoLuo: const?

JamesNK: `Task` not a primitive type so const is not allowed.

Issue #608, opened 2019-10-18
Add using block for Created Compression Stream


zsybupt:
Hi,
I find that compression stream created at [PipeExtensions](https://github.com/grpc/grpc-dotnet/blob/b43a6fc37e22a726b9e64aab002603211f31c7dd/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs#L472) does not got disposed after used, and I have checked the implemetation of GZipStream in corefx, it is indeed safe to do decompression without disposing the stream. But I think it may be better to dispose the stream to make it decouple from the detail implemtation of GZipStream, if the implementation changes, there may be a memory leak.

zsybupt:
I can provide a pull request if it is welcome

JamesNK:
A bunch of changes are happening to serialization/deserialization at the moment. I'll look at this at the same time as those changes are going on.

PR #629, opened 2019-10-26
Implement SerializationContext buffer writer

### LINKED ISSUE #609: implement SerializationContext.GetBufferWriter

JamesNK:
Fixes #609
Fixes https://github.com/grpc/grpc-dotnet/issues/608

@mgravell

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
@@ -1,0 +1,237 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System;
+using System.Buffers;
+using System.Buffers.Binary;
+using System.Diagnostics;
+using System.IO;
+using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+using Grpc.Core;
+using Grpc.Net.Compression;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal sealed class HttpContextSerializationContext : SerializationContext
+    {
+        private static readonly Status SendingMessageExceedsLimitStatus = new Status(StatusCode.ResourceExhausted, "Sending message exceeds the maximum configured message size.");
+
+        private readonly HttpContextServerCallContext _serverCallContext;
+        private InternalState _state;
+        private int? _payloadLength;
+        private ICompressionProvider? _compressionProvider;
+        private ArrayBufferWriter<byte>? _bufferWriter;
+
+        public PipeWriter ResponseBufferWriter { get; set; } = default!;
+
+        private bool DirectSerializationSupported => _compressionProvider == null && _payloadLength != null;
+
+        public HttpContextSerializationContext(HttpContextServerCallContext serverCallContext)
+        {
+            _serverCallContext = serverCallContext;
+        }
+
+        private enum InternalState : byte
+        {
+            Initialized,
+            CompleteArray,
+            IncompleteBufferWriter,
+            CompleteBufferWriter,
+        }
+
+        public void Reset()
+        {
+            _compressionProvider = ResolveCompressionProvider();
+
+            _payloadLength = null;
+            if (_bufferWriter != null)
+            {
+                // Reuse existing buffer writer
+                _bufferWriter.Clear();
+            }
+            _state = InternalState.Initialized;
+        }
+
+        private ICompressionProvider? ResolveCompressionProvider()
+        {
+            Debug.Assert(
+                _serverCallContext.ResponseGrpcEncoding != null,
+                "Response encoding should have been calculated at this point.");
+
+            var canCompress =
+                GrpcProtocolHelpers.CanWriteCompressed(_serverCallContext.WriteOptions) &&
+                !string.Equals(_serverCallContext.ResponseGrpcEncoding, GrpcProtocolConstants.IdentityGrpcEncoding, StringComparison.Ordinal);
+
+            if (canCompress)
+            {
+                if (_serverCallContext.MethodContext.CompressionProviders.TryGetValue(_serverCallContext.ResponseGrpcEncoding, out var compressionProvider))
+                {
+                    return compressionProvider;
+                }
+            }
+
+            return null;
+        }
+
+        public override void SetPayloadLength(int payloadLength)
+        {
+            switch (_state)
+            {
+                case InternalState.Initialized:
+                    _payloadLength = payloadLength;
+                    break;
+                default:
+                    ThrowInvalidState(_state);
+                    break;
+            }
+        }
+
+        public override void Complete(byte[] payload)
+        {
+            switch (_state)
+            {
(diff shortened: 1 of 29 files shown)


JamesNK:
Before:
```
---------------- |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 HandleCallAsync | 2.794 us | 0.0533 us | 0.0654 us | 357,894.2 |      0.2098 |           - |           - |               880 B |
```

After:
```
------------------------- |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 PipelinesHandleCallAsync | 2.472 us | 0.0296 us | 0.0262 us | 404,545.4 |      0.0687 |           - |           - |               296 B |
```

mgravell:
Wow, it looks like you implemented option 3; gotta say: I'm impressed. Way to over-deliver!

src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
```diff
@@ -0,0 +1,232 @@
...
+                    {
+                        Debug.Assert(_payloadLength != null, "A payload length is required for direct serialization.");
+
+                        EnsureMessageSizeAllowed(_payloadLength.Value);
+
+                        WriteHeader(ResponseBufferWriter, _payloadLength.Value, compress: false);
+                    }
+
+                    _state = InternalState.IncompleteBufferWriter;
+                    goto case InternalState.IncompleteBufferWriter;
```
JunTaoLuo: Woah I haven't seen a goto in a long time. Isn't the syntax the same without the goto (i.e. no break)?

src/Shared/DefaultSerializationContext.cs
```diff
@@ -16,17 +16,142 @@
...
 
 namespace Grpc.Shared
 {
     internal sealed class DefaultSerializationContext : SerializationContext
     {
-        public byte[]? Payload { get; set; }
+        private byte[]? _array;
+        private InternalState _state;
+
+        public bool DirectSerializatedSupported { get; internal set; }
```
JunTaoLuo: DirectSerializationSupported

src/Shared/DefaultSerializationContext.cs
```diff
@@ -16,17 +16,142 @@
...
+        public int? PayloadLength { get; set; }
+        public IBufferWriter<byte>? BufferWriter { get; internal set; }
+
+        private enum InternalState
```
JunTaoLuo: Curious if we should share the states with HttpContextSerializationContext.cs? These look the same.

src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
jtattermusch: also not a big fan of the goto

src/Shared/DefaultSerializationContext.cs
```diff
@@ -16,17 +16,142 @@
...
+using System.Buffers;
+using System.Buffers.Binary;
+using System.Runtime.CompilerServices;
 using Grpc.Core;
 
 namespace Grpc.Shared
 {
     internal sealed class DefaultSerializationContext : SerializationContext
     {
-        public byte[]? Payload { get; set; }
```
jtattermusch: The logic of DefaultSerializationContext looks very similar to HttpContextSerializationContext. It would be nice if we could somehow deduplicate.

src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
```diff
@@ -0,0 +1,232 @@
...
+using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+using Grpc.Core;
+using Grpc.Net.Compression;
+
+namespace Grpc.AspNetCore.Server.Internal
+{
+    internal sealed class HttpContextSerializationContext : SerializationContext
+    {
+        private static readonly Status SendingMessageExceedsLimitStatus = new Status(StatusCode.ResourceExhausted, "Sending message exceeds the maximum configured message size.");
```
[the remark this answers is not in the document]
JamesNK: It's a struct. You can't have const structs.

src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
```diff
@@ -0,0 +1,232 @@
...
+                        Debug.Assert(_bufferWriter != null, "Buffer writer has been set to get to this state.");
+
+                        var data = _bufferWriter.WrittenSpan;
+
+                        GrpcServerLog.SerializedMessage(_serverCallContext.Logger, typeof(object), data.Length);
+                        WriteMessage(data);
+                    }
+                    else
+                    {
+                        GrpcServerLog.SerializedMessage(_serverCallContext.Logger, _serverCallContext.MethodContext.ResponseType, _payloadLength.GetValueOrDefault());
```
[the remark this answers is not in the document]
JamesNK: I chose not to because this code will only be overridden by frameworks. They should test they are doing the right thing themselves using unit/functional tests.

Checking every serialization feels wasteful.

src/Shared/DefaultSerializationContext.cs
JamesNK: I tried to come up with a shared context but `HttpContextSerializationContext` is deeply integrated with `HttpContext`, and that type doesn't exist in the client.

src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
JamesNK: No. In C# you must have a break.

src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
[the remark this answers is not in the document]
JamesNK: I'm going to commit this as it is. If you think this is important, or there are bug reports around this issue after release, then it can be added later by wrapping the buffer writer returned.

src/Shared/DefaultSerializationContext.cs
JamesNK: I don't think it is worth sharing one enum.

ejona86:
This seems to have broken compatibility with grpc-java. I think it is because the gzip stream isn't "finish"ed. We can discuss on https://github.com/grpc/grpc/issues/20884.

src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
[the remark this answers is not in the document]
JunTaoLuo: Ah, forgot it's a struct.

src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
JunTaoLuo: TIL you can only fall through if the case is empty.

Issue #671, opened 2019-11-27
Add support for non-conforming containers by adding a public IGrpcServiceActivator abstraction


dotnetjunkie:
The current implementation of grpc-dotnet takes a hard dependency on the framework's `IServiceProvider` abstraction. This is, unfortunately, a problem for the so-called 'non-conforming' DI Containers such as Simple Injector, Ninject, Castle Windsor, and others. Those containers do not allow replacing the built-in MS.DI container (as DI Containers such as Autofac and Lamar do), but instead rely on working side-by-side with the MS.DI Container. Those libraries and their users rely on being able to hook into interception points that are defined by the framework. ASP.NET Core, for instance, has well-defined interception points such as `IControllerActivator`,`IViewComponentActivator`, and `ITagHelperActivator`.

Because of the lack of such publically defined interception point for the creation of gRPC services, it becomes impractical for those Containers and their users to integrate with grpc-dotnet in such way that their DI Container of choice is in control over the construction and lifetime management of these services.

grpc-dotnet does contain an `IGrpcServiceActivator<TService>` abstraction. This interface is a good starting point but, unfortunately, it is `internal`, making it impossible for anyone to replace it. Making it `public` as-is, however, has a few downsides of its own, such as:

* It exposes on another internal type, namely `GrpcActivatorHandle<TService>`. But that type seems to exist for internal purposes only. Exposing that would introduce unneeded complexity.
* `IGrpcServiceActivator<TService>` is a generic abstraction. Although the generic abstraction makes sense now it is internal, it would complicate creating implementations for third-parties. Ideally, a user should be able to replace the default implementation by calling `services.AddSingleton(new MyCustomGrpcServiceActivator(container))`. 

Instead, I would like to suggest:

* Making `IGrpcServiceActivator` public; this allows it to become an interception point.
* Making `IGrpcServiceActivator` a non-generic type; this simplifies creating custom implementations
* Simplifying `IGrpcServiceActivator`'s API to make most sense for the consumer; Mimicking the common abstractions defined by ASP.NET Core (such as `IControllerActivator`) makes a lot of sense here. `I<type>Activator` has become a common pattern within ASP.NET Core, and mimicking this API makes it easier for anyone who work with an activator of some sort.
* Splitting the current default implementation into two separate implementations, identical to what ASP.NET Core does with its controllers:
[...]

PR #677, opened 2019-12-03
Make service and interceptor activators public

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/671


davidfowl:
@JamesNK might be worthwhile adding a test with a dummy implementation (like a fake container).

jtattermusch:
I'm ok with this, but if we do that, we are committing to API stability for the interfaces, so I'd like us to think twice. @JamesNK  @JunTaoLuo

JamesNK:
The API shape has been debated a lot. I'll get the original requester of this feature to double check it works for them before the package is finalized.

Issue #99, opened 2019-02-15
Add support for grpc web


davidfowl:
grpcWeb is a different protocol to grpc over HTTP/2. At some point we should add support for this.

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2

ctaggart:
I want to be able to consume gRPC services from Blazor. You can read [The state of gRPC in the browser](https://grpc.io/blog/state-of-grpc-web/). It is pretty easy to set up a gRPC-web proxy for a gRPC service, so it would be nice to be able to servce gRPC-web, but we can get by without out. What we really need is the gRPC-web client.

JamesNK:
It would involve writing a custom `CallInvoker`, similar to the WIP managed .NET client.

There is nothing stopping someone writing one today in .NET Core 2.2. gRPC-web uses HTTP/1 tech which is fully supported by HttpClient.

joshlang:
I vote for this feature.

We intend to do a simple protobuf-over-websocket thing in our blazor app.  I'm insisting we call it `GhettoRPC`.  It's relatively easy because we don't absolutely need advanced gRPC features.  But it *is* rather silly, and we'd prefer grpcWeb.

razfriman:
Would love support for this as well. I came up with a grpc-gateway-like solution where I convert calls so I could reach my api via Blazor WASM. 

For anyone interested here is my project: https://github.com/razfriman/BlazorGrpc

JamesNK:
Server in-process proxy guidance: https://github.com/grpc/grpc-web/blob/master/IN-PROCESS-PROXY.md

Rora:
For anyone looking to get a working gRPC server & client Blazor wasm app, I've created a [simple proof of concept](https://github.com/Rora/aspnetcore-apis/tree/master/BlazorGrpcCodeFirst). The PoC uses a gRPC-web proxy hosted in a aspnet core service ([KnowitSolutions](https://github.com/KnowitSolutions/Grpc.Helpers) created the gRPC-web proxy), and consumes it in the blazor wasm app using a basic `CallInvoker` (created by [SteveSandersonMS](https://github.com/SteveSandersonMS/presentation-2019-10-NDCSydney/tree/master/demo/BlazorMart/BlazorMart.Client)).

The example does use a grpc code first libraries, but is pretty easy to convert to the contract first setup. From the top of my head; 
[...]

PR #695, opened 2019-12-14
gRPC-Web client and server

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/99

Implementation of https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md for ASP.NET Core server and .NET Core gRPC client.


src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
```diff
@@ -0,0 +1,170 @@
...
+            var innerResult = await _inner.ReadAsync(cancellationToken);
+            if (innerResult.Buffer.IsEmpty)
+            {
+                _currentDecodedBuffer = innerResult.Buffer;
+                _currentInnerBuffer = innerResult.Buffer;
+                return innerResult;
+            }
+
+            // Minimum valid base64 length is 4. Read until we have at least that much content
+            while (innerResult.Buffer.Length < 4)
```
halter73: For CancelPendingRead to work, we should check every ReadResult and check if it's canceled and return to the caller if it is. Otherwise, the developer using the Base64PipeReader might call CancelPendingRead, but never see a canceled read result. In the case we receive a canceled ReadResult before reading the minimum about of encoded data, I would make sure to return an empty-but-still-canceled ReadResult.

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
```diff
@@ -0,0 +1,170 @@
...
+
+            _inner.AdvanceTo(consumedPosition, examinedPosition);
+        }
+
+        private SequencePosition ResolvePosition(SequencePosition base64Position)
+        {
+            // TODO - is there a better way to get the index of a position in a sequence?
+            var length = _currentDecodedBuffer.Slice(0, base64Position).Length;
+
+            var resolvedPosition = _currentInnerBuffer.GetPosition(Base64.GetMaxEncodedToUtf8Length((int)length));
```
halter73: Could GetMaxEncodedToUtf8Length ever overestimate the number of bytes that were consumed in the _currentInnerBuffer because the the base 64 encoding happened to be more efficient than it could have been given a different payload of the same length? Or is Base64.GetMaxEncodedToUtf8Length guaranteed to be exact? That latter seems unlikely to me given the name.

JamesNK: It will divide by 3 (with roundup for base64 padding) then multiply by 4.

You're right that this could potentially resolve to a higher number because of the roundup, but in practice it won't happen because of how gRPC uses the PipeReader. ReadAsync will always return bytes in multiples of 3 (until the very end of the message), and PipeReader.Advance will always be called with ReadResult.Buffer.End. The only time roundup will happen is at the end of the message which is where we want the roundup.

I'll add a note that this is an assumption/limitation of the reader.

src/Grpc.AspNetCore.Web/Internal/GrpcWebProtocolHelpers.cs
```diff
@@ -0,0 +1,116 @@
...
+    internal static class GrpcWebProtocolHelpers
+    {
+        private const byte Cr = (byte)'\r';
+        private const byte Lf = (byte)'\n';
+        private const byte Colon = (byte)':';
+        private const byte Space = (byte)' ';
+
+        // Special trailers byte with eigth most significant bit set.
+        // Parsers will use this to identify regular messages vs trailers
+        private static readonly byte Trailers = 0x80;
```
Tratcher:
```suggestion
        private static readonly byte TrailersSignfier = 0x80;
```

src/Grpc.AspNetCore.Web/Internal/GrpcWebProtocolHelpers.cs
```diff
@@ -0,0 +1,116 @@
...
+            return total;
+        }
+
+        private static void WriteTrailersContent(Span<byte> buffer, IHeaderDictionary trailers)
+        {
+            var currentBuffer = buffer;
+
+            foreach (var kv in trailers)
+            {
+                foreach (var value in kv.Value)
```
Tratcher: You haven't actually verified at any point before this that the key and value only contain ASCII, or that there are no prohibited ASCII characters like control characters. The servers each have their own IHeaderDictionary implementation that enforces this.

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
JamesNK: Done

src/Grpc.AspNetCore.Web/Internal/GrpcWebFeature.cs
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            _trailers = new HeaderDictionary();
+        }
+
+        public PipeReader Reader => _pipeReader ?? _initialRequestFeature.Reader;
+
+        public PipeWriter Writer => _pipeWriter ?? _initialResponseFeature.Writer;
+
+        public Stream Stream => _initialResponseFeature.Stream;
```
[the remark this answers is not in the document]
JamesNK: gRPC doesn't expose either. You can get access to the HttpContext via an extension method but if you start writing directly to response Stream or BodyWriter then you're on your own.

I'll add a note though.

src/Grpc.AspNetCore.Web/Internal/GrpcWebMiddleware.cs
```diff
@@ -0,0 +1,157 @@
...
+
+                Log.GrpcWebRequestNotProcessed(_logger);
+            }
+
+            return _next(httpContext);
+        }
+
+        private async Task HandleGrpcWebRequest(HttpContext httpContext, GrpcWebMode mode)
+        {
+            var feature = SetGrpcWebFeature(mode, httpContext);
```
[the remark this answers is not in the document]
JamesNK: I had considered that. Because the gRPC-Web middleware changes the content type, if the gRPC-Web middleware runs again then it will no-op.

Are there other reasons to undo setting the features? Or is it considered something you should do to be a good middleware citizen?

src/Grpc.AspNetCore.Web/Internal/GrpcWebMiddleware.cs
```diff
@@ -0,0 +1,157 @@
...
+                    httpContext.Response.ContentType = responseContentType;
+                    Log.SendingGrpcWebResponse(_logger, responseContentType);
+                }
+
+                return Task.CompletedTask;
+            });
+
+            await _next(httpContext);
+
+            await feature.WriteTrailers();
```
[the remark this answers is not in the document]
JamesNK: If it is a gRPC-Web call then it needs to output trailers. The only time it doesn't is if the response has already been completed with CompleteAsync.

src/Grpc.AspNetCore.Web/Internal/GrpcWebFeature.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public IHeaderDictionary Trailers
+        {
+            get => _trailers;
+            set { _trailers = value; }
+        }
+
+        public async Task CompleteAsync()
+        {
+            // TODO(JamesNK): When CompleteAsync is called from another thread (e.g. deadline exceeded),
```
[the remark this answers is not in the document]
JamesNK: As middleware we can't add locks to the server gRPC library.

Could locking inside PipeWriter work? e.g. lock to prevent new calls to GetSpan/GetMemory/Advance from happening? But that wouldn't help in situations where gRPC server was halfway through writing a message.

src/Grpc.AspNetCore.Web/GrpcWebServiceExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <param name="services">The <see cref="IServiceCollection"/> for adding services.</param>
+        /// <returns>The <see cref="IServiceCollection"/> so that additional calls can be chained.</returns>
+        public static IServiceCollection AddGrpcWeb(this IServiceCollection services)
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            services.TryAddSingleton<GrpcWebMarkerService>();
```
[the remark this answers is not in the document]
JamesNK: Right now all it does is optionally set options. Requiring `AddGrpcWeb()` probably isn't needed.

Do you know of other middleware/plugins that only use AddXXX to set configuration that are optional?

src/Grpc.AspNetCore.Web/Internal/GrpcWebProtocolHelpers.cs
JamesNK: What validation do you suggestion?

Something like this...? https://github.com/aspnet/AspNetCore/blob/4dd9bfc492452174f9329c3e1fc4807cb393c05a/src/Servers/Kestrel/Core/src/Internal/Http/HttpHeaders.cs#L257-L276

src/Grpc.AspNetCore.Web/Internal/GrpcWebFeature.cs
```diff
@@ -0,0 +1,94 @@
...
+            _initialRequestFeature = httpContext.Features.Get<IRequestBodyPipeFeature>();
+            _initialResponseFeature = httpContext.Features.Get<IHttpResponseBodyFeature>();
+
+            if (grpcWebMode == GrpcWebMode.GrpcWebText)
+            {
+                _pipeReader = new Base64PipeReader(_initialRequestFeature.Reader);
+                _pipeWriter = new Base64PipeWriter(_initialResponseFeature.Writer);
+            }
+
+            _trailers = new HeaderDictionary();
```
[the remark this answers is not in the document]
JamesNK: If there are using the `Metadata` collection to add trailers then it has validation (alphanumeric plus a very limited set of characters). Probably a good idea to validate again at this level.

src/Shared/CommonGrpcProtocolHelpers.cs
```diff
@@ -0,0 +1,58 @@
...
+
+#endregion
+
+using System;
+
+namespace Grpc.Shared
+{
+    internal static class CommonGrpcProtocolHelpers
+    {
+        public static bool IsContentType(string contentType, string? s)
```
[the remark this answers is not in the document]
JamesNK: I don't need all of its functionality. And this helper method avoids allocating it.

src/Grpc.AspNetCore.Web/Internal/GrpcWebFeature.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public Task SendFileAsync(string path, long offset, long? count, CancellationToken cancellationToken) =>
+            _initialResponseFeature.SendFileAsync(path, offset, count, cancellationToken);
+
+        public Task StartAsync(CancellationToken cancellationToken) =>
+            _initialResponseFeature.StartAsync(cancellationToken);
+
+        public Task WriteTrailers()
+        {
+            if (!_isComplete)
```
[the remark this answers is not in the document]
JamesNK: WriteTrailers is called once next has complete. This check is to make it no-op there if CompleteAsync has already been called. I guess this should also be checked inside CompleteAsync incase someone calls it twice.

The current code should support the current gRPC server behavior, and also support potentially calling CompleteAsync inside gRPC server in the future.

src/Grpc.AspNetCore.Web/Internal/GrpcWebProtocolHelpers.cs
Tratcher: Right.

src/Grpc.AspNetCore.Web/GrpcWebServiceExtensions.cs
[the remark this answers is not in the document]
Tratcher: > Do you know of other middleware/plugins that only use AddXXX to set configuration that are optional?

StaticFiles is a good example.

src/Grpc.AspNetCore.Web/Internal/GrpcWebMiddleware.cs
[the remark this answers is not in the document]
Tratcher: > Are there other reasons to undo setting the features? Or is it considered something you should do to be a good middleware citizen?

Both. In the error handling scenario the route will change so you'll end up at a different endpoint. If the base64 PipeWriter is still in place that could get ugly. Reverting also avoids side-effects for middleware earlier in the pipeline.

src/Grpc.AspNetCore.Web/Internal/GrpcWebFeature.cs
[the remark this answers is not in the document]
Tratcher: You really need to re-think how you do timeouts then. Multi-threaded request/response access must be prevented. Calling CompleteAsync from a timer thread is fundamentally unsafe.

src/Grpc.AspNetCore.Web/Internal/GrpcWebFeature.cs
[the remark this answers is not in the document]
Tratcher: Better to throw in those situations then.

src/Grpc.AspNetCore.Web/GrpcWebServiceExtensions.cs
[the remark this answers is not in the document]
davidfowl: It’s usually hanging off a framework specific builder

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
```diff
@@ -0,0 +1,193 @@
...
+                if (innerResult.IsCompleted)
+                {
+                    // If the reader completes with less than 4 bytes then the base64 isn't valid..
+                    throw new InvalidOperationException("Unexpected end of data when reading base64 content.");
+                }
+
+                if (innerResult.IsCanceled)
+                {
+                    return new ReadResult(
+                        ReadOnlySequence<byte>.Empty,
```
halter73: Are you sure you can advance after returning this canceled ReadResult? I think this would result in unresolvable SequencePositions being passed to the next call to AdvanceTo. It's possible that _currentDecodedBuffer and _currentInnerBuffer might not even be set which would lead to a null ref.

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
```diff
@@ -0,0 +1,193 @@
...
+            var consumedPosition = ResolvePosition(consumed);
+            var examinedPosition = ResolvePosition(examined);
+
+            _inner.AdvanceTo(consumedPosition, examinedPosition);
+        }
+
+        private SequencePosition ResolvePosition(SequencePosition base64Position)
+        {
+            // TODO - is there a better way to get the index of a position in a sequence?
+            var length = _currentDecodedBuffer.Slice(0, base64Position).Length;
```
halter73: I think this is the standard way to get the length between SequencePositions, but I agree it would be nice if there was a helper method to avoid the intermediate ReadOnlySequence. The good news is that neither Slice() nor the Length property loop over the individual segments\*. ReadOnlySequence.Length uses the segment's "RunningIndex" to make calculating the length O(1).

https://github.com/dotnet/runtime/blob/377b3c7c3e364b3183696744c6a5c6968bcdffd4/src/libraries/System.Memory/src/System/Buffers/ReadOnlySequence.Helpers.cs#L501-L518

\* Slice will loop over segments if given a longs/ints instead of SequencePositions. But in this case, since the only long/int being passed in is 0 as the start index, looping should not happen in practice.

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
```diff
@@ -0,0 +1,193 @@
...
+
+            // IMPORTANT - this logic assumes that SequencePosition will be multiples of 3.
+            // GetMaxEncodedToUtf8Length will round up to next multiple of 3, so more data
+            // could be advanced internally than expected if not called using a mulitple of 3.
+            //
+            // Because ReadAsync always returns data in multiples of 3, and gRPC only uses
+            // ReadResult.Buffer.End to advance, the position should always be a multiple
+            // of 3. The only time it won't be is at the end of a message, in which case
+            // rounding up is the correct thing to do to consume padding.
```
halter73: Can this throw an InvalidOperationException or something to fail early if `length % 3 != 0`?

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
```diff
@@ -0,0 +1,193 @@
...
+                _inner.AdvanceTo(innerResult.Buffer.Start, innerResult.Buffer.End);
+                innerResult = await _inner.ReadAsync(cancellationToken);
+            }
+
+            // Limit result to complete base64 segments (multiples of 4)
+            var buffer = innerResult.Buffer.Slice(0, (innerResult.Buffer.Length / 4) * 4);
+
+            // The content can contain multiple fragments of base64 content
+            // Check for padding, and limit returned data to one fragment at a time
+            var paddingIndex = PositionOf(buffer, (byte)'=');
```
halter73: Is there a way you could use [ReadOnlySequence.TryReadTo](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Memory/src/System/Buffers/SequenceReader.Search.cs#L189) instead? I see you're doing some logic to make sure the inner buffer's length is a multiple of 4, so it might be tricky. The nice thing about TryReadTo is it basically combines PositionOf and Slice into a single operation meaning it doesn't need to loop over the segments twice in the multi-segment case.

JamesNK: Need to split on the last padding character (in 4th position), not the first.

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
JamesNK: If length % 3 != 0 then it was because you're at the end of the message. You still want to advance 4 to consume padding.

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
```diff
@@ -0,0 +1,193 @@
...
+            SequencePosition position = source.Start;
+            int total = 0;
+            while (source.TryGet(ref position, out ReadOnlyMemory<T> memory))
+            {
+                int index = memory.Span.IndexOf(value);
+                if (index != -1)
+                {
+                    return total + index;
+                }
+                else if (position.GetObject() == null)
```
halter73: Nit: Is this really necessary? Wouldn't the next TryGet fail in this case?

JamesNK: It breaks out of the while loop so there won't be a next TryGet. This method duplicates what ReadOnlySequence does.

halter73: Are you copying this from ReadOnlySequence.Enumerator.TryMove()? If so, I'm pretty sure the `_next.GetObject() == null` check is only there to protect against you already being at the end of the sequence. So if you need this check at all, you'd want to check if `source.Start.GetObject() == null` at the start of the method, not in the middle of the while loop when TryGet just returned true.

src/Grpc.AspNetCore.Web/Internal/GrpcWebMiddleware.cs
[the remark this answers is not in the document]
JamesNK: Done

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
JamesNK: It's the same as this - https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/src/libraries/System.Memory/src/System/Buffers/BuffersExtensions.cs#L18-L56 - expect it returns an index relative to the start of the sequence instead of a position.

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
JamesNK: Done.

btw I experimented and PipeReader.Advance isn't required after a canceled result, but you can choose to do it. I fixed some issues so CancelPendingReader -> Advance this can succeed.

src/Grpc.Net.Client.Web/GrpcWebHandler.cs
```diff
@@ -0,0 +1,127 @@
...
+            {
+                return SendAsyncCore(request, cancellationToken);
+            }
+
+            return base.SendAsync(request, cancellationToken);
+        }
+
+        private async Task<HttpResponseMessage> SendAsyncCore(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            request.Content = new GrpcWebRequestContent(request.Content, _mode);
```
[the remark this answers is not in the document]
JamesNK: That's already done before `SendAsync` will call `SendAsyncCore`

src/Grpc.AspNetCore.Web/Internal/GrpcWebFeature.cs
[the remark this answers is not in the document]
JamesNK: A solution would be to add a service that is responsible for writing messages. gRPC-Web could then use DI to wrap it, and add locking to the wrapper to co-ordinate with CompleteAsync.

This will require changes to the core gRPC library so I'll leave it as is until we decide we want to commit to supporting this.

src/Grpc.AspNetCore.Web/Internal/GrpcWebMiddleware.cs
```diff
@@ -0,0 +1,165 @@
...
+            // Modifying the request is required to stop Grpc.AspNetCore.Server from rejecting it
+            httpContext.Request.Protocol = GrpcWebProtocolConstants.Http2Protocol;
+            httpContext.Request.ContentType = ResolveContentType(GrpcWebProtocolConstants.GrpcContentType, httpContext.Request.ContentType);
+
+            // Update response content type back to gRPC-Web
+            httpContext.Response.OnStarting(() =>
+            {
+                // Reset request protocol back to its original value. Not doing this causes a 2 second
+                // delay when making HTTP/1.1 calls.
+                httpContext.Request.Protocol = initialProtocol;
```
JamesNK: @Tratcher @halter73 I noticed a perf issue when using this middleware with HTTP/1.1. Resetting the protocol fixed it. What was the cause?

image

Tratcher: It looks like it's trying to do one request per connection? I wonder if Kestrel's response logic uses the modifiable protocol field, it shouldn't.

Secondary.. Is your sample listening on IPv6? It looks like it's only listening on IPv4.

halter73: Definitely looks like a bug: https://github.com/aspnet/AspNetCore/issues/17975

JamesNK: > Secondary.. Is your sample listening on IPv6? It looks like it's only listening on IPv4.

It is setting up Kestrel endpoint with: `Listen(new IPEndPoint(IPAddress.Loopback, 5000))`. Is the two second delay when a connection starts HttpClient attempting to start a v6 connection and failing?

Tratcher: > Is the two second delay when a connection starts HttpClient attempting to start a v6 connection and failing?

Yeah, you can see it in the timestamps. Always use ListenLocalhost(port) or ListenAny(port) to get both IPv4 and IPv6.

src/Grpc.AspNetCore.Web/Internal/GrpcWebProtocolConstants.cs
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+namespace Grpc.AspNetCore.Web.Internal
+{
+    internal static class GrpcWebProtocolConstants
+    {
+        internal const string GrpcContentType = "application/grpc";
```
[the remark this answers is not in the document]
JamesNK: I don't want to because the client, server and web packages are all referenced by functional tests. Sharing a file between all of them will create type name conflicts.

Anyway, it's just a string constant that will never change.

src/Grpc.AspNetCore.Web/Internal/GrpcWebMiddleware.cs
```diff
@@ -0,0 +1,165 @@
...
+            return _next(httpContext);
+        }
+
+        private async Task HandleGrpcWebRequest(HttpContext httpContext, GrpcWebMode mode)
+        {
+            var feature = CreateAndSetGrpcWebFeature(mode, httpContext);
+
+            var initialProtocol = httpContext.Request.Protocol;
+
+            // Modifying the request is required to stop Grpc.AspNetCore.Server from rejecting it
```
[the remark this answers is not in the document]
JamesNK: That would require Grpc.AspNetCore.Server to have a dependency (or knowledge) or Grpc.AspNetCore.Web.

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
```diff
@@ -0,0 +1,304 @@
...
+            //
+            // JavaScript gRPC-Web trailer parsing logic for comparison:
+            // https://github.com/grpc/grpc-web/blob/55ebde4719c7ad5e58aaa5205cdbd77a76ea9de3/javascript/net/grpc/web/grpcwebclientreadablestream.js#L292-L309
+
+            var remainingContent = span;
+            while (remainingContent.Length > 0)
+            {
+                ReadOnlySpan<byte> line;
+
+                var lineEndIndex = remainingContent.IndexOf((byte)'\r');
```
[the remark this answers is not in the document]
JamesNK: I didn't know there was a overload to do that. Nice.

brmdbr:
I've build it from the sources to use in my Blazor client side app and I get the following error: WASM: ﻿Status(StatusCode=Internal, Detail="Error starting gRPC call: Cannot invoke method because it was wiped. See stack trace for details.")

Is this a problem related to this pr or am I implementing something wrong?

JamesNK:
Instead of passing `HttpClientHandler` to `GrpcWebHandler`, create `WasmHttpMessageHandler`. There seems to be a bug in Blazor that its not public. You can do this:

```cs
var wasmHttpMessageHandlerType = Assembly.Load("WebAssembly.Net.Http").GetType("WebAssembly.Net.Http.HttpClient.WasmHttpMessageHandler");
var wasmHttpMessageHandler = (HttpMessageHandler)Activator.CreateInstance(wasmHttpMessageHandlerType);
```

https://github.com/aspnet/AspNetCore/issues/17115

brmdbr:
Thanks @JamesNK, that seems to work, I no longer get the cannot invoke error. I get another error now: WASM: ﻿Status(StatusCode=Internal, Detail="Error starting gRPC call: Buffer is not large enough for header")

Using this code:

```
var wasmHttpMessageHandlerType = Assembly.Load("WebAssembly.Net.Http").GetType("WebAssembly.Net.Http.HttpClient.WasmHttpMessageHandler");
var wasmHttpMessageHandler = (HttpMessageHandler)Activator.CreateInstance(wasmHttpMessageHandlerType);
var httpClient = new HttpClient(new GrpcWebHandler(GrpcWebMode.GrpcWebText, wasmHttpMessageHandler));
var channel = GrpcChannel.ForAddress("https://localhost:5001", new GrpcChannelOptions
            {
                HttpClient = httpClient
            });
var client = new Protos.LessonService.LessonServiceClient(channel);
var lesson = await client.GetLessonAsync(new Protos.GetLessonRequest { AllowedChars = "asdfghjkl;qwertyuiop", NumberOfWords = 15, Seed = 1337 });
```

JamesNK:
Ohh, I forgot that I modified the client slightly.

You'll need to also compile and use Grpc.Net.Client.

brmdbr:
> Ohh, I forgot that I modified the client slightly.
> 
> You'll need to also compile and use Grpc.Net.Client.

That fixed it! Thanks a lot.

SteveSandersonMS:
> Instead of passing HttpClientHandler to GrpcWebHandler, create WasmHttpMessageHandler. There seems to be a bug in Blazor that its not public

Thanks - will be fixed with https://github.com/aspnet/Blazor/pull/1960

src/Grpc.Net.Client.Web/Internal/Base64ResponseStream.cs
```diff
@@ -0,0 +1,271 @@
...
+            }
+            underlyingReadData = SetRemainder(underlyingReadData);
+
+            // We want to read base64 data in multiples of 4
+            underlyingReadData = underlyingReadData.Slice(0, (underlyingReadData.Length / 4) * 4);
+
+            var availableReadData = underlyingReadData;
+
+            var totalRead = 0;
+            // Minimum valid base64 length is 4. Read until we have at least that much content
```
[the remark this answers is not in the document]
JamesNK: Correct

src/Grpc.AspNetCore.Web/Internal/Base64PipeWriter.cs
```diff
@@ -0,0 +1,190 @@
...
+                PreserveRemainder(_buffer.AsSpan(bytesToProcess, newRemainder));
+
+                // When writing base64 content we don't want any padding until the end.
+                // Process data in intervals of 3, and save the remainder at the start of a new span
+                var buffer = _inner.GetSpan((bytesToProcess / 3) * 4);
+                CoreAdvance(_buffer.AsSpan(0, bytesToProcess), buffer);
+            }
+            else
+            {
+                PreserveRemainder(_buffer.AsSpan(0, resolvedBytes));
```
JunTaoLuo: Add a comment that in this case resolvedBytes and newRemainder are equal.

src/Grpc.AspNetCore.Web/Internal/Base64PipeWriter.cs
```diff
@@ -0,0 +1,190 @@
...
+
+            var resolvedBytes = bytes + _remainder;
+            var newRemainder = resolvedBytes % 3;
+            var bytesToProcess = resolvedBytes - newRemainder;
+
+            if (bytesToProcess > 0)
+            {
+                PreserveRemainder(_buffer.AsSpan(bytesToProcess, newRemainder));
+
+                // When writing base64 content we don't want any padding until the end.
```
JunTaoLuo: I agree with this approach though technically, in the case of gRPC-web, I believe we are allowed to add padding in the middle right?

JamesNK: FlushAsync is called at the end of each message. I'll mention that in a comment.

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
halter73: This Base64PipeReader will be exposed to developers as `HttpContext.Request.BodyReader`, right? This would mean that developers can pass in arbitrary `examined` SequencePositions to PipeReader.AdvanceTo. In practice it makes sense that examined will always be `ReadResult.Buffer.End`, but we cannot know that for sure.

I still think we should throw if `length % 3 != 0` and we aren't at the end of the message.

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
```diff
@@ -0,0 +1,222 @@
...
+            var length = _currentDecodedBuffer.Slice(0, base64Position).Length;
+
+            // IMPORTANT - this logic assumes that SequencePosition will be multiples of 3.
+            // GetMaxEncodedToUtf8Length will round up to next multiple of 3, so more data
+            // could be advanced internally than expected if not called using a mulitple of 3.
+            //
+            // Because ReadAsync always returns data in multiples of 3, and gRPC only uses
+            // ReadResult.Buffer.End to advance, the position should always be a multiple
+            // of 3. The only time it won't be is at the end of a message, in which case
+            // rounding up is the correct thing to do to consume padding.
```
halter73: I'm resubmitting this comment, since the last one was on a resolved conversation which I cannot unresolve.

This Base64PipeReader will be exposed to developers as `HttpContext.Request.BodyReader`, right? This would mean that developers can pass in arbitrary `examined` SequencePositions to PipeReader.AdvanceTo. In practice it makes sense that examined will always be `ReadResult.Buffer.End`, but we cannot know that for sure.

I still think we should throw if `length % 3 != 0` and we aren't at the end of the message.

src/Grpc.Net.Client.Web/Internal/Base64ResponseStream.cs
```diff
@@ -0,0 +1,276 @@
...
+            do
+            {
+                var read = await _inner.ReadAsync(availableReadData, cancellationToken);
```
halter73: [In Kestrels HttpRequestStream](https://github.com/aspnet/AspNetCore/blob/478353a40feb5f37ef9f1fcea96ba1cf34e1b760/src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestStream.cs#L122) and the [runtime's PipeReaderStream](https://github.com/dotnet/runtime/blob/master/src/libraries/System.IO.Pipelines/src/System/IO/Pipelines/PipeReaderStream.cs#L82-L85) (which is used by PipeReader.AsStream()) we throw OperationCanceledExceptions if the underlying pipe was canceled. We should probably do the same here.

JamesNK: This code is calling `int ReadAsync(Memory<byte>)` on Stream, not PipeReader. There is no cancel read result

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
JamesNK: The pipereader doesn't know it is at the end of the message, but when it is then there will be padding, e.g. `(byte)'='`. I've added some validation.

testassets/InteropTestsClient/RunGrpcWebTests.ps1
```diff
@@ -0,0 +1,28 @@
...
+Write-Host "Use TLS: $use_tls" -ForegroundColor Cyan
+Write-Host
+
+foreach ($test in $allTests)
+{
+  foreach ($grpcWebMode in "GrpcWeb","GrpcWebText")
+  {
+    Write-Host "Running $grpcWebMode $test" -ForegroundColor Cyan
+
+    dotnet run --use_tls $use_tls --server_port 8080 --client_type httpclient --grpc_web_mode $grpcWebMode --test_case $test
```
jtattermusch: I'm assuming that with the right infrastructure in place, these tests can run against the code in grpc-web, but we are not currently testing that?

testassets/InteropTestsClient/RunGrpcWebTests.ps1
```diff
@@ -0,0 +1,28 @@
+﻿Param
+(
+    [bool]$use_tls = $false
+)
+
+$allTests =
+  "grpcweb_unary",
```
jtattermusch: are these "standard" scenarios that are documented somewhere or we just made those up?
For regular interop tests, we have https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md, I don't recall seeing a similar doc for grpc-web.
We should reach out to grpc-web team to see what's the best way to test against other implementations (="interop tests")

testassets/FunctionalTestsWebsite/Services/EchoService.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Grpc.Core;
+using Grpc.Gateway.Testing;
+
+namespace FunctionalTestsWebsite.Services
+{
+    /// <summary>
+    /// Written to as closely as possible mirror the behaviour of the C++ implementation in grpc/grpc-web:
+    /// https://github.com/grpc/grpc-web/blob/92aa9f8fc8e7af4aadede52ea075dd5790a63b62/net/grpc/gateway/examples/echo/echo_service_impl.cc
```
jtattermusch: Seems like a good start for a set of "interop" tests, but we should reach out to grpc-web team to confirm.

kokoro/build_nuget.sh
```diff
@@ -31,9 +31,11 @@ build/expand_dev_version.sh
 
 (cd src/Grpc.Net.Common && dotnet pack --configuration Release --output ../../artifacts)
 (cd src/Grpc.Net.Client && dotnet pack --configuration Release --output ../../artifacts)
+(cd src/Grpc.Net.Client.Web && dotnet pack --configuration Release --output ../../artifacts)
```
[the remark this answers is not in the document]
JamesNK: I think we should only push these two packages as previews.

Is your publish process a manual one or do you have a script? If you have a script then you would need a way to skip these two packages. Alternatively, we could set them to be non-visible.

src/Grpc.AspNetCore.Web/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+This is an experimental project. If you have issues or suggestions for features, please give feedback at https://github.com/grpc/grpc-dotnet/issues
+
+---
+
+[gRPC-Web](https://grpc.io/blog/state-of-grpc-web/) allows gRPC to be used over HTTP/1.1 connections. gRPC-Web enables using gRPC in new scenarios:
+
+- JavaScript browser applications can call gRPC services using the [gRPC-Web JavaScript client](https://github.com/grpc/grpc-web).
+- Blazor WebAssembly applications can call gRPC services using the .NET Core gRPC client.
+- gRPC services can be used in environments that don't have complete support for HTTP/2.
```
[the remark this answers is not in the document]
JamesNK: There are some proxies that that don't have complete support for HTTP/2, and some servers. IIS is an example. gRPC is not usable on IIS, but gRPC-Web is.

src/Grpc.AspNetCore.Web/README.md
```diff
@@ -0,0 +1,64 @@
...
+});
+```
+
+## Grpc.Net.Client.Web
+
+Grpc.Net.Client.Web provides a HttpClient delegating handler that configures the .NET Core gRPC client to send gRPC-Web calls.
+
+```csharp
+var httpClientHandler = new HttpClientHandler();
+var httpClient = new HttpClient(new GrpcWebHandler(GrpcWebMode.GrpcWebText, httpClientHandler));
```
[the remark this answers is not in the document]
JamesNK: I agree. I think a future task will be to figure out how to compose GrpcChannel and HttpClient together better.

src/Grpc.AspNetCore.Web/Internal/GrpcWebMode.cs
```diff
@@ -0,0 +1,27 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+namespace Grpc.AspNetCore.Web.Internal
+{
+    internal enum GrpcWebMode
```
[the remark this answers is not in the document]
JamesNK: Renamed to `ServerGrpcWebMode`

testassets/InteropTestsClient/RunGrpcWebTests.ps1
JamesNK: These are from https://github.com/johanbrandhorst/grpc-web-compatibility-test. I've added a comment.

testassets/FunctionalTestsWebsite/Services/EchoService.cs
JamesNK: They're taken from https://github.com/johanbrandhorst/grpc-web-compatibility-test

As far as I know there are no other interop tests. We should check that with grpc-web team.

testassets/InteropTestsClient/RunGrpcWebTests.ps1
JamesNK: No. https://github.com/johanbrandhorst/grpc-web-compatibility-test has Docker containers that allow you to manually run implementations against one another, but I couldn't see anything that is automated

kokoro/build_nuget.sh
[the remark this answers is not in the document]
jtattermusch: In the release process, I take all the nugets built by build_nuget.sh, sign them using a separate process and then push them all.

testassets/InteropTestsClient/InteropClient.cs
```diff
@@ -298,11 +316,82 @@ private async Task RunTestCaseAsync(IChannelWrapper channel, ClientOptions optio
                 case "server_compressed_streaming":
                     await RunServerCompressedStreamingAsync(client);
                     break;
+                case "grpcweb_unary":
```
jtattermusch: For InteropTestsClient and InteropTestsWebsite, I'd like to only add "official" interop tests that have a spec.
I know this is a convenient place to put the grpc-web test, but we probably don't want to mix official and unofficial interop tests.
If you think moving the tests elsewhere would be too difficult, can we at least add a note where these test are coming from and that they are temporary and commit to replacing the grpc-web tests with the official ones once the spec exists?

kokoro/build_nuget.sh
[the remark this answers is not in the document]
jtattermusch: I pushed a patch for the script that should solve our problem
https://github.com/grpc/grpc-dotnet/pull/695/commits/c9338c7cca5cc0d147c3064924214ff459cf94f0

testassets/InteropTestsWebsite/EchoService.cs
```diff
@@ -0,0 +1,112 @@
...
+using Grpc.Core;
+using Grpc.Gateway.Testing;
+
+namespace Grpc.Testing
+{
+    /// <summary>
+    /// Written to as closely as possible mirror the behaviour of the C++ implementation in grpc/grpc-web:
+    /// https://github.com/grpc/grpc-web/blob/92aa9f8fc8e7af4aadede52ea075dd5790a63b62/net/grpc/gateway/examples/echo/echo_service_impl.cc
+    /// </summary>
+    public class EchoService : Grpc.Gateway.Testing.EchoService.EchoServiceBase
```
jtattermusch: looks like you have a copy of the same service at `testassets/FunctionalTestsWebsite/Services/EchoService.cs`?

test/FunctionalTests/Infrastructure/GrpcTestFixture.cs
```diff
@@ -61,9 +62,24 @@ public GrpcTestFixture(Action<IServiceCollection>? initialConfigureServices = nu
 
         public HttpClient CreateClient(TestServerEndpointName? endpointName = null, DelegatingHandler? messageHandler = null)
         {
+            endpointName ??= TestServerEndpointName.Http2;
+
             var httpClientHandler = new HttpClientHandler();
             httpClientHandler.ServerCertificateCustomValidationCallback = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator;
 
+#if GrpcWebTests
```
jtattermusch: is this used from anywhere?

testassets/InteropTestsWebsite/EchoService.cs
JamesNK: Yes. That's so it is tested by CI

testassets/InteropTestsClient/InteropClient.cs
JamesNK: I'll add a note. If/when grpc-web interop tests become more established then we could move them to their own server/client.

JamesNK:
🎈 🎉 🍰

johanbrandhorst:
Hi, great work @JamesNK. Would you be interested in adding this implementation to the grpc-web compatibility tests I'm running? See https://github.com/johanbrandhorst/grpc-web-compatibility-test

muraray:
That's one hell of a serious pull request... grpc port to .net core paradigm, very much appreciate your work, thanks a ton... looking forward using it in our next project...

PR #706, opened 2019-12-27
Improve error message when SslCredentials has arguments

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/702


JunTaoLuo:
In addition to the error message, maybe we should clarify in the doc comment as well: https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.Net.Client/GrpcChannelOptions.cs#L33-L36

Issue #520, opened 2019-09-17
Method level Interceptors


cosullivan:
Hi,

Are there any plans to add support for method level interceptors?

I have a project which I have created a custom IServiceMethodProvider<> implementation for service discovery and would like the ability here to add interceptors for specific methods in the service. These interceptors would then be combined with any global or service level interceptors.

I have tested this functionality on a fork by adding it just for Unary methods and it is fairly straight forward to pass this through. 

Thanks,
Cain.

JamesNK:
None at the moment. This is the first time it has been requested. Right now configuration, including interceptors can be customized per-service.

cosullivan:
Hi,

I am happy to contribute a PR if you are OK with the idea of method level options/interceptors.

I have implemented it for Unary calls only in my fork as that's what I was interested in and the changes look like this;

Manual configuration:
```
services.AddGrpc().AddServiceOptions<GreeterServer>(serviceOptions => 
{
    serviceOptions.AddMethodOptions(nameof(GreeterService.SayHello), methodOptions =>                    
    { 
        methodOptions.Interceptors.Add<TracingInterceptor>();
    });    
})
```
To support configuration via IServiceMethodProvider<> I have changed the ServiceMethodProviderContext<TService> so the options can be modified during discovery time.

```
public void AddUnaryMethod<TRequest, TResponse>(Method<TRequest, TResponse> method, IList<object> metadata, UnaryServerMethod<TService, TRequest, TResponse> invoker, Action<GrpcServiceMethodOptions> configure)
```

Let me know your thoughts.

Thanks,
Cain.

JamesNK:
Interesting feature! I want to be careful about adding this because it has a lot of public API.

Some things to consider:
1. What about setting the other properties on a service against a method
2. Would configuring interceptors and other options using attributes be better?

cosullivan:
I made the change on a fork to minimize breakage to the public API, but there might be some better ways to do it.

In regards to your considerations,

1. Yes, I think a lot of settings could be added on a method level and it would make sense to do so. I would imaging that method level takes preference over service level over global.

2. Configuring via attributes is actually what I was wanting this change for. I have a custom IServiceMethodProvider which configures a service based entirely on attributes, similar to how a Controller can be configured with routes in MVC. 

I have it working with Unary methods only at the moment but this is what it looks like;

```
[GrpcService("Health")]
public class HealthService
{
    [GrpcMethod]
[...]

PR #712, opened 2020-01-06
Add IGrpcServiceOptionsMetadata

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/520

This allows options to be set via endpoint metadata. I haven't added any attributes for setting values, but it is possible for developers to easily write an attribute to set whatever value on service options they want.

`InterceptorAttribute` in this PR's tests is an example of attributes that can be created with this.

### DIFF
--- a/testassets/FunctionalTestsWebsite/Infrastructure/InterceptorAttribute.cs
+++ b/testassets/FunctionalTestsWebsite/Infrastructure/InterceptorAttribute.cs
@@ -1,0 +1,40 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


testassets/FunctionalTestsWebsite/Infrastructure/InterceptorAttribute.cs
```diff
@@ -0,0 +1,40 @@
...
+
+#endregion
+
+using Grpc.AspNetCore.Server;
+using Grpc.Core;
+using System;
+
+namespace FunctionalTestsWebsite.Infrastructure
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
```
JunTaoLuo: I can see class level attributes becoming a problem. It's essentially a service level option but has higher priority.

JamesNK: Yes, that's a downside to multiple configuration options.

I don't have a problem with service level attribute configuration overriding service level code configuration. Attributes taking precedence in programming models is fairly common in .NET

JunTaoLuo: This is something we will need documentation for. I feel it could be a common pit of failure.

anurse:
Ping: @JamesNK you said you were going to check in with @rynowak about this one.

JamesNK:
Won't do this at the moment. If there is more demand then might reopen and put some more thought into it.

PR #716, opened 2020-01-13
Separating out ResponseMessageContext logic from HttpContextServerCallContext

vaughanr:
Hi, 

I initially started trying to make the HttpContextServerCallContext class use generics instead passing in types into the constructor but realized it was too large a change because it is passed into quite a few other classes such as the HttpContextSerializationContext and various methods in PipeExtensions. 

So this change is just trying to decouple the HttpContextServerCallContext from the HttpContextSerializationContext by instead passing through a new class ResponseMessageContext which has the needed information.

I understand that this is a little larger pull request which might require someone spending some time going through it. My intention is to try get a little more involved in the project but I definitely don't want to take you guys from the great work you have been doing. 

If you would prefer to not make this change I am happy to delete the pr and find something else to contribute.

Thanks!

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/GrpcProtocolHelpers.cs
@@ -212,17 +212,5 @@ namespace Grpc.AspNetCore.Server.Internal
                 values.Add(AuthProperty.Create(name, Encoding.UTF8.GetBytes(value)));
             }
         }
-
-        internal static bool CanWriteCompressed(WriteOptions? writeOptions)
-        {
-            if (writeOptions == null)
-            {
-                return true;
-            }
-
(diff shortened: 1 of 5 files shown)


JamesNK:
What is the reason for this change? The context is allocated for every request. Splitting it up will add another per request allocation.

vaughanr:
Hi James,

thanks for having a look. 

The idea was to try take what was needed from the HttpContextServerCallContext when passing it into the HttpContextSerializationContext instead of the whole instance. The larger goal was to try cut down what is dependent on the implementation of that HttpContextServerCallContext to hopefully decouple it.

To be honest i was originally looking at making the HttpContextServerCallContext class use generic types instead of passing the request and response types into the constructor but it became too large a change because of what is dependent on it. 

I understand that you guys have longer term goals for what you want from these classes so its not an issue if you don't want to go that direction.

vaughanr:
I can see that there is no appetite for this change. Thats not an issue at it. It might have been a little ambitious for an early change. I will close and find something better to contribute.

PR #720, opened 2020-01-14
Use new Grpc.Core.Api to get response headers only when requested

JamesNK:
Avoid allocating delegates per call, and only loads Metadata with response headers when response headers are requested by the user.

Will benchmark soon.


JamesNK:
Before:
```
        Method |     Mean |     Error |    StdDev |     Op/s | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
-------------- |---------:|----------:|----------:|---------:|------------:|------------:|------------:|--------------------:|
 SayHelloAsync | 10.24 us | 0.1889 us | 0.1767 us | 97,619.8 |      0.0916 |           - |           - |             7.11 KB |
```

After:
```
-------------- |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 SayHelloAsync | 9.354 us | 0.1853 us | 0.2474 us | 106,903.8 |      0.0763 |           - |           - |             6.66 KB |
```

PR #766, opened 2020-02-12
Fix sending large grpc-web-text requests

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/752


test/FunctionalTests/Web/Client/IssueTests.cs
```diff
@@ -0,0 +1,862 @@
...
+            var response = await client.GetLibraryAsync(request);
+
+            // Assert
+            Assert.AreEqual("admin", response.UserId);
+        }
+
+        private const string JsonRequest = @"{
+  ""userId"": ""admin"",
+  ""carriers"": [
+    ""273"",
```
JunTaoLuo: Can we generate this content instead of leaving it in this giant file?

PR #788, opened 2020-02-24
Add option to ignore calls to unknown gRPC services

JamesNK:
The way routing needs to be setup to handle returning unimplemented for non-existent gRPC methods and services adds inefficiencies and allocations:

1. Catch-all background routes mean multiple endpoints match a URL so we can't use the routing fast path
2. Route parameters require string allocations of each parameter
3. Running a constraint on the global route will allocate a RouteValuesDictionary

~This PR removes "unimplemented" endpoints. Calling a URL that would have gone to these endpoints will now return HTTP status code 404. Clients will map 404 to Unimplemented.~

~Also remove unimplemented counter. Observing data around 404 requests can be used instead.~

Added an option so a dev can choose whether to process unmapped gRPC calls and return unimplemented instead of 404.

### DIFF
--- a/perf/benchmarkapps/GrpcAspNetCoreServer/Startup.cs
+++ b/perf/benchmarkapps/GrpcAspNetCoreServer/Startup.cs
@@ -37,7 +37,7 @@ namespace GrpcAspNetCoreServer
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddGrpc();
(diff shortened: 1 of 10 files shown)


JamesNK:
We're discussing this over email. Will wait for his feedback before merging.

JamesNK:
Talked with @jtattermusch and he is ok with this setting. Feedback to make the current behavior the default applied to PR. Apps can now choose to opt out of returning UNIMPLEMENTED.

Issue #798, opened 2020-03-02
Improve error when client attempts to call a server with an invalid TLS certificate


JamesNK:
Example: https://github.com/grpc/grpc-dotnet/issues/792

At the moment the developer gets a `RpcException` with the message "Error starting gRPC call: The SSL connection could not be established, see inner exception.".

However there is no inner exception because RpcException doesn't allow it. They have no way of knowing what the error is without setting up client logging. If they could see the inner exception then they would know what the issue is:

> System.Security.Authentication.AuthenticationException: The remote certificate is invalid according to the validation procedure.

Three options to improve this experience:
1. Include the inner exception(s) message in the RpcException details:

    > Error starting gRPC call: The SSL connection could not be established, see inner exception.
    > The remote certificate is invalid according to the validation procedure.

2. Update RpcException with a constructor that takes an inner exception. This would allow for us to provide more information with RpcException when created from exceptions thrown on the client.
However it could be confusing on the server. When a user throws an RpcException with an inner exception only the status and details are propagated to the client.
3. When an error occurs starting a gRPC call, throw the original exception rather than wrapping it in RpcException.

@jtattermusch what do you think?

PR #804, opened 2020-03-04
Include inner exception messages in RpcException status detail

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/798

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/ErrorMessageHelper.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/ErrorMessageHelper.cs
@@ -17,7 +17,6 @@
 #endregion
 
 using System;
-using System.Text;
 using Grpc.Shared;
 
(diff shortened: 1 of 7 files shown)


src/Grpc.AspNetCore.Server/Internal/ErrorMessageHelper.cs
```diff
@@ -26,10 +27,39 @@ internal static string BuildErrorMessage(string message, Exception exception, bo
...
+                    ? $"{exception.GetType().Name}: {exception.Message}"
+                    : BuildErrorMessage(exception);
+
+                return message + " " + exceptionDetails;
             }
 
             return message;
         }
+
+        private static string BuildErrorMessage(Exception ex)
```
JunTaoLuo: This should be a shared file between server and client.

PR #810, opened 2020-03-06
Fix GrpcWebHandler not always resetting HTTP version back to 2.0

JamesNK:
https://github.com/grpc/grpc-dotnet/pull/809 but on v2.28.x branch.


jtattermusch:
Are you going to upmerge to master branch as well?

Issue #803, opened 2020-03-04
Propagation supporting HttpContext or no context


shaneqld:
### Is your feature request related to a problem? Please describe.
EnableCallContextPropagation() works great when you have a ServerCallContext. However, when you don't have a ServerCallContext, an exception is thrown when you attempt to use a client.

```cs
if (serverCallContext == null)
{
    throw new InvalidOperationException("Unable to propagate server context values to the call. Can't find the current gRPC ServerCallContext.");
}
```

There are two scenarios where this occurs that I've encountered.

1. Handling a HTTP1 request with Mvc. Third-party services generally don't send you gRPC requests (e.g. Pub/Sub, Cloud Tasks, any service sending you webhooks). In this scenario we have a standard HttpContext that may have headers that we can propagate (potentially they may be injected by istio).
2. Initialization logic, background polling, etc.

In both of these examples. I am also hosting a gRPC service, where EnableCallContextPropagation() works properly.  But as soon as you use a client in any of the above scenarios, you must disable EnableCallContextPropagation() on those clients, losing the tracing information you would other have had.

### Describe the solution you'd like
Some potential options:

- Don't throw an exception if a ServerCallContext can't be found, or provide a mechanism to disable the exception. This will at least allow propagation where it's possible.
- Support propagating headers from HttpContext to support MVC/others (best effort basis).

### Describe alternatives you've considered
I have disabled propagation for any clients that I need to use outside of gRPC handlers.

anurse:
Triage Notes: Seems reasonable. We shouldn't change the default behavior but can add a flag/option to `EnableCallContextPropagation` to suppress the exception in this situation.

PR #814, opened 2020-03-11
Client factory integration option to ignore context not found

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/803

### DIFF
--- a/src/Grpc.AspNetCore.Server.ClientFactory/GrpcContextPropagationOptions.cs
+++ b/src/Grpc.AspNetCore.Server.ClientFactory/GrpcContextPropagationOptions.cs
@@ -24,11 +24,15 @@ namespace Grpc.AspNetCore.ClientFactory
     public class GrpcContextPropagationOptions
     {
         /// <summary>
-        /// Gets or sets a value indicating whether to ignore context not found errors when the client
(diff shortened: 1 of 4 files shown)


src/Grpc.AspNetCore.Server.ClientFactory/GrpcContextPropagationOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// </summary>
+    public class GrpcContextPropagationOptions
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether to ignore context not found errors when the client
+        /// attempts to propagate the call context. This can happen when the client is used outside
+        /// the context of a server gRPC call.
+        /// Default value is <c>false</c>.
+        /// </summary>
+        public bool IgnoreContextNotFoundError { get; set; }
```
anurse: Triage consensus: We're ok with this, or removing the `Error` suffix

anurse:
```suggestion
        public bool ThrowIfContextNotFound { get; set; } = true;
```

Another option. Nobody is to Z E S T Y about this though.

PR #830, opened 2020-03-19
adding readme for App Service scenarios

btardif:
Calling out changes to App Settings needed to make this work with App Service

### DIFF
--- a/examples/Browser/readme.md
+++ b/examples/Browser/readme.md
@@ -1,0 +1,5 @@
+# Using this sample with App Service Linux
+
+If you plan to deploy this sample to Azure App Service on a Linux App Service plan, the `Protocols` variable defined in `appsettings.json` should be set to a value of `Http1`.
+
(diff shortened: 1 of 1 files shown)


thelinuxfoundation:

* User @btardif isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

JamesNK:
The protocol isn't required for this sample at all.

I think a better change would be to remove it from the json file and leave Kestrel to default to Http1AndHttp2. That should be fine in app service.

btardif:
> The protocol isn't required for this sample at all.
> 
> I think a better change would be to remove it from the json file and leave Kestrel to default to Http1AndHttp2. That should be fine in app service.

ok I'll try that out and confirm.

I'll update the PR if this works

AnthonyGiretti:
Thank you very much. Really helpful, it fixes my issue!

PR #772, opened 2020-02-15
Refine browser example and add CORS sample code

hanabi1224:
Fix https://github.com/grpc/grpc-dotnet/issues/771 
Besides CORS sample, also try to use typescript and MVVM framework instead of vanilla javascript.  Please let me know your concerns : )

PR #835, opened 2020-03-26
Add CORS example

JamesNK:
Replaces https://github.com/grpc/grpc-dotnet/pull/772


examples/Spar/Server/ClientApp/.gitignore
```diff
@@ -0,0 +1,40 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
```
JunTaoLuo: Why is there a .gitignore in here? Can it be moved to the root one instead?

examples/Spar/Server/ClientApp/package-lock.json
```diff
@@ -0,0 +1,7564 @@
+{
```
JunTaoLuo: Does this need to be checked in?

JamesNK: Yes

examples/Spar/Server/ClientApp/.gitignore
JamesNK: A nested gitignore is a standard part of the template. I think it is fine.

PR #837, opened 2020-03-27
proposed changes for generalizing client factory to other implementations

mgravell:
(for example: protobuf-net.Grpc)

This PR is largely a discussion document intended to discuss (with code) some problems in the current `ClientFactory` implementation that make it hard to generalize.

Problem statements:

1. there is an unnecessary `where T : ClientBase` constraint; this precludes usage with implementations that don't start from that assumption; this could be non-`ClientBase` classes, or naked interfaces
2. the client factory is a unary registration; this makes it impossible to *add* a custom factory without breaking code that relies on the *old* factory; and the old default factory is not directly available
3. a lot of complex nuanced code around http-client and GRPC configuration is hidden in `internal` implementations without any mechanism to control them; any custom client would have to essentially copy *most* of the `ClientFactory` code base; this is undesirable and will invariably lead to code drift

Proposed solutions:

1. replace `where T : ClientBase` with `where T : class`; this is presumably a helper guide for callers, but notably, the implementation does not *use* the `ClientBase` knowledge in any useful constraint, and it doesn't actually enforce the requirements needed by the activator/constructor API in use (i.e. even with the constraint a client can be constructed that does not work)
2. use an enumerable service descriptor rather than a unary service descriptor; each service may *try* and create a client, returning `null` if it cannot do so; the first to do so: wins; make the default descriptor more aware of what it can / cannot create
3. make all of the default configuration bits available behind a reusable API that just returns a fully configured `CallInvoker`

Example usage of this is shown in `CustomClientFactoryTests`, but the key impact here can be summarized in a few lines of code:

``` c#
// service registration
services.AddGrpcClient<IMyCustomService>(o =>
{
    o.Address = new Uri("https://nowhere.com/");
});
// factory registration (this would usually be hidden in a library-specific helper method)
services.TryAddEnumerable(
    ServiceDescriptor.Singleton<GrpcClientFactory, CustomGrpcClientFactory>()
);
```

with the custom factory (very simple here, for test purposes; real implementations would be library specific - it is not expected that end consumers would ever need to touch this):

``` c#
public class CustomGrpcClientFactory : GrpcClientFactory
{
    private readonly IServiceProvider _serviceProvider;
    public CustomGrpcClientFactory(IServiceProvider serviceProvider)
        => _serviceProvider = serviceProvider;
    public override TClient? CreateClient<TClient>(string name) where TClient : class
    {
        // only knows how to create one thing
        if (typeof(TClient) == typeof(IMyCustomService))
        {
            var callInvoker = GetCallInvoker<TClient>(_serviceProvider, name);
            return (TClient)(object)new MyService(callInvoker);
        }
        return null;
    }
}
```

### DIFF
--- a/src/Grpc.Net.ClientFactory/GrpcClientFactory.cs
+++ b/src/Grpc.Net.ClientFactory/GrpcClientFactory.cs
@@ -17,6 +17,10 @@
 #endregion
 
 using Grpc.Core;
+using System;
+using System.Net.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Grpc.Net.ClientFactory.Internal;
 
 namespace Grpc.Net.ClientFactory
 {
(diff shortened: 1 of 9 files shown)


JamesNK:
> replace where T : ClientBase with where T : class

Sure

> use an enumerable service descriptor rather than a unary service descriptor; each service may try and create a client, returning null if it cannot do so; the first to do so: wins; make the default descriptor more aware of what it can / cannot create

I'm not a fan of this. I think a better solution would be being able to register different factories for different types.

> make all of the default configuration bits available behind a reusable API that just returns a fully configured CallInvoker

Maybe. I'll look at your PR and see whether there is a way to do it in a way I like.

JamesNK:
Replaced by https://github.com/grpc/grpc-dotnet/pull/851

PR #851, opened 2020-04-03
Support per-client factories and non-ClientBase implementations

JamesNK:
Builds on top of https://github.com/grpc/grpc-dotnet/pull/837

Changes:
* Added `GrpcClientFactory<TClient>`. Libraries can wrap `AddGrpcClient<TClient>` and register a factory with DI.
* Split creating a call invoker from creating a client

// @mgravell


src/Grpc.Net.ClientFactory/GrpcClientFactory.Generic.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    /// <typeparam name="TClient">The gRPC client type.</typeparam>
+    public abstract class GrpcClientFactory<TClient> : GrpcClientFactory where TClient : class
+    {
+        /// <summary>
+        /// Create a gRPC client instance for the specified <typeparamref name="TClient"/> and configuration name.
+        /// </summary>
+        /// <param name="name">The configuration name.</param>
+        /// <returns>A gRPC client instance.</returns>
+        public abstract TClient CreateClient(string name);
```
JamesNK, line 35: protected?

src/Grpc.Net.ClientFactory/GrpcClientFactory.Generic.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <summary>
+        /// Create a gRPC client instance for the specified <typeparamref name="TFallbackClient"/> and configuration name.
+        /// </summary>
+        /// <typeparam name="TFallbackClient">The gRPC client type.</typeparam>
+        /// <param name="name">The configuration name.</param>
+        /// <returns>A gRPC client instance.</returns>
+        public override sealed TFallbackClient CreateClient<TFallbackClient>(string name)
+        {
+            var client = CreateClient(name) as TFallbackClient;
+            if (client == null)
```
JamesNK, line 46: Check that TClient exactly matches TFallbackClient?

src/Grpc.Net.ClientFactory/GrpcClientFactory.cs
```diff
@@ -32,6 +36,26 @@ public abstract class GrpcClientFactory
...
+        public abstract TClient CreateClient<TClient>(string name) where TClient : class;
+
+        /// <summary>
+        /// Gets a default CallInvoker instance for the specified <typeparamref name="TClient"/> and configuration name.
+        /// </summary>
+        /// <typeparam name="TClient">The gRPC client type.</typeparam>
+        /// <param name="serviceProvider">The service provider.</param>
+        /// <param name="name">The configuration name.</param>
+        /// <returns>A gRPC client instance.</returns>
+        protected static CallInvoker GetCallInvoker<TClient>(IServiceProvider serviceProvider, string name) where TClient : class
```
JamesNK, line 48: An alternative to this would be making GrpcCallInvokerFactory public.

src/Grpc.Net.ClientFactory/GrpcClientFactory.cs
```diff
@@ -32,6 +36,25 @@ public abstract class GrpcClientFactory
...
-        public abstract TClient CreateClient<TClient>(string name) where TClient : ClientBase;
+        public abstract TClient CreateClient<TClient>(string name) where TClient : class;
+
+        /// <summary>
+        /// Create a <see cref="CallInvoker"/> instance for the specified configuration name.
+        /// </summary>
+        /// <param name="serviceProvider">The service provider.</param>
+        /// <param name="name">The configuration name.</param>
+        /// <returns>A gRPC call invoker instance.</returns>
+        protected static CallInvoker CreateCallInvoker(IServiceProvider serviceProvider, string name)
```
mgravell, line 47: just to note: I was on the fence here as to whether this might prefer to be `protected virtual`, to allow `override`; not hugely invested either way, but probably merits a second look

test/Grpc.Net.ClientFactory.Tests/CustomClientFactoryTests.cs
```diff
@@ -0,0 +1,139 @@
...
+            services.AddGrpcClient<IMyCustomService>(o =>
+            {
+                o.Address = new Uri("https://nowhere.com/");
+            });
+            services.AddHttpClient("TestClient");
+
+            // this would be via a custom extension method or similar
+            if (registerCustomProvider)
+            {
+                // doing this twice to check that we don't end up with multiple factory instances
```
mgravell, line 94: (nit: comment no longer applies)

mgravell:
before I go too far in looking over the code, I think there's something very fundamental that needs consideration, as it may necessitate rework; in particular: this step in the tests is, IMO, invalid:

``` c#
services.TryAddSingleton<GrpcClientFactory<IMyCustomService>, CustomGrpcClientFactory<IMyCustomService>>();
```

In my mind, I was thinking of registration as something akin to

``` c#
services.AddCodeFirstClientFactory(); // library-specific extension method
// ...
services.AddGrpcClient<DefaultLookingClient>(...);
services.AddGrpcClient<IMyCustomService>(...);
```

(which pairs with `services.AddCodeFirstGrpc();` at the server); the point being, that the `services.AddCodeFirstClientFactory();` is the only one that differs from any other gRPC client factory usage; however, this won't work with the design as suggested here, because the `<T>` is not known at that time. The only way to square that circle would be to rework to something like:

``` c#
services.AddCodeFirstGrpcClient<DefaultLookingClient>(...);
services.AddCodeFirstGrpcClient<IMyCustomService>(...);
```

with libraries adding all of the overloads of:

``` c#
public static Whatever AddCodeFirstGrpcClient<TService>(this IServiceCollection services, {overload parameters}) {
services.TryAddSingleton<GrpcClientFactory<T>, CodeFirstGrpcClientFactory<T>>();
return services.AddGrpcClient<T>({overload parameters});
}
```

This is *possible*, but is a lot more convoluted, and is one of the things I was trying to avoid by allowing providers to handle entire *categories* of types.

JamesNK:
```cs
services.AddCodeFirstClientFactory(); // library-specific extension method
```
I get what you want but IMO this doesn't compose well. What happens if someone else registers a client factory in an app and it also thinks it can create `IMyCustomService`? You're stuck. The order that services are registered in becomes very important.

```cs
(which pairs with `services.AddCodeFirstGrpc();` at the server); 
```
A difference here is it is actually legitimate to have multiple providers for a gRPC service. For example, the standard gRPC and gRPC HTTP API providers register different ASP.NET Core endpoints from the same service. Meanwhile in the client factory only one person can actually create the client.

```cs
AddCodeFirstGrpcClient<IMyCustomService>(params)
```
That's how I imagined this being used. It explicitly allows someone to say that this client will be created by this factory. A few 2 line extension methods isn't that complicated.

An alternative could be to make the factory override a parameter on the client options:
```cs
services.AddGrpcClient<IMyCustomService>(o =>
{
    o.Address = new Uri("https://nowhere.com/");
    o.Factory = new CodeFirstGrpcFactory();
});
```
But I suspect you would probably still want to have a `AddCodeFirstGrpcClient` helper method that sets that value for people.

mgravell:
I see your point re a 3rd player. And I'm not set on any particular implementation - the only important thing is that we all understand the problems and requirements. It sounds like you've got a good handle on that. I was mulling around with some similar ideas on the client options approach - it may be fine; perhaps an extension method there would also work for the library - something like:

``` c#
services.AddGrpcClient<IMyCustomService>(o =>
{
    o.Address = new Uri("https://nowhere.com/");
    o.SomeLibraryExtensionMethod(); // so callers don't need to think
});
```

But: this is moving in a good direction! Should I kill the other PR so we can focus on this one?

JamesNK:
@mgravell I've updated the approach to use configuration on the options.

I think the best way to use this is to create an extension method that wraps calling CreateGrpcClient and ConfigureClientCreator. Example of Refit doing this with the base HttpClientFactory.

https://github.com/reactiveui/refit/blob/master/Refit.HttpClientFactory/HttpClientFactoryExtensions.cs

mgravell:
I like the new direction a lot; very easy to plug into, and still supports multiple providers. :thumbsup:

test/Grpc.Net.ClientFactory.Tests/GrpcHttpClientBuilderExtensionsTests.cs
```diff
@@ -222,6 +209,133 @@ public void AddInterceptorGeneric_NotFromGrpcClientFactory_ThrowError()
...
+
+            // Act
+            var clientFactory = serviceProvider.GetRequiredService<GrpcClientFactory>();
+            var ex = Assert.Throws<InvalidOperationException>(() => clientFactory.CreateClient<Greeter.GreeterClient>(nameof(Greeter.GreeterClient)));
+
+            // Assert
+            Assert.AreEqual("The System.Object instance returned by the configured client creator is not compatible with Greet.Greeter+GreeterClient.", ex.Message);
+        }
+
+        private class DerivedGreaterClient : Greeter.GreeterClient
```
jtattermusch, line 332: typo: /Greater/Greeter/

JamesNK:
@mgravell Can you please test that this works with your implementation. If fixes or changes need to be made, they'll need to happen before the next release.

It is available on the nightly NuGet feed:
https://github.com/grpc/grpc-dotnet#grpc-nuget-feed

mgravell:
Awesome; I'll reset my https://github.com/protobuf-net/protobuf-net.Grpc/pull/80 branch and refactor it based on the nightly

soroshsabz:
related to https://github.com/protobuf-net/protobuf-net.Grpc/issues/11

PR #826, opened 2020-03-18
Sync and up test credentials from grpc main repo

jiangtaoli2016:

PR #855, opened 2020-04-06
Update version to 2.28.0

jtattermusch:


jtattermusch:
The interop tests failure seems to be due to the updated certificates from https://github.com/grpc/grpc-dotnet/pull/826 (which would need to be backported to the v2.28.x branch) https://source.cloud.google.com/results/invocations/fe680af3-1d6e-4b93-af8a-e61643c1a881/targets/github%2Fgrpc%2Finterop_test/tests

PR #865, opened 2020-04-10
Change global options to be configured with UseGrpcWeb method

JamesNK:
Removes AddGrpcWeb method. Not a breaking change as package is still prerelease.


src/Grpc.AspNetCore.Web/GrpcWebOptions.cs
```diff
@@ -26,6 +26,6 @@ public class GrpcWebOptions
         /// <summary>
         /// Gets or sets a flag indicating whether gRPC-Web is enabled by default for endpoints that have not specifically opted in or out.
         /// </summary>
-        public bool GrpcWebEnabled { get; set; }
```
JunTaoLuo, line 29: Why the naming change?

JamesNK: I think it is more descriptive. It sets the default enable state. Do you disagree?

JunTaoLuo: I would have gone with something like EnableByDefault but it's not a big deal.

PR #882, opened 2020-04-24
Add SubdirectoryHandler for injecting a subdirectory into gRPC calls

JamesNK:


JamesNK:
Half sample, half verification that the same works end-to-end.

PR #885, opened 2020-04-27
Fix flaky client cancellation tests

JamesNK:
Logging is not sync and there is no easy hook to await so put test in a helper that will retry a couple of times.


davidfowl:
What does “logging is not sync” mean here?

JamesNK:
The error is being logged in a catch statement. In this situation the cancellation token passed to HttpClient.SendAsync is canceled and an error is thrown. SendAsync is not throwing that error synchronously so the catch is reached after the assert.

PR #891, opened 2020-05-02
Add gRPC benchmark client

JamesNK:
Includes changes from https://github.com/grpc/grpc-dotnet/pull/890

Moves client from aspnet/benchmarks to this repo.

This will eventually delete BenchmarkClient project.


perf/benchmarkapps/GrpcClient/grpc-client.yml
```diff
@@ -0,0 +1,18 @@
+﻿jobs:
+  grpc-client:
+    source:
+      repository: https://github.com/jamesnk/grpc-dotnet.git
+      branchOrCommit: jamesnk/interoptestclient
+      project: perf/benchmarkapps/GrpcClient/GrpcClient.csproj
+    readyStateText: gRPC Client
+    runtimeVersion: '5.0.*'
```
sebastienros, line 8: I see that you copied from the signalr client, but we'll change it eventually as I prefer these to not vary on each run by default, just when specified on the CI runs.

perf/benchmarkapps/GrpcClient/GrpcClient.csproj
```diff
@@ -0,0 +1,43 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <Protobuf Include="..\Shared\benchmark_service.proto" GrpcServices="Client" Link="Protos\benchmark_service.proto" />
+    <Protobuf Include="..\Shared\messages.proto" GrpcServices="Client" Link="Protos\messages.proto" />
+    
+    <ProjectReference Include="..\..\..\src\Grpc.Net.Client.Web\Grpc.Net.Client.Web.csproj" />
+    <ProjectReference Include="..\..\..\src\Grpc.Net.Client\Grpc.Net.Client.csproj" />
+
+    <PackageReference Include="Google.Protobuf" Version="$(GoogleProtobufPackageVersion)" />
```
sebastienros, line 22: If that's useful you can also emit these versions with the event source such they are available in the charts/db.

perf/benchmarkapps/GrpcClient/BenchmarksEventSource.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System.Diagnostics.Tracing;
+
+namespace GrpcClient
+{
+    internal sealed class BenchmarksEventSource : EventSource
```
sebastienros, line 5: Where do you use it? I couldn't find any metric exposed.

perf/benchmarkapps/GrpcClient/grpc-client.yml
JamesNK: So remove runtime version?

perf/benchmarkapps/GrpcClient/BenchmarksEventSource.cs
JamesNK: https://github.com/JamesNK/grpc-dotnet/blob/jamesnk/benchmarch-grpcclient/perf/benchmarkapps/GrpcClient/Program.cs

???

perf/benchmarkapps/GrpcClient/grpc-client.yml
JamesNK: All the other client yml files are specifying a runtime, e.g. wrk.yml

```
sdkVersion: 3.1.102
aspNetCoreVersion: '3.1'
runtimeVersion: '3.1'
```

Tell me what to put here and I'll change it.

perf/benchmarkapps/GrpcClient/BenchmarksEventSource.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System.Diagnostics.Tracing;
```
JunTaoLuo, line 1: Copyright

perf/benchmarkapps/GrpcClient/grpc-client.yml
sebastienros: These values would be fine

PR #896, opened 2020-05-05
Add GrpcChannelOptions.HttpHandler and channel default to invoker

JamesNK:
* Add GrpcChannelOptions.HttpHandler
* Change channel to default to using HttpMessageInvoker instead of HttpClient
* Channel will dispose active calls when it is disposed
* Minor perf improvements (headers Add to TryAddWithoutValidation)

@stephentoub this applies some of your perf changes from https://github.com/JamesNK/Http2Perf/pull/1

After:
```
-------------- |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 SayHelloAsync | 9.730 us | 0.1655 us | 0.1382 us | 102,776.2 |      0.0763 |           - |           - |             6.05 KB |
```

Before:
```
-------------- |---------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 SayHelloAsync | 7.860 us | 0.1571 us | 0.1746 us | 127,222.9 |      0.0610 |           - |           - |             5.25 KB |
```

### DIFF
--- a/src/Grpc.Net.Client/GrpcChannel.cs
+++ b/src/Grpc.Net.Client/GrpcChannel.cs
@@ -19,6 +19,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net.Http;
 using System.Threading;
 using Grpc.Core;
@@ -65,7 +66,7 @@ namespace Grpc.Net.Client
         {
             _methodInfoCache = new ConcurrentDictionary<IMethod, GrpcMethodInfo>();
 
-            // Dispose the HttpClient if...
+            // Dispose the HTTP client/handler if...
             //   1. No client/handler was specified and so the channel created the client itself
             //   2. User has specified a client/handler and set DisposeHttpClient to true
             _shouldDisposeHttpClient = (channelOptions.HttpClient == null && channelOptions.HttpHandler == null)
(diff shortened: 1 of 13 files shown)


src/Grpc.Net.Client/Internal/StreamExtensions.cs
```diff
@@ -107,16 +107,34 @@ private static async Task<(int length, bool compressed)?> ReadHeaderAsync(Stream
                 // If the message is larger then the array will be replaced when the message size is known.
                 buffer = ArrayPool<byte>.Shared.Rent(minimumLength: 4096);
 
-                var headerDetails = await ReadHeaderAsync(responseStream, buffer, cancellationToken).ConfigureAwait(false);
+                int read;
```
JunTaoLuo, line 110: Looks like the ReadHeaderAsync method is no longer used anywhere else after you relocated the logic. I think we can remove it.

stephentoub:
> 9.730 us => 7.860 us

Nice!

src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -90,21 +95,29 @@ internal GrpcChannel(Uri address, GrpcChannelOptions channelOptions) : base(addr
...
-            //
-            // gRPC deadline should be the recommended way to timeout gRPC calls.
-            //
-            // https://github.com/dotnet/corefx/issues/41650
-            httpClient.Timeout = Timeout.InfiniteTimeSpan;
-
-            return httpClient;
+            // HttpMessageInvoker should always dispose handler if Disposed is called on it.
+            // Decision to dispose invoker is controlled by _shouldDisposeHttpClient.
+            var httpInvoker = new HttpMessageInvoker(handler ?? new HttpClientHandler(), disposeHandler: true);
```
stephentoub, line 102: What are the expectations around proxies with the library?  If the expectation is that this will pick up the system's proxy settings, then this is fine.  If that's not needed, there's potentially some additional perf to be had by setting UseProxy = false on the handler.

Same question about redirects.  Is the expectation that HTTP redirects will automatically be followed?  If no, you can reduce some overhead by setting AllowAutoRedirect to false.

JamesNK: I think both of those are wanted. Sticking with defaults will avoid surprise, and there is always the option to turn them off if you want.

Do they have much performance impact?

stephentoub: UseProxy can, in particular on Windows; depends on Windows settings.

AllowAutoRedirect has a small per-request overhead, but probably not noticeable for anything other than a memory trace.

PR #900, opened 2020-05-09
Add extension methods for configuring all gRPC clients

bastianeicher:
- Add `.ConfigureAllGrpcClients` extension method.
  This is useful e.g. for globally registering gRPC client interceptors.
- Add `AddGrpcCallContextPropagation` extension method. This is an alternative to `EnableCallContextPropagation` that applies to call gRPC clients instead of a specific one.

### DIFF
--- a/src/Grpc.AspNetCore.Server.ClientFactory/GrpcServerServiceExtensions.cs
+++ b/src/Grpc.AspNetCore.Server.ClientFactory/GrpcServerServiceExtensions.cs
@@ -1,0 +1,37 @@
+﻿using System;
+using Grpc.AspNetCore.ClientFactory;
+using Grpc.Core;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace Grpc.AspNetCore.Server.ClientFactory
+{
+    /// <summary>
+    /// Extension methods for <see cref="IServiceCollection"/>.
+    /// </summary>
+    public static class GrpcServerServiceExtensions
+    {
+        /// <summary>
+        /// Configures the server to propagate values from a call's <see cref="ServerCallContext"/>
+        /// onto all gRPC clients.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/>.</param>
+        /// <returns>The service collection.</returns>
+        public static IServiceCollection AddGrpcCallContextPropagation(this IServiceCollection services)
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            services.TryAddSingleton<ContextPropagationInterceptor>();
+            services.AddHttpContextAccessor();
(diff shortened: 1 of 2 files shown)


JamesNK:
I'm not a fan of this approach. There is no equivalent of "configure all" methods on the underlying `HttpClientFactory`, which is what gRPC extends.

What I would recommend for applying configuration to multiple clients in a reusable way is to simply have a method that takes `IHttpClientBuilder`.

```cs
public IHttpClientBuilder ApplyMyConfiguration(this IHttpClientBuilder builder)
{
    // Do whatever you want here
    return builder;
}

services
    .AddGrpcHttpClient<GreeterClient>(o => o.Address = "https://localhost:5001")
    .ApplyMyConfiguration();
```

bastianeicher:
Hi @JamesNK,

thank you for the feedback. Unfortunately your proposed solution does not quite achieve the effect I'm looking for. My goal is to be able to globally register gRPC client interceptors in a way roughly similar to global gRPC server interceptors.

For example, to add [OpenTracing gRPC Instrumentation](https://github.com/opentracing-contrib/csharp-grpc):

```csharp
services.AddGrpc(options =>
    options.Interceptors.Add<ServerTracingInterceptor>());
services.ConfigureAllGrpcClients((options, provider) =>
    options.Interceptors.Add(provider.GetRequiredService<ClientTracingInterceptor>()));
```

Another use case: I have a library that provides an extension method like `.AddMyClients()` which only takes an URI as the input and internally calls `.AddGrpcClient()` multiple times for different services. `.ConfigureAllGrpcClients()` would allow me to "inject" cross-cutting configuration into the resutling gRPC clients without having to change this library.

bastianeicher:
At first I hoped I could achieve the desired effect without any changes to `Grpc.Net.ClientFactory` by calling this:

```csharp
services.TryAddSingleton<ContextPropagationInterceptor>();
services.AddTransient<IConfigureOptions<GrpcClientFactoryOptions>>(services =>
    new ConfigureNamedOptions<GrpcClientFactoryOptions>(null, options =>
        options.Interceptors.Add(services.GetRequiredService<ContextPropagationInterceptor>())));
```

But since `ContextPropagationInterceptor` is `internal` this does not build. I'm currently using this somewhat hacky workaround instead:

```csharp
services.ConfigureAll<GrpcClientFactoryOptions>(_ => {});
new NullHttpClientBuilder(services).EnableCallContextPropagation();

// ....

class NullHttpClientBuilder : IHttpClientBuilder
{
    public string? Name => null;

    public IServiceCollection Services { get; }

    public NullHttpClientBuilder(IServiceCollection services)
    {
        Services = services;
    }
}
```

I hope you'll reconsider my PR, or alternatively change the visiblity of `ContextPropagationInterceptor` to `public`.

JunTaoLuo:
@JamesNK We should decide what to do with this PR. If we don't plan on taking it, we should just close.

JunTaoLuo:
We don't plan on taking this feature.

PR #909, opened 2020-05-15
Fixed AsyncStreamReaderExtensions.ReadAllAsync argument validation

gfoidl:
The check for `streamReader == null` was performed on iterating. With this change the check is performed when getting the `IAsyncEnumerable` (eager), i.e. when `ReadAllAsync` is called.

It's an extension method, so it's rather unlikely to be called as `null.ReadAllAsync()`, but to be on the safe side an fail early I believe this should get in :wink:

### DIFF
--- a/src/Grpc.Net.Common/AsyncStreamReaderExtensions.cs
+++ b/src/Grpc.Net.Common/AsyncStreamReaderExtensions.cs
@@ -34,13 +34,18 @@ namespace Grpc.Core
         /// <param name="streamReader">The stream reader.</param>
         /// <param name="cancellationToken">The cancellation token to use to cancel the enumeration.</param>
         /// <returns>The created async enumerable.</returns>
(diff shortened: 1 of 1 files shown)


thelinuxfoundation:

* User @gfoidl isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

JamesNK:
Moving the `EnumeratorCancellation` is interesting. This still works:

```cs
var cts = new CancellationTokenSource();

await foreach (var item in call.ResponseStream.ReadAllAsync().WithCancellation(cts.Token))
{
    messages.Add(item.Message);

    cts.Cancel();
}
```

What is suppose to happen if someone specifies both?

```cs
var cts1 = new CancellationTokenSource();
var cts2 = new CancellationTokenSource();

await foreach (var item in call.ResponseStream.ReadAllAsync(cts1.Token).WithCancellation(cts2.Token))
{
    messages.Add(item.Message);

    cts1.Cancel();
}
```

JamesNK:
I tested and it is fine. The state machine honors both.

gfoidl:
The C#-compiler guys did a brilliant job 👍 
Thanks.

Issue #915, opened 2020-05-20
Client: server-streaming does not collect duplicated (fault+ctx) trailers during an RpcException


mgravell:
Scenario: the server (native in this case) deliberately faults with an `RpcException` including some trailers in the middle of a server-streaming operation, and that trailer has also been added to the server context, i.e.

``` c#
var faultTrailers = new Metadata();
faultTrailers.Add("sum", sum);
// **and** importantly
serverContext.ResponseTrailers.Add("sum", sum);
throw new RpcException(new Status(StatusCode.Internal, "oops"), faultTrailers);
```

## Expectation

as observed with the native Google client:

The client-side call `await streamReader.MoveNext(cancellationToken)` will surface an `RpcException` with the correct `StatusCode`, `Detail` and `Trailers` i.e. one of `"sum", sum`

## Actual

as observed with Grpc-Dotnet

The client-side call `await streamReader.MoveNext(cancellationToken)` surfaces an `RpcException` with the correct `StatusCode`, `Detail` - but:

- sometimes the `Trailers` collection is empty
- sometimes the `Trailers` collection has it *twice*, i.e. `"sum", "10,10"`

I haven't figured out the rules for when it is empty and when it is duplicated; when I run *just* the one test, it tends to be empty; when I run a bunch of tests it tends to be duplicated. Possibly a thread race condition.

## Repro

See the `ServerStreaming(true)` test case [here](https://github.com/protobuf-net/protobuf-net.Grpc/blob/master/tests/protobuf-net.Grpc.Test.Integration/StreamTests.cs) - comparing `NativeStreamTests` to `ManagedStreamTests`.

A good place to put a breakpoint is in [`Reshape.cs`](https://github.com/protobuf-net/protobuf-net.Grpc/blob/master/src/protobuf-net.Grpc/Internal/Reshape.cs), `ServerStreamingAsyncImpl`, on the line:

``` c#
metadata?.SetTrailers(fault);
```

(which is where the exception is surfaced)

Note that this is the exact same (native) server in both tests, so presumably this is a client bug.

Note: I have added `IsManagedClient` that changes the test's expectation to reflect what I am observing (so my tests acknowledge reality for now)

JamesNK:
@mgravell 2.30.0 is on NuGet with this fix. You can update your package reference and remove any related hacks.

PR #918, opened 2020-05-21
Fix sending and receiving duplicate header names

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/915

PR fixes concatenated values. I also noticed the server wasn't sending duplicate response headers.

I couldn't replicate having no headers. If might be fixed in 2.29.0-pre1.

@mgravell

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
@@ -364,11 +364,11 @@ namespace Grpc.AspNetCore.Server.Internal
                     {
                         if (entry.IsBinary)
                         {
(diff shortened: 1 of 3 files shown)


mgravell:
Awesome. I merged my changes to update to 2.29.0-pre1 earlier today - will re-run my tests tomorrow.

Note: right now, I have both clients Vs the Google server, which has found a couple of deltas. Would it be useful for you if I try to update my test rig to do the full matrix of clients vs servers? Honestly, it was mostly done this way because it is easier to run the Google server in a fixture, but I could look at running Kestrel in a fixture.

JamesNK:
Sure, if it isn't too much time for you, and it is useful for you as well then go ahead.

I'm not worried about being exactly identical with StatusCodes and detail messages, but differences around content and metadata are important to find.

PR #923, opened 2020-05-23
Create ArrayBufferWriter with predefined length if possible

danielmarbach:
I'm new to this code base. I tried to verify the assumptions in the TODO but I couldn't really so I figured I'll use the condition to check if the payload length is initialized for now.

### DIFF
--- a/src/Grpc.Net.Client/Internal/DefaultSerializationContext.cs
+++ b/src/Grpc.Net.Client/Internal/DefaultSerializationContext.cs
@@ -113,10 +113,7 @@ namespace Grpc.Net.Client.Internal
 
         private IBufferWriter<byte> ResolveBufferWriter()
         {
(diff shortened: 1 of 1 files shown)


thelinuxfoundation:

* User @danielmarbach isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

danielmarbach:
I signed it

JamesNK:
Thanks, but I don't think this is useful at the moment. We're not using `IBufferWriter<bytes>` yet. An update to the Protobuf serializer to support that API is happening now. Until then an array is allocated elsewhere to write to.

When we do use a buffer writer I will probably write a custom writer where the underlying array comes from an array pool.

Issue #842, opened 2020-03-30
Grpc Client - Interceptor - only created once (singleton)


rizi:
I've created an interceptor for gRPC clients like it's described here: https://docs.microsoft.com/en-us/aspnet/core/grpc/clientfactory?view=aspnetcore-3.1 and it's working fine.

However I want that the inceptor (in my sample below it's called MyClientInterceptor) is created for each request, at the moment it's created on the first call and all subsequent calls using the same instance of MyClientInterceptor.

I also registered MyClientIncepter in Unity 5 (I use Microsoft Unity as IOC) as transient, but that doesn't help because .AddInterceptor() adds the interceptor to a collection instead of resolving it for each request.

Is there some way I can archive this?
I want/need to do this because I have to resolve some other instances from the current ioc-container (the child container that's automatically created for each request).

I've registered the interceptor like this (there are also some overloads for interceptor but the beaviour is the same):

```
services
    .AddGrpcClient<Greeter.GreeterClient>(o =>
    {
        o.Address = new Uri("https://localhost:5001");
    })
    .AddInterceptor<MyClientInterceptor>();
```

br

JamesNK:
Also register MyClientInterceptor as a singleton:

```cs
services.AddSingleton<MyClientInterceptor>();
```

rizi:
> Also register MyClientInterceptor as a singleton:
> 
> ```cs
> services.AddSingleton<MyClientInterceptor>();
> ```

But I want exactly the opposite of a singleton, I want that MyClientIntercetor is created every time(transient). But it seems to register it transient has no effect, maybe I'm doing something wrong?

JamesNK:
I see. I read your question wrong. I'll take a look.

JamesNK:
Currently transient interceptors are not supported. We might be able to add support for them in a future version of the gRPC client factory.

rizi:
> Currently transient interceptors are not supported. We might be able to add support for them in a future version of the gRPC client factory.

Thx for the fast response.

gsmeets:
@JamesNK are singletons the only supported interceptor on the client? I'm trying to add a scoped interceptor, but that also doesn't seem to work.

I'm running into this issue porting my own NET Core 2.2 implementation to 3.1 now, removing as much of my own stuff as possible. It's a fairly common use case to want to piggy back some request based meta data from for instance an web request onto the grpc request.

JamesNK:
> are singletons the only supported interceptor on the client?

[...]

PR #953, opened 2020-06-15
Refactor client factory to support per-instance interceptors and scoped DI

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/842

This PR makes it possible to have interceptor instances created for each client instance created.

Note that there is a behavior change. Previously a single interceptor instance was the default. The new default behavior is to create a new interceptor each time. Lifetime can be controlled by explicitly defining how an interceptor is created by DI.

```cs
services
  .AddGrpcClient<Greeter.GreeterClient>(o => o.Address = new Uri("http://localhost"))
  .AddInterceptor<CallbackInterceptor>();

// Make interceptor a singleton
services.AddSingleton<CallbackInterceptor>();
```


JamesNK:
I want to wait on merging this PR.

I'm worried about GrpcChannelOptions and how we currently recreate the channel for each client. That works today because we reuse a HttpClient, which is were most of the cost is, but I worry about future features like load balancing. Load balancing has state so we'd want to persist a channel for the lifetime of the HttpClient. Should ensure this problem is solved before committing this PR and potentially putting ourselves in a corner.

JamesNK:
Will revisit in the future.

diogonborges:
Sorry if I'm being redundant but does this mean that currently the only supported Lifetime is Singleton?

murrays:
Guys this is causing us real issues. The problem is that we can't access the current 'scope' to manage per request options like tenancy.  We are intercepting client calls and the intention was to pull from a scoped object containing per request items and stuffing  into the headers.  However when we inject the scoped object (i.e. added to container via TryAddScoped() we always get the global instance.  I am now having to get clients to add in header stuff manually which is error prone and not that great.  James, are you still looking at a fix for this?

JamesNK:
I'll revisit this once load balancing is supported.

If using the client factory isn't a good solution then create channels and clients, and register them in DI, yourself.

Issue #960, opened 2020-06-22
GRPC synchronous call blocking infinitely (using Grpc.Net.ClientFactory)


timmi-on-rails:
I have prepared a minimal sample project that demonstrates the problem, here: https://github.com/timmi-on-rails/grpc/raw/master/GrpcClientFactory.zip

Basically I use Microsofts Dependency Injection Framework in a WPF app.
Then I use the extension AddGrpcClient to add the GrpClient to my service collection.
Finally calling the service (synchronously) on button click will block forever.
Doing the same using the async overloads works as expected.

### What version of gRPC and what language are you using?
2.29.0
C#

### What operating system (Linux, Windows,...) and version?
Windows 10 (1809)

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
dotnet Version 3.1.4

### What did you do?
See sample project: https://github.com/timmi-on-rails/grpc/raw/master/GrpcClientFactory.zip

### What did you expect to see?
The sync call returning like in a console application.

### What did you see instead?
Endless blocking call.

timmi-on-rails:
I found out some more details, that should help finding the problem:

1. The infinite blocking call is line 472 in https://github.com/grpc/grpc-dotnet/blob/v2.29.0/src/Grpc.Net.Client/Internal/GrpcCall.cs
So basically the returned Task of method SendAsync of HttpClient class does not complete.

2. When I supply my own HttpClient instance, the problem goes away. Is there something wrong related to the IHttpClientFactory?
```csharp
            services.AddGrpcClient<Greeter.GreeterClient>(options =>
                {
                    options.Address = new Uri("https://localhost:5001");
                }).ConfigureChannel(o =>
                {
                    o.HttpClient = new System.Net.Http.HttpClient();
                });
``` 

Any ideas?

timmi-on-rails:
Found it!

There is a bug in "Microsoft.Extensions.Http", see here: https://github.com/dotnet/extensions/issues/999
If I update the version to 3.0.0 (which is the first version with a bugfix), the problem is solved.

For testing I changed https://github.com/grpc/grpc-dotnet/blob/v2.29.0/src/Grpc.Net.ClientFactory/Grpc.Net.ClientFactory.csproj
in line 17.

Before the version was 2.1.1, which is defined in the dependencies.props file.

Please confirm and update the package version.

JamesNK:
Good investigation.

[...]

PR #962, opened 2020-06-25
Update extensions dependencies to 3.0.3

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/960

@JunTaoLuo @shirhatti I'm sure what package version to update to. The fix is in `3.0.x`, but that is out of support. Our packages target 3.0 so `3.1.x` is probably not appropriate. Thoughts?


JunTaoLuo:
If 3.0.3 fixes the issue then I think we should go with that for now. Is there any plans to ever update our dependency to a different major/minor due to support reasons? Wouldn't that constitute a breaking change (if we update to a new major)?

JamesNK:
We'll probably support netcoreapp3.0 for a long long time. Yes it would be a breaking change.

JamesNK:
@JunTaoLuo Like and subscribe 🙏

shirhatti:
Can we multi-target? I understand not wanting to break netcoreapp3.0 customers but it feels awful taking a dependency on an unsupported package for netcoreapp3.1+

PR #973, opened 2020-07-09
Use latest version of OpenTelemetry with gRPC in sample

JamesNK:
Updated packages, startup API, and use new gRPC instrumentation.


JamesNK:
Might wait until packages are on NuGet before merging this.

PR #985, opened 2020-07-16
Add an experimental github actions workflow.

jtattermusch:


jtattermusch:
I'm going to merge this just to try out github actions (looks like the PR needs to be merged for the action to become active). If things don't work, I'm going to fix/revert.

Issue #988, opened 2020-07-18
Sharing and reusing of protos in large projects next to impossible, help!


ChiefInnovator:
### Sharing and reusing of protos in large projects next to impossible, help!

We are in the midst of a large microservice implementation with gRPC.  We want to share common protos across several services.  Examples include address, GPS coordinates, phone number, version, person, and more.  Basically any data that can be common across a service we do not want to be redefined in the service.

For example, say we had an Address proto and did not define an addressline3 field.  We do not want to go to every service project and have to add addressline3 to that proto.

So we came up with three approaches to share protos.  Unfortunately, none solves the problem completely.  Translation, they do not solve the problem.

**Option 1 - Have a library protos project where all our protos live.**
We stick the protos in a shared library and use NuGet package Kungfu to copy the protos into the consuming project.  This worked great for sharing the packages, but failed to deliver due to assembly conflicts.

Service A and Service B consume the protos. Our clients for these services are defined in separate assemblies.  The challenge is that a consuming application may using clients from both Service A and Service B.  In that case, .NET does not know which types to use and blows up at us.

This also had the following complexity:
- Oversharing of protos that are not needed in Service B, but needed in Service A, and vice-versa.
- Cannot distinguish between protos from private services (i.e. backend) and public services (public APIs).
- Client and service projects would get all protos and you would have to go through each one and select whether they would be compiled and whether you wanted Client, Server, or Both classes generated.

**Option 2 - We share our protos using GitHub Submodules**
This is really no different that Option #1, just a different mechanism for sharing the protos.  It has all the same challenges as well as the added complexity of doing this in source control.

Developer - "Architect! Architect! It hurts when I do this!
Architect - "Don't do it"

**Option 3 - Define a common protos in shared NuGet library, define service specific protos in services**
The idea is that we would have a shared assembly just for our common types. The thinking is that we would have a shared NuGet package which had our compiled common protos. Then we would have services define those types specific to their services.  We got this idea from how the gRPC tooling sharing the Google Wellknown Protos. Unfortunately this is a challenge too. In our attempts, the service protos need access to the common proto files, not just the compiled protos.

### Summary
We are going to try a fourth option which is a common repo with multiple projects to mirror all our services and clients.

**Protos Repo**
protos\common\address.proto
protos\common\person.proto
protos\product\productspecific.proto
protos\serviceA\serviceAspecific.proto
protos\serviceB\serviceBspecific.proto
[...]

PR #992, opened 2020-07-21
Add liber example

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/988

Example of how to share proto messages in common .NET libraries.

### DIFF
--- a/examples/Liber/Common/Common.csproj
+++ b/examples/Liber/Common/Common.csproj
@@ -1,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Protobuf Include="..\Proto\common.proto" Link="Protos\common.proto" />
+
+    <PackageReference Include="Google.Protobuf" Version="$(GoogleProtobufPackageVersion)" />
+    <PackageReference Include="Grpc.Tools" Version="$(GrpcPackageVersion)" PrivateAssets="All" />
+  </ItemGroup>
+
+</Project>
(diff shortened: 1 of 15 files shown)


examples/Liber/Common/Common.csproj
```diff
@@ -0,0 +1,14 @@
...
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Protobuf Include="..\Proto\common.proto" Link="Protos\common.proto" />
+
+    <PackageReference Include="Google.Protobuf" Version="$(GoogleProtobufPackageVersion)" />
+    <PackageReference Include="Grpc.Tools" Version="$(GrpcPackageVersion)" PrivateAssets="All" />
+  </ItemGroup>
```
chwarr, line 12: On my team, we make sure to include the .proto files in the NuGet packages as an escape hatch for advanced consumers (e.g., custom code gen options). This has been a Good Thing<sup>TM</sup> for us, and I think the ecosystem would benefit if this example did something similar.

We use something like this in the .csproj files to pack them in the .nupkg:

```xml
<ItemGroup>
  <None Include="@(ProtoBuf)">
    <CopyToOutputDirectory>Never</CopyToOutputDirectory>
    <Pack>true</Pack>
    <!-- We use MakeRelative to replicate the input directory hierarchy under the build\proto directory. -->
    <PackagePath>build\proto\$([MSBuild]::MakeRelative($(ProtoIncludeFolder), %(Identity)))</PackagePath>
  </None>
</ItemGroup>
```

In this example, `$(ProtoIncludeFile)` would be "..\Proto". (This snippet is actually from a Directory.Build.targets in our shared contracts with C# enhancements repository. If you want more details, email me at my Microsoft address and I can point you at the internal repo.)

JamesNK: Interesting. I'll leave it as is for the moment. When I have more time available to research best practices for publishing NuGet packages I'll look into this more.

PR #996, opened 2020-07-22
Update to use prerelease Grpc.Tools and Google.Protobuf

JamesNK:
Will currently fail because packages aren't on a NuGet feed yet.


build/sources.props
```diff
@@ -8,6 +8,7 @@
       https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json;
       https://dotnet.myget.org/F/roslyn/api/v3/index.json;
+      C:\Development\Source\grpc-dotnet\packages;
```
JunTaoLuo, line 11: I assume we don't want this?

JamesNK: yeah

JamesNK:
Replaced with https://github.com/grpc/grpc-dotnet/pull/1031

PR #1007, opened 2020-08-02
Fix benchmark client incorrectly measuring warmup RPS

JamesNK:


perf/benchmarkapps/GrpcClient/Program.cs
```diff
@@ -112,6 +112,7 @@ private static async Task StartScenario()
             var cts = new CancellationTokenSource();
             cts.CancelAfter(TimeSpan.FromSeconds(_options.Duration + _options.Warmup));
 
+            _warmingUp = true;
```
davidfowl, line 115: Is this volatile?

PR #1031, opened 2020-08-26
Update Grpc.Core and Google.Protobuf dependencies

JamesNK:
Use new Protobuf APIs with Grpc.Tools codegen and Google.Protobuf package.


JamesNK:
Code generation is updated with new Protobuf codegen (use IBufferMessage) but not new Grpc.Tools codegen (use Span APIs in marshallers).

Will update the GrpcPackageVersion tomorrow before merging.

Issue #784, opened 2020-02-20
Optimize header allocation when using array writer


JamesNK:
See https://github.com/grpc/grpc-dotnet/pull/775

Header allocation when writing messages in the client could be optimized to be provider but IBufferWriter.

PR #1033, opened 2020-08-30
Client serialization optimizations

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/784

* Reuse serialization contexts when streaming
* Initialize write buffer initial capacity
* Write directly to pooled array when length is known and there is no compression


src/Grpc.Net.Client/Internal/GrpcCallSerializationContext.cs
```diff
@@ -0,0 +1,321 @@
...
+        }
+
+        public override void Complete(byte[] payload)
+        {
+            switch (_state)
+            {
+                case InternalState.Initialized:
+                    _state = InternalState.CompleteArray;
+
+                    GrpcCallLog.SerializedMessage(_call.Logger, typeof(object), payload.Length);
```
JunTaoLuo, line 155: Should this use _call.RequestType?

Issue #1036, opened 2020-09-02
HttpClient created by client factory should allow multiple HTTP/2 connections


JamesNK:
When available, client factory should use SocketsHttpHandler with enable multiple HTTP/2 connections enabled.

PR #1037, opened 2020-09-02
Client factory handler supports multiple HTTP/2 connections

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1036

### DIFF
--- a/src/Grpc.Net.Client/GrpcChannel.cs
+++ b/src/Grpc.Net.Client/GrpcChannel.cs
@@ -115,30 +115,6 @@ namespace Grpc.Net.Client
             return httpInvoker;
         }
 
-#if NET5_0
-        private static bool IsSocketsHttpHandler(HttpMessageHandler handler)
-        {
-            if (handler is SocketsHttpHandler)
-            {
-                return true;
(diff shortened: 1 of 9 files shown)


src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -100,31 +101,11 @@ private static HttpMessageInvoker CreateInternalHttpInvoker(HttpMessageHandler?
...
 
 #if NET5_0
-            // HttpClientHandler has an internal handler that sets request telemetry header.
-            // If the handler is SocketsHttpHandler then we know that the header will never be set
-            // so wrap with a telemetry header setting handler.
-            if (IsSocketsHttpHandler(handler))
-            {
-                handler = new TelemetryHeaderHandler(handler);
-            }
+            handler = HttpHandlerFactory.EnsureTelemetryHandler(handler);
```
JunTaoLuo, line 108: So the IsSocketsHttpHandler method can be removed in this class?

JunTaoLuo (approved):
One cleanup nit

PR #1001, opened 2020-07-29
Bump Grpc.Core dependency to 2.31.0-pre2

jtattermusch:
I'll cut the release branch v2.31.x after this.

PR #1041, opened 2020-09-03
Bump version to 2.32.0-pre1

apolcyn:
Following https://github.com/grpc/grpc-dotnet/pull/1001 as an earlier example

### DIFF
--- a/build/version.props
+++ b/build/version.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
 
     <!-- package version of grpc-dotnet -->
-    <GrpcDotnetVersion>2.31.0-dev</GrpcDotnetVersion>
+    <GrpcDotnetVersion>2.32.0-pre1</GrpcDotnetVersion>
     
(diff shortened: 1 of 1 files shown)


build/version.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
 
     <!-- package version of grpc-dotnet -->
-    <GrpcDotnetVersion>2.31.0-dev</GrpcDotnetVersion>
+    <GrpcDotnetVersion>2.32.0-pre1</GrpcDotnetVersion>
```
JamesNK, line 5: GrpcDotnetAssemblyFileVersion in this file should be 2.32.0.0 as well

JamesNK (changes requested):

build/version.props
apolcyn: thanks, done

JamesNK (approved):

PR #1045, opened 2020-09-09
Fix repeatedly decoding base64 with large grpc-web-text request

JamesNK:
Large base64 encoded messages are slow to decode. Each time ReadAsync is called all the buffered data is re-decoded.

This PR changes base64 decoding to be incremental. Only new data is decoded and it is appended to a sequence of decoded data.

3 megabyte message, IIS host:
* Before - 6000 ms
* After - 100 ms

### DIFF
--- a/src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
+++ b/src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
@@ -16,11 +16,9 @@
 
 #endregion
 
-using Microsoft.AspNetCore.Authorization.Infrastructure;
 using System;
 using System.Buffers;
 using System.Buffers.Text;
-using System.Diagnostics;
 using System.IO.Pipelines;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
```diff
@@ -158,31 +170,57 @@ public async override ValueTask<ReadResult> ReadAsync(CancellationToken cancella
             }
 
             // Limit result to complete base64 segments (multiples of 4)
-            var buffer = innerResult.Buffer.Slice(0, (innerResult.Buffer.Length / 4) * 4);
+            var newResultLength = innerResult.Buffer.Length - _currentInnerBufferRead;
+            var newResultValidLength = (newResultLength / 4) * 4;
```
JunTaoLuo, line 174: Curious if bitmasking would be faster than a division and a multiplication here. Not sure if the compiler is smart enough to automatically do that. Also, I would assume the effect here would be pretty small anyways, but I was just curious.

JunTaoLuo (approved):

src/Grpc.AspNetCore.Web/Internal/Base64PipeReader.cs
JamesNK: Anything here will be dwarfed by the calls to ReadAsync and base64 decoding.

PR #1046, opened 2020-09-11
Fix client to only filter known gRPC headers

JamesNK:


src/Grpc.Net.Client/Internal/GrpcProtocolHelpers.cs
```diff
@@ -91,6 +83,32 @@ public static Metadata BuildMetadata(HttpResponseHeaders responseHeaders)
...
+            {
+                case ':':
+                    // ASP.NET Core includes pseudo headers in the set of request headers
+                    // whereas, they are not in gRPC implementations. We will filter them
+                    // out when we construct the list of headers on the context.
+                    return true;
+                case 'g':
+                case 'G':
+                    // Exclude known grpc headers. This matches Grpc.Core client behavior.
+                    return string.Equals(name, GrpcProtocolConstants.StatusTrailer, StringComparison.OrdinalIgnoreCase)
```
rynowak, line 103: What's the benefit of excluding these? I get that this matches, but why?

JamesNK: Status/message have a strongly typed API for accessing them, and encoding/accept-encoding are automatically used for you by the client.

rynowak: Cool beans, but why is it **better** that they don't appear in the header collection? It surprises the heck out of me 😆

rynowak: Does it work this way for non-metadata responses?

test/FunctionalTests/Client/TrailerMetadataTests.cs
```diff
@@ -17,6 +17,7 @@
 #endregion
 
 using System.Linq;
+using System.Text;
```
rynowak, line 20: Worth having a test where casing doesn't match exactly?

JunTaoLuo (approved):

src/Grpc.Net.Client/Internal/GrpcProtocolHelpers.cs
JamesNK: > Cool beans, but why is it **better** that they don't appear in the header collection? It surprises the heck out of me 😆

My guess is they didn't want to confuse users by showing metadata that the user didn't add on the server. I don't have a strong opinion, but since Grpc.Core does this, and compatibility with Grpc.Core is a goal, we'll follow their lead until there is a reason not to.

rynowak: Understood. IMO this would make perfect sense if this was called metadata instead of headers.

PR #1060, opened 2020-09-23
Support aggressive trimming

JamesNK:
Attributes to support aggressive trimming. Will add a test app that does trimming soon.


davidfowl:
Why only 5.0?

JamesNK:
grpc-dotnet multi-targets 3.0 and 5.0, and the attributes are only present in .NET 5.

When the solution multi-targets again in the future I'll refactor `#ifdefs` to be like feature flags, i.e. `#if LINKER_ATTRIBUTES_SUPPORTED`, and define in supported targets.

davidfowl:
We added them in source to handle ns2.0 style libraries. Trimming is agnostic of the target so you can do the same here

JamesNK:
Could, but why? Everything multi-targets 5.0 so 5.0 packages will always be used in an app with trimming.

davidfowl:
You don't have to but trimming isn't coupled to the target framework

JamesNK:
Added integration test. Requires a change in Grpc.Core.Api. Blocked while waiting for it to be updated.

JunTaoLuo (approved):

PR #1073, opened 2020-10-07
Fix potential torn reads by counters

JamesNK:
32bit and ARM might have a torn read of counter `long` values. Volatile ensures correct value is always read.

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
@@ -136,31 +136,31 @@ namespace Grpc.AspNetCore.Server.Internal
                 // This is the convention for initializing counters in the RuntimeEventSource (lazily on the first enable command).
                 // They aren't disabled afterwards...
 
-                _totalCallsCounter ??= new PollingCounter("total-calls", this, () => _totalCalls)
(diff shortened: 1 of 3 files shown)


src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
```diff
@@ -136,31 +136,31 @@ protected override void OnEventCommand(EventCommandEventArgs command)
                 // This is the convention for initializing counters in the RuntimeEventSource (lazily on the first enable command).
                 // They aren't disabled afterwards...
 
-                _totalCallsCounter ??= new PollingCounter("total-calls", this, () => _totalCalls)
+                _totalCallsCounter ??= new PollingCounter("total-calls", this, () => Volatile.Read(ref _totalCalls))
```
gfoidl, line 139: It's about `Interlocked.Read` vs. `Volatile.Read`.
In ASP.NET Core for the kestrel counters I've used `Volatile.Read`, which does the correct read, then opened https://github.com/dotnet/docs/pull/20984 to update the docs. The consensus over there is to use `Interlocked.Read` as it's documented to prevent torn-reads.

I'm fine with either variant here, just wanted to point this fact out (and for completeness).

JunTaoLuo (approved):

PR #1079, opened 2020-10-18
Update Microservicer sample to net5.0

JamesNK:
Only works with 0.5 preview that hasn't been published yet. Wait until 0.5 is on NuGet.org before merging.


JunTaoLuo (approved):
Yup, we'll be able to support 5.0 when 0.5.0-preview is released.

PR #1102, opened 2020-11-04
Optimize deadline logic

JamesNK:
Deadline changes/optimizations:
* Avoid allocating request cancellation token unless accessed
* Avoid allocating deadline cancellation token unless accessed
* Changed to a TCS and only allocated/used if deadline is raised
* Remove multiple deadline implementations

Deadline set but not exceeded:

Before: 660,137
After: 698,672

No deadline set:

Before: 741,130
After: 742,572


JamesNK:
@tratcher @halter73 A deadline can get raised anytime. I've put flags+locks to ensure it doesn't trigger as the request is finishing, and the request doesn't finish as the deadline is being raised.

If you have time, could you review this PR. Particularly ServerCallDeadlineManager - https://github.com/grpc/grpc-dotnet/pull/1102/files#diff-cd2d5444511bb28ea54006227c8d3c5d0742e7b2178a882fb81770eaf1723124

src/Grpc.AspNetCore.Server/Internal/ServerCallDeadlineManager.cs
```diff
@@ -0,0 +1,252 @@
...
+            }
+        }
+
+        public ValueTask DisposeAsync()
+        {
+            // Deadline registration needs to be disposed with DisposeAsync, and the task completed
+            // before the lock can be disposed.
+            // Awaiting deadline registration and deadline task ensures it has finished running, so there is
+            // no way for deadline logic to attempt to wait on a disposed lock.
+            var disposeTask = _longDeadlineTimer.DisposeAsync();
```
halter73, line 221:
```c#
private static void DeadlineExceededCallback(object? state) => _ = ((ServerCallDeadlineManager)state!).DeadlineExceededAsync();
```

Because the DeadlineExceededCallback discards the Task returned by DeadlineExceededAsync(), awaiting Timer.DisposeAsync() does not ensure that DeadlineExceededAsync() has finished running. _deadlineCts could get disposed before a pending call to Cancel() causing an unobserved ODE.

Since TimerCallback isn't Task-returning, I think [using a Timer-backed awaitable](https://github.com/dotnet/aspnetcore/blob/8c8cb4dce235b099d1e419b6a039217eb56a77e6/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs#L1792-L1816) is the way to go.

JamesNK: The manager gets around that problem by using a TCS - _deadlineExceededCompleteTcs. If it is set then DisposeAsync will await it before disposing the CTS.

However, now that I think about it, I could just avoid using a Timer altogether and use a Task.Delay. A deadline only happens once.

```cs
private async Task StartDeadlineAsync(TimeSpan time)
{
    await Task.Delay(time);
    await DeadlineExceededAsync();
}

public ServerCallDeadlineManager(...)
{
    _deadlineTask = StartDeadlineAsync(...);
}
```

JamesNK: Hmm, but a CTS would be needed to stop the `Task.Delay`, and there would be an `OCE`.

I think the current approach would be more efficient.

halter73: Another option is to have the TimerCallback complete yet another TCS. You could then await that TCS's Task at the start of DeadlineExceededAsync(), invoke DeadlineExceededAsync() in the ServerCallDeadlineManager ctor, and await the Task returned by DeadlineExceededAsync() in DisposeAsync().

halter73: Is _deadlineExceededCompleteTcs guaranteed to be set (if it will be set) before DisposeAsync is called?

src/Grpc.AspNetCore.Server/Internal/ServerCallDeadlineManager.cs
```diff
@@ -0,0 +1,263 @@
...
+
+            _systemClock = clock;
+
+            var timerMilliseconds = GetTimerDueTime(timeout, maxTimerDueTime);
+            if (timerMilliseconds == maxTimerDueTime)
+            {
+                // Create timer and set to field before setting time.
+                // Ensures there is no weird situation where the timer triggers
+                // before the field is set. Shouldn't happen because only long deadlines
+                // will take this pass but better to be safe than sorry.
```
Tratcher, line 94:
```suggestion
                // will take this path but better to be safe than sorry.
```

src/Grpc.AspNetCore.Server/Internal/ServerCallDeadlineManager.cs
```diff
@@ -0,0 +1,263 @@
...
+
+        private CancellationTokenSource? _deadlineCts;
+        private CancellationTokenRegistration _requestAbortedRegistration;
+        private TaskCompletionSource<object?>? _deadlineExceededCompleteTcs;
+
+        public DateTime Deadline { get; }
+        public bool IsCallComplete { get; private set; }
+        public bool IsDeadlineExceededStarted => _deadlineExceededCompleteTcs != null;
+
+        public CancellationToken CancellationToken
```
Tratcher, line 47: What's actually using this token now? Just app code?

src/Grpc.AspNetCore.Server/Internal/CallHandlers/UnaryServerCallHandler.cs
```diff
@@ -54,17 +54,14 @@ protected override async Task HandleCallAsyncCore(HttpContext httpContext, HttpC
                 throw new RpcException(new Status(StatusCode.Cancelled, "No message returned from method."));
             }
 
-            if (serverCallContext.DeadlineManager != null && serverCallContext.DeadlineManager.CancellationToken.IsCancellationRequested)
+            if (serverCallContext.DeadlineManager != null && serverCallContext.DeadlineManager.IsDeadlineExceededStarted)
```
Tratcher, line 57: This IsDeadlineExceededStarted seems racy. Is that OK because you're not expecting the deadline expiration to do anything at this point, nobody should be listening to it anymore?

Tratcher (approved):
Congratulations, my brain hurts.

src/Grpc.AspNetCore.Server/Internal/ServerCallDeadlineManager.cs
JamesNK: Yes. `ServerCallContext.CancellationToken` is exposed to devs.

src/Grpc.AspNetCore.Server/Internal/ServerCallDeadlineManager.cs
```diff
@@ -0,0 +1,263 @@
...
+            // Ensure an in-progress deadline is finished before disposing.
+            // Need to await to avoid race between canceling CT and disposing it.
+            if (_deadlineExceededCompleteTcs != null)
+            {
+                await _deadlineExceededCompleteTcs.Task;
+            }
+            Dispose(true);
+        }
+
+        protected virtual void Dispose(bool disposing)
```
JunTaoLuo, line 257: nit: I don't think there are any classes that extends ServerCallDeadlineManager anymore, so does this still need to be protected?

src/Grpc.AspNetCore.Server/Internal/CallHandlers/UnaryServerCallHandler.cs
JamesNK: This part is racy by design. Basically it for the common case where a deadline has occurred while executing a service method. When that happens there is no point attempting to write the response, which would just error.

I added a comment about it.

src/Grpc.AspNetCore.Server/Internal/ServerCallDeadlineManager.cs
```diff
@@ -0,0 +1,264 @@
...
+            {
+                await _deadlineExceededCompleteTcs.Task;
+            }
+            DisposeCore();
+        }
+
+        private void DisposeCore()
+        {
+            _deadlineCts?.Dispose();
+            _requestAbortedRegistration.Dispose();
```
halter73, line 261: Since the callback associated with _requestAbortedRegistration Cancels _deadlineCts, this should probably be disposed before _deadlineCts.

src/Grpc.AspNetCore.Server/Internal/ServerCallDeadlineManager.cs
```diff
@@ -0,0 +1,264 @@
...
+            try
+            {
+                await _serverCallContext.DeadlineExceededAsync();
+                lock (this)
+                {
+                    IsCallComplete = true;
+
+                    // Canceling CTS will trigger registered callbacks.
+                    // Exception could be thrown from them.
+                    _deadlineCts?.Cancel();
```
halter73, line 196: Since cancelling _deadlineCts can invoke user-defined callbacks, it would feel safer if all the attempts to cancel it were done outside of a lock (and better yet, dispatched to the ThreadPool), but I don't know the codebase well enough to know if this is really necessary.

halter73 (approved):

JunTaoLuo (approved):

PR #1145, opened 2020-12-11
DO NOT MERGE: test-only PR

jtattermusch:
testing whether the cla/linuxfoundation check is gone


jtattermusch:
The cla/linuxfoundation check seems to be gone now.

PR #1165, opened 2021-01-10
Fix setting user-agent in client when running in the browser

JamesNK:
Fixes https://github.com/dotnet/aspnetcore/issues/29183

### DIFF
--- a/src/Grpc.Net.Client.Web/Internal/OperatingSystem.cs
+++ b/src/Grpc.Net.Client.Web/Internal/OperatingSystem.cs
@@ -1,0 +1,39 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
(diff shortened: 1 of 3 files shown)


src/Grpc.Net.Client.Web/Internal/OperatingSystem.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    internal class OperatingSystem : IOperatingSystem
+    {
+        public static readonly OperatingSystem Instance = new OperatingSystem();
+        
+        public bool IsBrowser { get; }
+                
+        private OperatingSystem()
+        {
+            IsBrowser = RuntimeInformation.IsOSPlatform(OSPlatform.Create("browser"));
```
JunTaoLuo, line 36: Huh, I didn't know it was possible to create a "browser" OS platform and it'll just work.

JunTaoLuo (approved):

PR #1167, opened 2021-01-11
Roll forward using LatestMajor behaviour

JunTaoLuo:
Fixes https://github.com/grpc/grpc-dotnet/issues

Currently the roll forward behaviour is set to Major which
> Roll forward to lowest higher major version, and lowest minor version, if requested major version is missing. If the requested major version is present, then the Minor policy is used.

This breaks when both 3.1 and 5.0 runtimes are installed as was the scenario in the linked issue. Specifically, the tool itself targets netcoreapp3.0 and ends up using the 3.1 runtime due to the Major roll forward policy. However, the MSBuild locator picks the 5.0 SDK which ends up loading a NugetSDKResolver binary that depends on 5.0 runtime assemblies. This conflict results in the assembly not found error. To resolve this, we can just rely on the LatestMajor roll forward policy, this ensures we run on the latest runtime, which will be compatible with the latest installed SDK.

### DIFF
--- a/src/dotnet-grpc/dotnet-grpc.csproj
+++ b/src/dotnet-grpc/dotnet-grpc.csproj
@@ -14,6 +14,8 @@
     <WarningsNotAsErrors>$(WarningsNotAsErrors);CA2007</WarningsNotAsErrors>
     <NoWarn>$(NoWarn);CA2007</NoWarn>
 
+    <RollForward>LatestMajor</RollForward>
+
     <!-- The runtimeconfig.template.json file is used to allow roll-forward for future runtimes. We target netcoreapp3.0 as the lowest TFM. 
       See https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md#roll-forward-in-absence-of-candidate-fx for details-->
   </PropertyGroup>
--- a/src/dotnet-grpc/runtimeconfig.template.json
+++ b/src/dotnet-grpc/runtimeconfig.template.json
@@ -1,3 +1,0 @@
-{
-  "rollForwardOnNoCandidateFx": 2
-}


JunTaoLuo:
cc @shirhatti

shirhatti:
@JunTaoLuo What's happens when I have the 3.1 SDK and 5.0 runtime installed? It's something that works before your change?

src/dotnet-grpc/dotnet-grpc.csproj
```diff
@@ -14,6 +14,8 @@
     <WarningsNotAsErrors>$(WarningsNotAsErrors);CA2007</WarningsNotAsErrors>
     <NoWarn>$(NoWarn);CA2007</NoWarn>
 
+    <RollForward>LatestMajor</RollForward>
+
     <!-- The runtimeconfig.template.json file is used to allow roll-forward for future runtimes. We target netcoreapp3.0 as the lowest TFM. 
       See https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md#roll-forward-in-absence-of-candidate-fx for details-->
```
JamesNK, line 20: Comment no longer valid as file has been removed?

```suggestion
```

src/dotnet-grpc/dotnet-grpc.csproj
```diff
@@ -14,6 +14,8 @@
     <WarningsNotAsErrors>$(WarningsNotAsErrors);CA2007</WarningsNotAsErrors>
     <NoWarn>$(NoWarn);CA2007</NoWarn>
 
+    <RollForward>LatestMajor</RollForward>
```
JamesNK, line 17: Add comment to explain why this is needed

JunTaoLuo:
@shirhatti Yes that should work. the Major roll forward policy means prefer the requested major version but roll forward to the nearest newer version if the requested version is not present. 

Suppose the 3.1 SDK is present and the 3.1 runtime that was installed along with it is also present, the tool will resolve the 3.1 SDK and run on the 3.1 runtime, which works.

Suppose the 3.1 SDK is present but the 3.1 runtime was removed (i.e. manually) and replaced 5.0 runtime was installed, the tool will still work since the 5.0 runtime contain assemblies that are newer than the ones required by the 3.1 SDK. Assuming there were no breaking 5.0 changes that affects the 3.1 SDK, it should work. I can try this scenario to double check tomorrow though it seems a little bizarre due to the numerous manual steps required to get there. Also, if 5.0 runtime was installed, I don't see why the SDK wasn't upgraded to 5.0 as well.

JamesNK (approved):

PR #1183, opened 2021-01-27
Change gRPC client factory to set HttpHandler

JamesNK:
Using HttpHandler is faster than HttpClient.

There is one behaverial change (address could also be configured on HttpClient) but that isn't the typical way of using the client factory. I've improved the error message to made it clear how to use the client factory.

### DIFF
--- a/src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
+++ b/src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
@@ -40,7 +40,7 @@ namespace Grpc.Net.ClientFactory.Internal
             _loggerFactory = loggerFactory;
         }
 
-        public CallInvoker CreateCallInvoker(HttpMessageHandler httpHandler, string name, GrpcClientFactoryOptions clientFactoryOptions)
+        public CallInvoker CreateCallInvoker(HttpMessageHandler httpHandler, string name, Type type, GrpcClientFactoryOptions clientFactoryOptions)
         {
             if (httpHandler == null)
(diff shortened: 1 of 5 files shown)


src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
```diff
@@ -40,15 +40,15 @@ public GrpcCallInvokerFactory(ILoggerFactory loggerFactory)
             _loggerFactory = loggerFactory;
         }
 
-        public CallInvoker CreateCallInvoker(HttpClient httpClient, string name, GrpcClientFactoryOptions clientFactoryOptions)
+        public CallInvoker CreateCallInvoker(HttpMessageHandler httpHandler, string name, Type type, GrpcClientFactoryOptions clientFactoryOptions)
```
JunTaoLuo, line 43: This is a breaking change isn't it?

JunTaoLuo (approved):
Changes look good but I'm not sure how we should communicate the breaking change.

src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
JamesNK: What do you mean? GrpcCallInvokerFactory is internal so its not a public API breaking change

JunTaoLuo: Duh... I for some reason thought this was public API.

PR #1187, opened 2021-02-02
gRPC client retries

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1189

This PR adds support for [gRPC retries](https://github.com/grpc/proposal/blob/master/A6-client-retries.md). When retries are enabled, a wrapping call (either `RetryCall` or `HedgingCall`) is created that will manage making HTTP requests to the server. The wrapping call will create `GrpcCall` instances that will perform the HTTP request to the server.

* All of the gRPC retries spec is supported.
* `GrpcCall` is mostly unchanged.
* `RetryCall` and `HedgingCall` (and their base type) handle retry logic, message buffering, throttling, etc. Once a gRPC call is commited then reposibility is handled back to it for returning a response.
* Strongly typed `ServiceConfig` type for client configuration. Currently it is only ever manually created by the client, but in the future deserializing from JSON/Protobuf will be supported for populating it by a naming resolver.
* Pay-for-play. gRPC calls that don't have retries enabled will use `GrpcCall` without performance impact of features required by retries.
* Clear separate between retry logic and standard HTTP gRPC calls.

---

Other bugs fixed during this PR:
* ~Fix - An error from the server to the client will now correctly update the status of in-progress server streaming and bidirectional streaming calls. This matches Grpc.Core behavior.~ https://github.com/grpc/grpc-dotnet/pull/1198
* ~Change - A completed client streaming and bidirectional streaming call in the client will throw `RpcException` from `RequestStream.WriteAsync`. This matches Grpc.Core behavior.~ https://github.com/grpc/grpc-dotnet/pull/1199
* ~Change - Don't log cancellation exception from HttpClient on deadline exceeded in client.~ https://github.com/grpc/grpc-dotnet/pull/1201

---

TODO

- [x] Hedging with exceeded deadline
- [x] Hedging with throttling
- [x] Retry limit configuration
- [x] Buffer limit configuration
- [x] Refactor configuration objects to improve intellisense
- [x] Handle `ResponseStream.MoveNext` cancellation

---

Work in progress on https://github.com/grpc/proposal/blob/master/A6-client-retries.md

```cs
var serviceConfig = new ServiceConfig
{
    MethodConfigs =
    {
        new MethodConfig
        {
            Names = { Name.AllServices },
            RetryPolicy = new RetryThrottlingPolicy
            {
                MaxAttempts = 5,
                InitialBackoff = TimeSpan.FromSeconds(1),
                RetryableStatusCodes = { StatusCode.Unavailable }
            }
        }
    }
};

var channel = GrpcChannel.Create("https://localhost:5001", new GrpcChannelOptions
{
    ServiceConfig = serviceConfig
});
var client = new Greet.GreeterClient(channel);

// Make gRPC call with retries
var reply = await client.SayHelloAsync(new HelloRequest { Name = ".NET" });
Console.WriteLine(reply.Message);
```


src/Grpc.Net.Client/Internal/Retry/RetryCall.cs
```diff
@@ -0,0 +1,357 @@
...
+                        // Call has already been commited. This could happen if written messages exceed
+                        // buffer limits, which causes the call to immediately become commited and to clear buffers.
+                        return;
+                    }
+
+                    Status status = responseStatus.Value;
+
+                    var retryPushbackMS = GetRetryPushback(currentCall);
+
+                    // Failures only could towards retry throttling if they have a known, retriable status.
```
Tratcher, line 208:
```suggestion
                    // Failures only count towards retry throttling if they have a known, retriable status.
```

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
```diff
@@ -0,0 +1,410 @@
...
+
+            ValidatePolicy(hedgingPolicy);
+        }
+
+        private void ValidatePolicy(HedgingPolicy hedgingPolicy)
+        {
+            //if (retryThrottlingPolicy.MaxAttempts == null)
+            //{
+            //    throw CreateException(_method, RetryPolicy.MaxAttemptsPropertyName);
+            //}
```
Tratcher, line 68: ?

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
```diff
@@ -0,0 +1,410 @@
...
+
+            try
+            {
+                var hedgingDelay = _hedgingPolicy.HedgingDelay.GetValueOrDefault();
+
+                while (_callsAttempted < MaxRetryAttempts)
+                {
+                    _ = StartCall(startCallFunc);
+
+                    await HedgingDelayAsync(hedgingDelay).ConfigureAwait(false);
```
Tratcher, line 268: "If a non-fatal status code is received from a hedged request, then the next hedged request in line is sent immediately, shortcutting its hedging delay."
https://github.com/grpc/proposal/blob/8a695490fed36c59d2974f1bb38ab399447769fb/A6-client-retries.md#hedging-policy

Where would that be handled?

Tratcher (commented):
The lazy config model doesn't seem like a good fit here. It results in a lot of per-request validation and parsing (sometimes repeatedly per request) that could have been done once up front when creating the channel.

src/Grpc.Net.Client/Internal/Retry/RetryCall.cs
```diff
@@ -0,0 +1,357 @@
...
+            _random = new Random();
+
+            ValidatePolicy(retryPolicy);
+
+            _nextRetryDelayMilliseconds = Convert.ToInt32(retryPolicy.InitialBackoff.GetValueOrDefault().TotalMilliseconds);
+        }
+
+        private void ValidatePolicy(RetryPolicy retryPolicy)
+        {
+            if (retryPolicy.MaxAttempts == null)
```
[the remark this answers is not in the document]
JamesNK: Moved validation to creating cached method information.

src/Grpc.Net.Client/Internal/Retry/RetryCall.cs
```diff
@@ -0,0 +1,357 @@
...
+        private int _nextRetryDelayMilliseconds;
+
+        private GrpcCall<TRequest, TResponse>? _activeCall;
+
+        public RetryCall(RetryPolicy retryPolicy, GrpcChannel channel, Method<TRequest, TResponse> method, CallOptions options)
+            : base(channel, method, options, LoggerName, retryPolicy.MaxAttempts.GetValueOrDefault())
+        {
+            _retryPolicy = retryPolicy;
+
+            _random = new Random();
```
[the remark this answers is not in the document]
JamesNK: I knew Random wasn't thread safe, but I didn't know there was also issues when creating many instances. Thanks for the heads up.

I'll create one per channel, and access via a lock.

src/Grpc.Net.Client/Internal/GrpcCall.NonGeneric.cs
```diff
@@ -90,5 +91,84 @@ protected bool TryGetTrailers([NotNullWhen(true)] out Metadata? trailers)
...
+                case HttpStatusCode.BadGateway:  // 502
+                case HttpStatusCode.ServiceUnavailable:  // 503
+                case HttpStatusCode.GatewayTimeout:  // 504
+                    return StatusCode.Unavailable;
+                default:
+                    if ((int)httpStatusCode >= 100 && (int)httpStatusCode < 200)
+                    {
+                        // 1xx. These headers should have been ignored.
+                        return StatusCode.Internal;
+                    }
```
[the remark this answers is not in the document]
JamesNK: This is just follow the standard mapping of HTTP to gRPC status codes. 1xx should never be seen but I'll leave this as is.

JamesNK:
@halter73 @JunTaoLuo @jkotalik @wtgodbe @davidfowl

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
JamesNK: Hmm, I think I missed that piece of logic.

src/Grpc.Net.Client/Configuration/HedgingPolicy.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        /// <summary>
+        /// The first call will be sent immediately, but the subsequent
+        /// hedged call will be sent at intervals of every delay. Set this
+        /// to 0 to immediately send all hedged calls.
+        /// </summary>
+        public TimeSpan? HedgingDelay
+        {
+            get => ConvertHelpers.ConvertDurationText(GetValue<string>(HedgingDelayPropertyName));
```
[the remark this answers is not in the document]
JamesNK: I've moved validating retry/hedging logic up before it is used with a call.

I want to keep the overall ServiceConfig validation lazy-ish because it is quite possible that some of it isn't used, i.e. config for methods that aren't called.

src/Grpc.Net.Client/Configuration/ConfigObject.cs
```diff
@@ -0,0 +1,68 @@
...
+            Inner = inner;
+        }
+
+        object IConfigValue.Inner => Inner;
+
+        internal T? GetValue<T>(string key)
+        {
+            if (Inner.TryGetValue(key, out var value))
+            {
+                return (T?)value;
```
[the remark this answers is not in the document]
JamesNK: gRPC client will likely be the only one setting values in these collections. Either via setters, or by loading JSON into untyped dictionaries/lists, and then being read by setters.

If someone sets a wrong value then I don't care that much that the exception message isn't super descriptive.

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
JamesNK: Done.

Right now a TCS combined with WhenAny is being used to escape the `Task.Delay` here. I'm pretty sure there are performance reasons why not canceling a Task.Delay you no longer care about is bad.

Double checking: I should use a CTS instead, right?

JamesNK: Updated so Task.Delay is canceled if not completed.

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
```diff
@@ -0,0 +1,404 @@
...
+
+        public HedgingCall(HedgingPolicyInfo hedgingPolicy, GrpcChannel channel, Method<TRequest, TResponse> method, CallOptions options)
+            : base(channel, method, options, LoggerName, hedgingPolicy.MaxAttempts)
+        {
+            _hedgingPolicy = hedgingPolicy;
+            _activeCalls = new List<IGrpcCall<TRequest, TResponse>>();
+
+            if (_hedgingPolicy.HedgingDelay > TimeSpan.Zero)
+            {
+                _delayInterruptTcs = new TaskCompletionSource<object?>(TaskCreationOptions.None);
```
Tratcher, line 58: Is this intentionally None and not RunContinuationsAsynchronously?
```suggestion
                _delayInterruptTcs = new TaskCompletionSource<object?>(TaskCreationOptions.RunContinuationsAsynchronously);
```

You're completing the Tcs in a lock and can then run the callback inline.

JamesNK: This was intentional at one point but it can be continued async now.

src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
```diff
@@ -0,0 +1,504 @@
...
+        }
+
+        protected abstract void OnCommitCall(IGrpcCall<TRequest, TResponse> call);
+
+        protected bool HasClientStream()
+        {
+            return Method.Type == MethodType.ClientStreaming || Method.Type == MethodType.DuplexStreaming;
+        }
+
+        protected void SetNewActiveCallUnsynchronized(IGrpcCall<TRequest, TResponse> call)
```
JunTaoLuo, line 358: Might want to mention in a comment that this call can only be made when holding a lock.

src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
```diff
@@ -0,0 +1,504 @@
...
+            {
+                _ctsRegistration?.Dispose();
+                CancellationTokenSource.Cancel();
+
+                if (CommitedCallTask.IsCompletedSuccessfully())
+                {
+                    CommitedCallTask.Result.Dispose();
+                }
+
+                ClearRetryBuffer();
```
JunTaoLuo, line 416: Yikes this means there's locking in a dispose call. I recall this being risky especially if there is a deadlock. Is there any way to avoid locking here?

src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
```diff
@@ -0,0 +1,504 @@
...
+    internal abstract partial class RetryCallBase<TRequest, TResponse> : IGrpcCall<TRequest, TResponse>
+        where TRequest : class
+        where TResponse : class
+    {
+        private readonly TaskCompletionSource<IGrpcCall<TRequest, TResponse>> _commitedCallTcs;
+        private RetryCallBaseClientStreamReader<TRequest, TResponse>? _retryBaseClientStreamReader;
+        private RetryCallBaseClientStreamWriter<TRequest, TResponse>? _retryBaseClientStreamWriter;
+        private CancellationTokenRegistration? _ctsRegistration;
+
+        protected object Lock { get; } = new object();
```
JunTaoLuo, line 47: `Lock` is a bit too generic and doesn't convey what it's used to synchronize. Looking at the usages, maybe `BufferLock`, `WriteLock` or `OutputLock`?

src/Grpc.Net.Client/Internal/Retry/RetryCall.cs
```diff
@@ -0,0 +1,320 @@
...
+        protected override void OnCommitCall(IGrpcCall<TRequest, TResponse> call)
+        {
+            _activeCall = null;
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            lock (Lock)
+            {
+                base.Dispose(disposing);
```
JunTaoLuo, line 240: Hmm? Doesn't `base.Dispose` also lock when disposing is true?

src/Grpc.Net.Client/Internal/Retry/RetryCall.cs
```diff
@@ -0,0 +1,320 @@
...
+            {
+                CompatibilityExtensions.Assert(call.ClientStreamWriter != null);
+
+                if (ClientStreamWriteOptions != null)
+                {
+                    call.ClientStreamWriter.WriteOptions = ClientStreamWriteOptions;
+                }
+
+                await call.WriteClientStreamAsync(WriteNewMessage, message).ConfigureAwait(false);
+                BufferedCurrentMessage = false;
```
JunTaoLuo, line 273: Hmm this value is read while holding the lock, should the lock be held when writing it?

src/Grpc.Net.Client/Internal/Retry/RetryCall.cs
```diff
@@ -0,0 +1,320 @@
...
+                    }
+
+                    Status? responseStatus;
+
+                    HttpResponseMessage? httpResponse = null;
+                    try
+                    {
+                        currentCall.CancellationToken.ThrowIfCancellationRequested();
+
+                        CompatibilityExtensions.Assert(currentCall._httpResponseTask != null, "Request should have be made if call is not preemptively cancelled.");
```
JunTaoLuo, line 132:
```suggestion
                        CompatibilityExtensions.Assert(currentCall._httpResponseTask != null, "Request should have been made if call is not preemptively cancelled.");
```

src/Grpc.Net.Client/Internal/Retry/RetryCall.cs
```diff
@@ -0,0 +1,320 @@
...
+                {
+                    await call.ClientStreamWriter.CompleteAsync().ConfigureAwait(false);
+                }
+            });
+        }
+
+        private async Task DoClientStreamActionAsync(Func<IGrpcCall<TRequest, TResponse>, Task> action)
+        {
+            var call = await GetActiveCallAsync(previousCall: null).ConfigureAwait(false);
+            while (true)
```
JunTaoLuo, line 285: Hmm, I feel like I'm missing some small nuance, why is this looped?

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
```diff
@@ -0,0 +1,409 @@
...
+                    }
+                }
+            }
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            lock (Lock)
+            {
+                base.Dispose(disposing);
```
JunTaoLuo, line 346: Same locking question here.

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
```diff
@@ -0,0 +1,409 @@
...
+            }
+
+            Status? responseStatus;
+
+            HttpResponseMessage? httpResponse = null;
+            try
+            {
+                call.CancellationToken.ThrowIfCancellationRequested();
+
+                CompatibilityExtensions.Assert(call._httpResponseTask != null, "Request should have be made if call is not preemptively cancelled.");
```
JunTaoLuo, line 91:
```suggestion
                CompatibilityExtensions.Assert(call._httpResponseTask != null, "Request should have been made if call is not preemptively cancelled.");
```

JunTaoLuo (commented):
Some minor questions

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
```diff
@@ -0,0 +1,409 @@
...
+                // Headers were returned. We're commited.
+                CommitCall(call, CommitReason.ResponseHeadersReceived);
+
+                // Wait until the call has finished and then check its status code
+                // to update retry throttling tokens.
+                var status = await call.CallTask.ConfigureAwait(false);
+                if (status.StatusCode == StatusCode.OK)
+                {
+                    // Success. Exit retry loop.
+                    Channel.RetryThrottling?.CallSuccess();
```
jkotalik, line 116: Lot's of nullable checks for RetryThrottling, anyway we can make it non-nullable?

src/Grpc.Net.Client/Configuration/MethodConfig.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// Initializes a new instance of the <see cref="MethodConfig"/> class.
+        /// </summary>
+        public MethodConfig() { }
+        internal MethodConfig(IDictionary<string, object> inner) : base(inner) { }
+
+        /// <summary>
+        /// Gets or sets the retry policy for outgoing calls.
+        /// A retry policy can't be combined with <see cref="HedgingPolicy"/>.
+        /// </summary>
+        public RetryPolicy? RetryPolicy
```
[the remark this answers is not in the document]
JamesNK: That's possible, but I want MethodConfig to match the method confg sturucture in proto/JSON. They place these fields in a oneof, but they're still two separate fields.

src/Grpc.Net.Client/Configuration/RetryPolicy.cs
```diff
@@ -0,0 +1,105 @@
...
+            set => SetValue(MaxBackoffPropertyName, ConvertHelpers.ToDurationText(value));
+        }
+
+        /// <summary>
+        /// Gets or sets the backoff multiplier.
+        /// The backoff will be multiplied by <see cref="BackoffMultiplier"/> after each retry
+        /// attempt and will increase exponentially when the multiplier is greater than 1.
+        /// This property is required and must be greater than 0.
+        /// </summary>
+        public double? BackoffMultiplier
```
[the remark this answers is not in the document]
JamesNK: A backoff multiplier less than 1 allows the multiplier to decrease after each retry.

This validation range matches the spec.

src/Grpc.Net.Client/Internal/GrpcProtocolHelpers.cs
```diff
@@ -402,12 +402,23 @@ public static StatusCode ResolveRpcExceptionStatusCode(Exception ex)
                 }
                 else if (current is IOException)
                 {
+                    // TODO(JamesNK): IOException is also returned for aborted requests.
```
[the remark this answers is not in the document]
JamesNK: https://github.com/grpc/grpc-dotnet/issues/1214

src/Grpc.Net.Client/GrpcChannelOptions.cs
```diff
@@ -65,6 +66,56 @@ public sealed class GrpcChannelOptions
...
+        /// using <see cref="ServiceConfig"/>.
+        /// </para>
+        /// <para>
+        /// A <c>null</c> value removes the maximum retry attempts limit. Defaults to 5.
+        /// </para>
+        /// <para>
+        /// Note: Experimental API that can change or be removed without any prior notice.
+        /// </para>
+        /// </summary>
+        public int? MaxRetryAttempts { get; set; }
```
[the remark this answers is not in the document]
JamesNK: I agree somewhat. I choose this name because it matches what Java has on its channel.

Also retries and hedging are both retry strategies. One strategy happens serially (retries) and the other in parallel (hedging).

I prefer `MaxRetryAttempts` over `MaxAttempts` or `MaxRetryOrHedgingAttempts`.

src/Grpc.Net.Client/Configuration/ServiceConfig.cs
```diff
@@ -0,0 +1,66 @@
...
+using Grpc.Net.Client.Internal.Configuration;
+
+namespace Grpc.Net.Client.Configuration
+{
+    /// <summary>
+    /// A <see cref="ServiceConfig"/> represents information about a service.
+    /// </summary>
+    public sealed class ServiceConfig : ConfigObject
+    {
+        private const string MethodConfigPropertyName = "method_config";
```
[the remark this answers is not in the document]
JamesNK: Bug. Good spotting.

src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -330,5 +380,57 @@ public void Dispose()
...
+                Service = service;
+                Method = method;
+            }
+
+            public string? Service { get; }
+            public string? Method { get; }
+
+            public override bool Equals(object? obj) => obj is MethodKey n ? Equals(n) : false;
+
+            public bool Equals(MethodKey other) => other.Service == Service && other.Method == Method;
```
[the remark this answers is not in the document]
JamesNK: `string.Equals(s1, s2, StringComparsion.Ordinal)` is slightly slower than `s1 == s2`.

Instead I'll add a comment that case is important.

src/Grpc.Net.Client/Internal/GrpcMethodInfo.cs
```diff
@@ -26,13 +29,110 @@ namespace Grpc.Net.Client.Internal
...
+                    InitialBackoff = r.InitialBackoff.GetValueOrDefault(),
+                    MaxBackoff = r.MaxBackoff.GetValueOrDefault(),
+                    BackoffMultiplier = r.BackoffMultiplier.GetValueOrDefault(),
+                    RetryableStatusCodes = r.RetryableStatusCodes.ToList()
+                };
+            }
+
+            if (methodConfig.HedgingPolicy != null)
+            {
+                m.HedgingPolicy = CreateHedgingPolicy(methodConfig.HedgingPolicy);
```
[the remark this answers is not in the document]
JamesNK: I used it in the unit tests. I'll make them consistent.

src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
JamesNK: The Unsynchronized suffix does that.

src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
JamesNK: Correct me if I'm wrong but locking in dispose doesn't seem that terrible. There is no finalizer so if a deadline did happen then it won't happen on the finalizer thread.

src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
JamesNK: There is one lock for everything locking related. I didn't see the need to have more and introduce the possibility of deadlocks.

src/Grpc.Net.Client/Internal/Retry/RetryCall.cs
JamesNK: Yes. It is the same lock so it doesn't matter.

src/Grpc.Net.Client/Internal/Retry/RetryCall.cs
JamesNK: Added comment

src/Grpc.Net.Client/Configuration/MethodConfig.cs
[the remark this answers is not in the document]
JunTaoLuo: Ah I see.

src/Grpc.Net.Client/GrpcChannel.cs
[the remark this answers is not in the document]
JunTaoLuo: Oh I didn't know that, I thought they were equivalent performance-wise.

src/Grpc.Net.Client/Configuration/RetryThrottlingPolicy.cs
```diff
@@ -0,0 +1,60 @@
...
+// limitations under the License.
+
+#endregion
+
+using System.Collections.Generic;
+
+namespace Grpc.Net.Client.Configuration
+{
+    /// <summary>
+    /// The retry throttling policy for a server.
```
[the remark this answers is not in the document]
JamesNK: Linked to spec.

src/Grpc.Net.Client/GrpcChannel.cs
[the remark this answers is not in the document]
JamesNK: The extra overhead is very very small, i.e. testing the StringComparison value

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
JamesNK: Still nullable, but I've centralized calls to retry throttling and the number of null checks.

src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
```diff
@@ -0,0 +1,504 @@
...
+            }
+
+            throw new InvalidOperationException("Can't get the call trailers because the call has not completed successfully.");
+        }
+
+        public void Dispose() => Dispose(true);
+
+        public void StartUnary(TRequest request)
+        {
+            StartCore(call => call.StartUnaryCore(CreatePushUnaryContent(request, call)));
```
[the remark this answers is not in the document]
JamesNK: I've spent a ton of time reducing duplication.

Right now I think it is good balance between reduced duplication and understandability.

src/Grpc.Net.Client/Configuration/RetryPolicy.cs
```diff
@@ -0,0 +1,102 @@
...
+
+using System;
+using System.Collections.Generic;
+using Grpc.Core;
+using Grpc.Net.Client.Internal.Configuration;
+
+namespace Grpc.Net.Client.Configuration
+{
+    /// <summary>
+    /// The retry policy for outgoing calls.
```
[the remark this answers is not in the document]
JamesNK: I've added details about allowed ranges of values to the XML doc comments.

JunTaoLuo (approved):
Changes look good though I wouldn't be surprised if there are some synchronization corner cases that we missed. Given that the features added are marked with experimental status, I'm okay with merging the changes and fix issues that come up.

Let's wait for @jtattermusch to take a look as well.

src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
```diff
@@ -0,0 +1,518 @@
...
+            Channel = channel;
+            Method = method;
+            Options = options;
+            _commitedCallTcs = new TaskCompletionSource<IGrpcCall<TRequest, TResponse>>(TaskCreationOptions.RunContinuationsAsynchronously);
+            BufferedMessages = new List<ReadOnlyMemory<byte>>();
+
+            // If the passed in token is canceled then we want to cancel the retry cancellation token.
+            if (options.CancellationToken.CanBeCanceled)
+            {
+                _ctsRegistration = options.CancellationToken.Register(state => ((RetryCallBase<TRequest, TResponse>)state!).CancellationTokenSource.Cancel(), this);
```
halter73, line 80: This looks like this will nullref if `options.CancellationToken` is already cancelled causing the Register callback runs inline.

It's probably best to just construct the CTS and register in one go when `options.CancellationToken.CanBeCanceled` with `CreateLinkedTokenSource`. Unless you want to avoid disposing the CTS which I understand if that's the case.

src/Grpc.Net.Client/GrpcChannelOptions.cs
```diff
@@ -65,6 +66,56 @@ public sealed class GrpcChannelOptions
...
+
+        /// <summary>
+        /// Gets or sets the maximum buffer size in bytes that can be used to store sent messages when retrying
+        /// or hedging calls. If the buffer limit is exceeded then no more retry attempts are made and all
+        /// hedging calls but one will be canceled. This limit is applied across all calls made using the channel.
+        /// <para>
```
halter73, line 89: > all hedging calls but one will be canceled.

Why not allow as many hedging calls as can fit under the limit?

src/Grpc.Net.Client/Internal/ClientStreamWriterBase.cs
```diff
@@ -0,0 +1,104 @@
...
+using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using Grpc.Core;
+using Microsoft.Extensions.Logging;
+
+#if NETSTANDARD2_0
+using ValueTask = System.Threading.Tasks.Task;
+#endif
```
halter73, line 28: Nit: This doesn't appear to be used.

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
```diff
@@ -0,0 +1,421 @@
...
+                {
+                    // Cancel the Task.Delay that's no longer needed.
+                    // https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/519ef7d231c01116f02bc04354816a735f2a36b6/AsyncGuidance.md#using-a-timeout
+                    _hedgingDelayCts.Cancel();
+                }
+
+                lock (Lock)
+                {
+                    // Need to recreate the delay TCS/CTS as needed for next cycle.
+                    if (_delayInterruptTcs.Task.IsCompleted)
```
halter73, line 312: Nit: This is always true here, right?

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
```diff
@@ -0,0 +1,421 @@
...
+                    {
+                        _delayInterruptTcs = new TaskCompletionSource<object?>(TaskCreationOptions.RunContinuationsAsynchronously);
+                    }
+                    if (_hedgingDelayCts.IsCancellationRequested)
```
halter73, line 316: And this looks like it's always canceled here.

src/Grpc.Net.Client/GrpcChannelOptions.cs
JamesNK: If you're making a unary call (one message that exceeds limit) then other hedging calls haven't started.

If you're making a client streaming call (a new message has exceeded the limit) then the other hedging calls won't have all the client messages. They need to be canceled.

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
JamesNK: No. If there is no interrupt - the hedging delay ran to completition - then this TCS is still pending. In that case it is reused.

halter73: Just above in this method we have:

```
                if (completedTask != _delayInterruptTcs.Task)
                {
                    // Task.Delay won. Check CTS to see if it won because of cancellation.
                    _hedgingDelayCts.Token.ThrowIfCancellationRequested();
                    return;
                }
```

Therefore `_delayInterruptTcs.Task` *must* be completed in order to continue to here, no? I don't see a new `_delayInterruptTcs` that's constructed anywhere but here and in the ctor.

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
JamesNK: If there is no interrupt and the overall call wasn't cancelled then this CTS won't be canceled.

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
JamesNK: Ah right. I think this might have changed because of a review comment. I'll investigate.

jtattermusch (approved):
Left a few comments, but overall LGTM. Thanks for implementing this!

Directory.Build.props
```diff
@@ -18,7 +18,7 @@
     <WarningsNotAsErrors>$(WarningsNotAsErrors);CS1591</WarningsNotAsErrors>
 
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
-    <LangVersion>8.0</LangVersion>
+    <LangVersion>9.0</LangVersion>
```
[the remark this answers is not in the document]
JamesNK: Some delegates are marked as static which is new in C# 9.

Making them static doesn't change how they behave, it just validates that there is no chance of accidently allocating a closure.

src/Grpc.Net.Client/Internal/Retry/CommitReason.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        ResponseHeadersReceived,
+        FatalStatusCode,
+        ExceededAttemptCount,
+        DeadlineExceeded,
+        Throttled,
+        BufferExceeded,
+        PushbackStop,
+        UnexpectedError,
+        Canceled
```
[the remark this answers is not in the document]
JamesNK: No it's canceled with one l. e.g. [TaskCanceledException](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcanceledexception?view=net-5.0)

src/Grpc.Net.Client/Internal/Configuration/ConvertHelpers.cs
```diff
@@ -0,0 +1,106 @@
...
+        public static TimeSpan? ConvertDurationText(string? text)
+        {
+            if (text == null)
+            {
+                return null;
+            }
+
+            if (text.Length > 0 && text[text.Length - 1] == 's')
+            {
+                return TimeSpan.FromSeconds(Convert.ToDouble(text.Substring(0, text.Length - 1), CultureInfo.InvariantCulture));
```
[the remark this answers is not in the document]
JamesNK: I've change this to use decimal and then multiple the result to get ticks.

Since timers are only precise down to milliseconds, having the *exact* value shouldn't matter.

src/Grpc.Net.Client/Configuration/MethodName.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// and the method config is the default for all methods on all services.
+    /// <see cref="MethodName.Default"/> represents this global default name.
+    /// </para>
+    /// <para>
+    /// When determining which method config to use for a given RPC, the most specific match wins. A method config
+    /// with a configured <see cref="MethodName"/> that exactly matches a call's method and service will be used
+    /// instead of a service or global default method config.
+    /// </para>
+    /// </remarks>
+    public sealed class MethodName
```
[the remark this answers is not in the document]
JamesNK: Done

src/Grpc.Net.Client/Configuration/ServiceConfig.cs
```diff
@@ -0,0 +1,69 @@
...
+
+#endregion
+
+using System.Collections.Generic;
+using Grpc.Net.Client.Internal.Configuration;
+
+namespace Grpc.Net.Client.Configuration
+{
+    /// <summary>
+    /// A <see cref="ServiceConfig"/> represents information about a service.
```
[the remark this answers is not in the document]
JamesNK: Done

JamesNK:
🍰 🥳 🎈 🎉

src/Grpc.Net.Client/Internal/Retry/CommitReason.cs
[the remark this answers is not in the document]
jtattermusch: yeah but there's e.g. .NET `CancellationToken`. I guess not even .NET itself is being consistent here (or there's some english grammar nuance that I'm missing). 
Fair to leave as is.

PR #1203, opened 2021-02-17
gRPC-Web - Add .NET Standard 2.0 support

JamesNK:
Adds `netstandard2.0` target to `Grpc.Net.Client.Web`.


src/Grpc.Net.Client.Web/Internal/GrpcWebRequestContent.cs
```diff
@@ -61,6 +61,16 @@ private async Task SerializeTextToStreamAsync(Stream stream)
 
         protected override bool TryComputeLength(out long length)
         {
+            // ContentLength is calculated using the inner content's TryComputeLength.
+            var contentLength = _inner.Headers.ContentLength;
+            if (contentLength != null)
+            {
+                length = _mode == GrpcWebMode.GrpcWebText
```
JunTaoLuo, line 68: Curious, why did the return value change and where did this computation originate?

JunTaoLuo (approved):
Looks good, but I'm curious about one change in GrpcWebRequestContent

PR #1232, opened 2021-03-10
Fix client hang when call has missing or invalid trailers

JamesNK:


JunTaoLuo (approved):

JamesNK:
Build failures look like a problem with Google infrastructure. Unrelated.

PR #1233, opened 2021-03-11
v2.36.x - Fix client hang when call has missing or invalid trailers

JamesNK:
Cherry picked changes from #1232 onto v2.36.x release branch.

I found this issue when testing. I don't think another preview is required but I want to include this fix with 2.36.

@jtattermusch


jtattermusch (approved):
LGTM

PR #1253, opened 2021-03-31
Move static Log classes out of generic types

JamesNK:
Fields on static `Log` classes inside a generic type are duplicated with each type parameter.

Move static logging outside of generic types.


JunTaoLuo (approved):
Oh cool. How much does this save?

JamesNK:
I haven't measured. Depends on how many services and message types you have in an app.

Issue #1264, opened 2021-04-09
Cache and reuse channels in gRPC client factory


JamesNK:
gRPC client factory currently always creates a new channel. This has been fine because creating them is cheap and they haven't had any state.

gRPC retries adds state to the channel (throttling counters). Load balancing will add much more state.

PR #1265, opened 2021-04-09
GrpcChannel created by factory has same lifetime as handlers

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1264

* GrpcChannel now has same lifetime as underlying HttpHandlers
* GrpcChannel is disposed with HttpHandlers
* Default lifetime changed to 30 minutes


JamesNK:
Too many problems with this approach.

PR #1350, opened 2021-07-25
Update HedgingPolicy.cs

VahidFarahmandian:
Setting null to HedgingDelay will cause InvalidOperationException and based on the internal code(https://github.com/grpc/grpc-dotnet/blob/bf26e0377abfa86f34778ee1347565f02d90ffb5/src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs#L31), this property should not be a nullable property


JamesNK:
HedgingDelay should support being "unset" (i.e. `null`) so the bug is the validation of the value. I've fixed that here: https://github.com/grpc/grpc-dotnet/pull/1351

PR #1359, opened 2021-08-01
Update SDK and targets to .NET 6

JamesNK:
Support .NET 6 and add new targets on client and server:
* Remove custom telemetry handler from client
* Use NonValidated headers in client to fix thread-safety issues
* Don't error if client upgrades protocol to HTTP/3


JunTaoLuo (approved):
Looks good

PR #1373, opened 2021-08-12
Use dotnet-6 containers

profporridge:
Looks like this example might have been missed when updating to dotnet 6 - at least I needed to make this change to build the example.


JamesNK:
Hi, thanks for calling this out. I've created a PR that updates this a long with some other changes to clean up that sample. It replaces this PR.

https://github.com/grpc/grpc-dotnet/pull/1379

Issue #1374, opened 2021-08-13
gRPC-Web on .NET Framework should set Version to 1.1


JamesNK:
An error is thrown unless `GrpcWebHandler.HttpVersion` is set to `HttpVersion.Version11`:

> Exception when processing request: Grpc.Core.RpcException: Status(StatusCode="Internal", Detail="Error starting gRPC call. ArgumentException: Only HTTP/1.0 and HTTP/1.1 version requests are currently supported. Parameter name: value", DebugException="System.ArgumentException: Only HTTP/1.0 and HTTP/1.1 version requests are currently supported. Parameter name: value    at System.Net.HttpWebRequest.set_ProtocolVersion(Version value)    at System.Net.Http.HttpClientHandler.CreateAndPrepareWebRequest(HttpRequestMessage request)    at System.Net.Http.HttpClientHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken) --- End of stack trace from previous location where exception was thrown ---    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)    at Grpc.Net.Client.Web.GrpcWebHandler.<SendAsyncCore>d__18.MoveNext() --- End of stack trace from previous location where exception was thrown ---    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)    at Grpc.Net.Client.Internal.GrpcCall`2.<RunCall>d__70.MoveNext()")    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()

Either add a new target for net45 or check for runtime in code.

PR #1377, opened 2021-08-15
Default HTTP version to 1.1 when using gRPC-Web on netfx

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1374


jtattermusch (approved):
LGTM

PR #1379, opened 2021-08-16
Update containers sample to use .NET 6 SDK image

JamesNK:


profporridge (commented):
Is there a reason that you didn't update the aspnet image on line 11 as well? https://github.com/grpc/grpc-dotnet/blob/02a9259cea77994822de31414e5c01beceec03d0/examples/Container/Frontend/Dockerfile#L11

Same for the other Dockerfile https://github.com/grpc/grpc-dotnet/blob/02a9259cea77994822de31414e5c01beceec03d0/examples/Container/Backend/Dockerfile#L11

In my [now closed pull request](#1373) that inspired this PR I updated both references but I'm wondering if that was a mistake on my part - perhaps we want to build with the 6.0 SDK but run on the aspnet 5 image?

JamesNK:
The global.json requires .NET 6, so the SDK container must be 6.0. However, the app itself is still .NET 5 and only needs runtime of 5.0.

When .NET 6 is RTM I will update all the samples to .NET 6 and change the runtime to 6.0.

profporridge:
> The global.json requires .NET 6, so the SDK container must be 6.0. However, the app itself is still .NET 5 and only needs runtime of 5.0.
> 
> When .NET 6 is RTM I will update all the samples to .NET 6 and change the runtime to 6.0.

That makes sense, thanks. 

I can confirm that the example fails to start if you change line 11 to aspnet:6.0
```
$ docker-compose up 
Recreating container_container-frontend_1 ... done
Starting container_container-backend_1    ... done
Attaching to container_container-backend_1, container_container-frontend_1
container-backend_1   | It was not possible to find any compatible framework version
container-backend_1   | The framework 'Microsoft.AspNetCore.App', version '5.0.0' was not found.
container-backend_1   |   - The following frameworks were found:
container-backend_1   |       6.0.0-preview.7.21378.6 at [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
container-backend_1   | 
container-backend_1   | You can resolve the problem by installing the specified framework and/or SDK.
container-backend_1   | 
container-backend_1   | The specified framework can be found at:
container-backend_1   |   - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.AspNetCore.App&framework_version=5.0.0&arch=x64&rid=debian.11-x64
container_container-backend_1 exited with code 150
```

profporridge (approved):

jtattermusch (approved):
LGTM.

PR #1387, opened 2021-08-20
Fix error from tests HttpEventSourceListener

JamesNK:
Error in HttpEventSourceListener is consistently failing builds. I'm not sure why the original PR that added it didn't catch it 🤷


jtattermusch (approved):
LGTM.

Issue #1384, opened 2021-08-18
Calling Grpc service method should throw if AsyncAuthInterceptor throws


MarS0K:
### What version of gRPC and what language are you using?

Grpc.Net.Client 2.38.0.

### What operating system (Linux, Windows,...) and version?

 OS Name:     Windows
 OS Version:  10.0.19042
 OS Platform: Windows
 RID:         win10-x64

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)

TargetFramework net5.0
.NET Core SDK version5.0.400

### What did you do?

I use AsyncAuthInterceptor for CallCredentials when call Grpc service method.
But method call never completes when AsyncAuthInterceptor throws.
Is it expected behavior?

Sample code:
```
        [Test]
        public void CallCredentialsWithHttps_WhenAsyncAuthInterceptorThrow_ShouldThrow()
        {
            // Arrange
            var httpClient = ClientTestHelpers.CreateTestClient(async request =>
            {
                var reply = new HelloReply { Message = "Hello world" };
                var streamContent = await ClientTestHelpers.CreateResponseContent(reply).DefaultTimeout();
                return ResponseUtils.CreateResponse(HttpStatusCode.OK, streamContent);
            });
            var invoker = HttpClientCallInvokerFactory.Create(httpClient);

            // Act
            var expectedException = new Exception("Some AsyncAuthInterceptor Exception");

            var callCredentials = CallCredentials.FromInterceptor((context, metadata) =>
            {
                return Task.FromException(expectedException);
            });

            var actualException = Assert.ThrowsAsync<Exception>(async () =>
            {
                var call = invoker.AsyncUnaryCall<HelloRequest, HelloReply>(ClientTestHelpers.ServiceMethod, string.Empty, new CallOptions(credentials: callCredentials), new HelloRequest());
                await call.ResponseAsync.DefaultTimeout();
            });

            Assert.AreSame(expectedException, actualException);
        }
```

### What did you expect to see?

If AsyncAuthInterceptor throws I expect to see that call throws exception.

### What did you see instead?

If AsyncAuthInterceptor throws the call never completes (throws TimeoutException in sample code).

JamesNK:
Thanks for the good reproduction. I have a fix for this that will be in the next release.

PR #1388, opened 2021-08-20
Fix client timeout when error in custom CallCredential

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1384

Also fixes a flakey test that failed during CI.


JamesNK:
Latest failure is fixed by https://github.com/grpc/grpc-dotnet/pull/1387

JamesNK:
I think I figured out https://github.com/grpc/grpc-dotnet/pull/1387 for real.

There is a race when the event source starts between an event getting written and the logger being set in the ctor.

jtattermusch (approved):

PR #1401, opened 2021-09-03
Fix streaming with no response messages not successful

JamesNK:
Fixes a bug I noticed while fixing https://github.com/grpc/grpc-dotnet/pull/1398

A call that doesn't return a message (e.g. a server streaming call that completes without streaming any messages) isn't recognized as successful.


Caglar1881 (changes requested):
😂

captainsafia (approved):

PR #1402, opened 2021-09-06
Client handle server cancellation for deadline

JamesNK:
Handle the situation where the server exceeds deadline and sends RST_STREAM with status CANCELLED to the client, and the client receives it before its local timer executes.

### DIFF
--- a/src/Grpc.Net.Client/Internal/GrpcCall.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCall.cs
@@ -650,7 +650,7 @@ namespace Grpc.Net.Client.Internal
                 // The server could exceed the deadline and return a CANCELLED status before the
                 // client's deadline timer is triggered. When CANCELLED is received check the
                 // deadline against the clock and change status to DEADLINE_EXCEEDED if required.
-                var deadlineExceeded = s.StatusCode == StatusCode.Cancelled && _deadline != DateTime.MaxValue && _deadline - Channel.Clock.UtcNow <= TimeSpan.Zero;
+                var deadlineExceeded = s.StatusCode == StatusCode.Cancelled && IsDeadlineExceeded();
                 if (deadlineExceeded)
                 {
                     s = new Status(StatusCode.DeadlineExceeded, s.Detail, s.DebugException);
@@ -769,7 +769,7 @@ namespace Grpc.Net.Client.Internal
                     // the client has processed that it has exceeded or not.
                     lock (this)
                     {
-                        if (_deadline <= Channel.Clock.UtcNow)
+                        if (IsDeadlineExceeded())
                         {
                             GrpcCallLog.DeadlineExceeded(Logger);
                             GrpcEventSource.Log.CallDeadlineExceeded();
@@ -810,6 +810,11 @@ namespace Grpc.Net.Client.Internal
             }
 
             return true;
+        }
+
+        private bool IsDeadlineExceeded()
+        {
+            return _deadline <= Channel.Clock.UtcNow;
         }
 
         private async Task ReadCredentials(HttpRequestMessage request)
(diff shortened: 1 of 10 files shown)


src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -622,8 +640,20 @@ internal bool ResolveException(string summary, Exception ex, [NotNull] out Statu
...
+                // deadline against the clock and change status to DEADLINE_EXCEEDED if required.
+                if (s.StatusCode == StatusCode.Cancelled &&
+                    _deadline != DateTime.MaxValue &&
+                    _deadline - Channel.Clock.UtcNow <= TimeSpan.Zero)
+                {
+                    s = new Status(StatusCode.DeadlineExceeded, s.Detail, s.DebugException);
+                }
+
+                status = s;
+                resolvedException = CreateRpcException(s);
```
captainsafia, line 656:
```suggestion
                resolvedException = CreateRpcException(status);
```

Nit: reads a little clearer although there's no functional difference.

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -622,8 +640,20 @@ internal bool ResolveException(string summary, Exception ex, [NotNull] out Statu
...
             else
             {
-                status = GrpcProtocolHelpers.CreateStatusFromException(summary, ex);
-                resolvedException = CreateRpcException(status.Value);
+                var s = GrpcProtocolHelpers.CreateStatusFromException(summary, ex);
+
+                // The server could exceed the deadline and return a CANCELLED status before the
+                // client's deadline timer is triggered. When CANCELLED is received check the
+                // deadline against the clock and change status to DEADLINE_EXCEEDED if required.
+                if (s.StatusCode == StatusCode.Cancelled &&
```
captainsafia, line 648:
```suggestion
                var didExceedDeadline = s.StatusCode == StatusCode.Cancelled && _deadline != DateTime.MaxValue && _deadline - Channel.Clock.UtcNow <= TimeSpan.Zero;
                if (didExceedDeadline)
```

src/Shared/CompatibilityHelpers.cs
```diff
@@ -40,5 +40,14 @@ public static bool IsCompletedSuccessfully<T>(this ValueTask<T> task)
             return task.IsCompletedSuccessfully;
         }
 #endif
+
+        public static int IndexOf(string s, char value, StringComparison comparisonType)
```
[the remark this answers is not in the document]
JamesNK: If a method is only used once then the `#ifdef` is left inline. If a method is used multiple times then it makes sense to centralize it here.

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -622,8 +640,21 @@ internal bool ResolveException(string summary, Exception ex, [NotNull] out Statu
...
-                resolvedException = CreateRpcException(status.Value);
+                var s = GrpcProtocolHelpers.CreateStatusFromException(summary, ex);
+
+                // The server could exceed the deadline and return a CANCELLED status before the
+                // client's deadline timer is triggered. When CANCELLED is received check the
+                // deadline against the clock and change status to DEADLINE_EXCEEDED if required.
+                var deadlineExceeded = s.StatusCode == StatusCode.Cancelled && _deadline != DateTime.MaxValue && _deadline - Channel.Clock.UtcNow <= TimeSpan.Zero;
+                if (deadlineExceeded)
+                    _deadline != DateTime.MaxValue &&
+                    _deadline - Channel.Clock.UtcNow <= TimeSpan.Zero)
```
JamesNK, line 651:
```suggestion
                if (deadlineExceeded)
```

src/Grpc.Net.Client/Internal/GrpcCall.cs
JamesNK: `s` is used here because `status` is nullable.

JamesNK:
@captainsafia Additional feedback or approval so this can be merged? 🙏

captainsafia (approved):

PR #1404, opened 2021-09-08
Bump version to v2.40.0-pre1 (on v2.40.x branch)

jtattermusch:
I'll publish the 2.40.0-pre1 release after merging this.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -3,7 +3,7 @@
     <BenchmarkDotNetPackageVersion>0.12.1</BenchmarkDotNetPackageVersion>
     <GoogleProtobufPackageVersion>3.15.8</GoogleProtobufPackageVersion>
     <GrpcDotNetPackageVersion>2.39.0-pre1</GrpcDotNetPackageVersion>  <!-- Used by example projects -->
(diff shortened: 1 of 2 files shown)


captainsafia (approved):
LGTM.

What's the reasoning for publishing a new release after merging vs. after?

jtattermusch:
> LGTM.
> 
> What's the reasoning for publishing a new release after merging vs. after?

The version of the nuget packages that will be created during the release (there's a CI job that creates the nugets) will be the version we are setting in this PR - so we first need to merge to update the version and then start with building the packages.

Also, from practical perspective, sometimes when folks see the PR after which the new release will be cut, they remember that we still need to squeeze in a fix before we release. So the PR to set version to "-preX" acts as a heads up "I'm about to cut a release" (and once a release tag is created, it should stay immutable, so we can't that it back).

FTR this is the release process docs: https://github.com/grpc/grpc-dotnet/blob/master/doc/release_process.md

jtattermusch:
I created the release https://github.com/grpc/grpc-dotnet/releases/tag/v2.40.0-pre1 and published the nuget packages.

PR #1411, opened 2021-09-14
Add more info to UserAgent header

### LINKED ISSUE #1392: Report info about .NET version and platform in user-agent
We've recently added info about .NET version to Grpc.Core: https://github.com/grpc/grpc/pull/25889, so the user-agent string in Grpc.Core looks sth like this: `grpc-csharp/2.38.0-dev (.NET 5.0.0; CLR 5.0.0; netstandard2.0; x64) grpc-c/16.0.0 (osx; chttp2)`

Currently the user-agent string of grpc-dotnet is very basic (e.g. `grpc-dotnet/2.36.0.0`) and there are situations in which .NET version info comes very handy (and platform like e.g. 'linux/windows/osx')

captainsafia:
Fixes #1392.

User agent string prior to this change: grpc-dotnet/2.41.0.0 
User agent string with this change: grpc-dotnet/2.4.1.0.0 (.NET 5.0.7; CLR 5.0.7; net461; osx x64)

### DIFF
--- a/src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
@@ -22,8 +22,6 @@ using System.Diagnostics;
 using System.Linq;
 using System.Net.Http.Headers;
 using System.Reflection;
-using System.Runtime.InteropServices;
-using System.Runtime.Versioning;
 using Grpc.Core;
 using Grpc.Net.Compression;
 
@@ -90,54 +88,11 @@ namespace Grpc.Net.Client.Internal
         static GrpcProtocolConstants()
         {
             UserAgentHeader = "User-Agent";
-            UserAgentHeaderValue = GetUserAgentString();
+            UserAgentHeaderValue = UserAgentGenerator.GetUserAgentString();
             TEHeader = "TE";
             TEHeaderValue = "trailers";
 
             DefaultMessageAcceptEncodingValue = GetMessageAcceptEncoding(DefaultCompressionProviders);
-        }
-
-        private static string GetUserAgentString()
-        {
-            var userAgent = "grpc-dotnet";
-
-            // Use the assembly file version in the user agent.
-            // We are not using `AssemblyInformationalVersionAttribute` because Source Link appends
-            // the git hash to it, and sending a long user agent has perf implications.
-            var assembly = typeof(GrpcProtocolConstants).Assembly;
-            var assemblyVersion = assembly
-                .GetCustomAttributes<AssemblyFileVersionAttribute>()
-                .FirstOrDefault();
-
-            Debug.Assert(assemblyVersion != null);
-
-            // Assembly file version attribute should always be present,
-            // but in case it isn't then don't include version in user-agent.
-            if (assemblyVersion != null)
-            {
-                // 1.24.0
-                userAgent += $"/{assemblyVersion.Version} ";
-            }
-
-            // (.NET 5.0.7;
-            userAgent += $"({RuntimeInformation.FrameworkDescription}; ";
-            // CLR 5.0.0;
-            userAgent += $"CLR {Environment.Version}; ";
-
-            // .NETCoreApp,Version=v6.0; 
-            var targetFramework = assembly
-                .GetCustomAttributes<TargetFrameworkAttribute>()
-                .FirstOrDefault()?
-                .FrameworkName;
-            if (!string.IsNullOrEmpty(targetFramework))
-            {
-                userAgent += $"{targetFramework}; ";
-            }
-
-            // x64)
-            userAgent += Environment.Is64BitProcess ? "x64)" : "x32)";
-
-            return userAgent;
         }
     }
 }
--- a/src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
@@ -22,6 +22,8 @@ using System.Diagnostics;
 using System.Linq;
 using System.Net.Http.Headers;
 using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
 using Grpc.Core;
 using Grpc.Net.Compression;
 
@@ -87,13 +89,23 @@ namespace Grpc.Net.Client.Internal
 
         static GrpcProtocolConstants()
         {
+            UserAgentHeader = "User-Agent";
+            UserAgentHeaderValue = GetUserAgentString();
+            TEHeader = "TE";
+            TEHeaderValue = "trailers";
+
+            DefaultMessageAcceptEncodingValue = GetMessageAcceptEncoding(DefaultCompressionProviders);
+        }
+
+        private static string GetUserAgentString()
+        {
             var userAgent = "grpc-dotnet";
 
             // Use the assembly file version in the user agent.
             // We are not using `AssemblyInformationalVersionAttribute` because Source Link appends
             // the git hash to it, and sending a long user agent has perf implications.
-            var assemblyVersion = typeof(GrpcProtocolConstants)
-                .Assembly
+            var assembly = typeof(GrpcProtocolConstants).Assembly;
+            var assemblyVersion = assembly
                 .GetCustomAttributes<AssemblyFileVersionAttribute>()
                 .FirstOrDefault();
 
@@ -103,15 +115,29 @@ namespace Grpc.Net.Client.Internal
             // but in case it isn't then don't include version in user-agent.
             if (assemblyVersion != null)
             {
-                userAgent += "/" + assemblyVersion.Version;
+                // 1.24.0
+                userAgent += $"/{assemblyVersion.Version} ";
             }
 
-            UserAgentHeader = "User-Agent";
-            UserAgentHeaderValue = userAgent;
-            TEHeader = "TE";
-            TEHeaderValue = "trailers";
+            // (.NET 5.0.7;
+            userAgent += $"({RuntimeInformation.FrameworkDescription}; ";
+            // CLR 5.0.0;
+            userAgent += $"CLR {Environment.Version}; ";
 
-            DefaultMessageAcceptEncodingValue = GetMessageAcceptEncoding(DefaultCompressionProviders);
+            // .NETCoreApp,Version=v6.0; 
+            var targetFramework = assembly
+                .GetCustomAttributes<TargetFrameworkAttribute>()
+                .FirstOrDefault()?
+                .FrameworkName;
+            if (!string.IsNullOrEmpty(targetFramework))
+            {
+                userAgent += $"{targetFramework}; ";
+            }
+
+            // x64)
+            userAgent += Environment.Is64BitProcess ? "x64)" : "x32)";
+
+            return userAgent;
         }
     }
 }
--- a/src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
+++ b/src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
@@ -26,6 +26,8 @@ namespace Grpc.Net.Client.Internal
 {
     internal static class UserAgentGenerator
     {
+        internal static bool IsMono { get; } = Type.GetType("Mono.Runtime") != null;
+
         internal static string GetUserAgentString()
         {
             var assembly = typeof(GrpcProtocolConstants).Assembly;
@@ -42,7 +44,7 @@ namespace Grpc.Net.Client.Internal
             
             return GetUserAgentString(
                 processArch: RuntimeInformation.ProcessArchitecture,
-                runtimeVersion: IsMono() ? null : Environment.Version, 
+                runtimeVersion: IsMono ? null : Environment.Version, 
                 assemblyVersion: assemblyVersion,
                 // RuntimeInformation.FrameworkDescription is only supported for
                 // .NET Framework 4.7.1 and above. If targetting net461 or earlier,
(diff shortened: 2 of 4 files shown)


src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
```diff
@@ -103,15 +115,29 @@ static GrpcProtocolConstants()
...
-            UserAgentHeaderValue = userAgent;
-            TEHeader = "TE";
-            TEHeaderValue = "trailers";
+            // (.NET 5.0.7;
+            userAgent += $"({RuntimeInformation.FrameworkDescription}; ";
+            // CLR 5.0.0;
+            userAgent += $"CLR {Environment.Version}; ";
+
+            // .NETCoreApp,Version=v6.0; 
+            var targetFramework = assembly
```
jtattermusch, line 128: Grpc.Core currently produces strings like "net45", "netstandard2.0" etc.  Can we stay consistent? it's not an absolute must but there's value in consistency.

src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
```diff
@@ -103,15 +115,29 @@ static GrpcProtocolConstants()
             // but in case it isn't then don't include version in user-agent.
             if (assemblyVersion != null)
             {
-                userAgent += "/" + assemblyVersion.Version;
+                // 1.24.0
+                userAgent += $"/{assemblyVersion.Version} ";
```
jtattermusch, line 119: nit: I think assemblyVersion is actually a 4-part thing, so comment is wrong?
2.38.0.0?

note that Grpc.Core has a 3-part version number in user-agent string (since it doesn't extract the version number from the assemblyVersion but uses a different mechanism), but we can keep the 4-part format in grpc-dotnet for consistency with its older releases.

qq: it would be nice if the version number contained info about whether the version is a pre-release or a stable release (e.g. `2.38.0` vs `2.38.0-pre1`) since the distinction is quite important. 
Ideally we'd want to use this value as the version string (it's also what gets used as the nuget package version)
https://github.com/grpc/grpc-dotnet/blob/094532f0fbd3c9c877614beca7a73a797f3eba02/build/version.props#L5

src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
```diff
@@ -103,15 +115,29 @@ static GrpcProtocolConstants()
...
             }
 
-            UserAgentHeader = "User-Agent";
-            UserAgentHeaderValue = userAgent;
-            TEHeader = "TE";
-            TEHeaderValue = "trailers";
+            // (.NET 5.0.7;
+            userAgent += $"({RuntimeInformation.FrameworkDescription}; ";
```
jtattermusch, line 123: nit: RuntimeInformation.FrameworkDescription is available starting from .NET Framework 4.7.1+, but we had plans to support the legacy net46+ (for grpc-web and grpc over winhttp2 handler). 
So if we added the net461 target, this won't build.

What you currently have is fine but perhaps worth adding a TODO that if net46x targets are added in the future, this will need more work (we can get inspired with what Grpc.Core does here. https://github.com/grpc/grpc/pull/25889/files#diff-e0b4c69507f528047f6f1d956277df9468e730ca55066821869e7b140d96b66dR207

src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
```diff
@@ -103,15 +115,29 @@ static GrpcProtocolConstants()
...
+                .FirstOrDefault()?
+                .FrameworkName;
+            if (!string.IsNullOrEmpty(targetFramework))
+            {
+                userAgent += $"{targetFramework}; ";
+            }
 
-            DefaultMessageAcceptEncodingValue = GetMessageAcceptEncoding(DefaultCompressionProviders);
+            // x64)
+            userAgent += Environment.Is64BitProcess ? "x64)" : "x32)";
```
jtattermusch, line 138: nit: Grpc.Core uses x86  for 32-bit (and x64 for AMD64).

Also, just supporting AMD architecture won't be enough.  we definitely need at least arm64 (since that's currently supported by .NET and ARM architecture is becoming increasingly important these days).

jtattermusch: I think you should use  RuntimeInformation.ProcessArchitecture  (which is readily available on platforms that grpc-dotnet currently supports) https://github.com/grpc/grpc/blob/7c10e1b9fe9b751cffc3e2d2c0f37feb3163fe93/src/csharp/Grpc.Core/Internal/CommonPlatformDetection.cs#L80

src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
```diff
@@ -103,15 +115,29 @@ static GrpcProtocolConstants()
...
+
+            return userAgent;
```
jtattermusch, line 140: we could really use info about the operating system / platform. 
Grpc.Core provides that info in the second part of the user agent string that corresponds to C-core (`grpc-csharp/2.38.0-dev (.NET 5.0.0; CLR 5.0.0; netstandard2.0; x64) grpc-c/16.0.0 (osx; chttp2)`) - but since grpc-dotnet doesn't have the C-core layer, we need to include platform info ourselves.  Use `RuntimeInformation.IsOSPlatform()`?

test/Grpc.Net.Client.Tests/AsyncUnaryCallTests.cs
```diff
@@ -76,14 +77,18 @@ public async Task AsyncUnaryCall_Success_HttpRequestMessagePopulated()
             Assert.AreEqual("identity,gzip", httpRequestMessage.Headers.GetValues(GrpcProtocolConstants.MessageAcceptEncodingHeader).Single());
             Assert.AreEqual(null, requestContentLength);
 
-            var userAgent = httpRequestMessage.Headers.UserAgent.Single()!;
-            Assert.AreEqual("grpc-dotnet", userAgent.Product?.Name);
-            Assert.IsTrue(!string.IsNullOrEmpty(userAgent.Product?.Version));
+            var grpcVersion = httpRequestMessage.Headers.UserAgent.First();
```
jtattermusch, line 80: we probably need some tests for the user-agent functionality (e.g. at least check the format looks the way we expect it to look with a regex).

src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
```diff
@@ -86,14 +88,24 @@ internal static string GetMessageAcceptEncoding(Dictionary<string, ICompressionP
...
+        {
+            UserAgentHeader = "User-Agent";
+            UserAgentHeaderValue = GetUserAgentString();
+            TEHeader = "TE";
+            TEHeaderValue = "trailers";
+
+            DefaultMessageAcceptEncodingValue = GetMessageAcceptEncoding(DefaultCompressionProviders);
+        }
+
+        private static string GetUserAgentString()
```
jtattermusch, line 100: nit: extract the functionality for computing the user agent string to a separate  static class (and then add tests for that class)?

jtattermusch (commented):
Thanks! I left some comments.

src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
captainsafia: Done! Let me know if you have any feedback on the formatting for this.

src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
captainsafia: Good call! I've changed this to read from `AssemblyInformationalVersionAttribute`. I saw a comment around concerns with the git hash being in that string so I've stripped out the git hash from the generated user agent.

test/Grpc.Net.Client.Tests/AsyncUnaryCallTests.cs
captainsafia: Followed the recommendation below and added a separate test class.

src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
```diff
@@ -0,0 +1,164 @@
...
+            // CLR 5.0.0;
+            userAgent += GetRuntimeVersion(runtimeVersion);
+            // net6.0;
+            userAgent += GetFrameworkName(frameworkName);
+            // windows  
+            userAgent += $"{operatingSystem} ";
+            // x64)
+            userAgent += $"{GetProcessArch(processArch)})";
+
+            string GetRuntimeVersion(Version? runtimeVersion) => runtimeVersion != null ? $"CLR {runtimeVersion}; " : string.Empty;
```
JamesNK, line 77: Avoid possibility of accidental closure
```suggestion
            static string GetRuntimeVersion(Version? runtimeVersion) => runtimeVersion != null ? $"CLR {runtimeVersion}; " : string.Empty;
```

src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            string GetProcessArch(Architecture processArch) => processArch.ToString().ToLowerInvariant();
```
JamesNK, line 79:
```suggestion
            static string GetProcessArch(Architecture processArch) => processArch.ToString().ToLowerInvariant();
```

src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
```diff
@@ -0,0 +1,164 @@
...
+                // .NET Framework 4.7.1 and above. If targetting net461 or earlier,
+                // the framework description will need to be resolved manually
+                // using reflection.
+                runtimeInformation: RuntimeInformation.FrameworkDescription,
+                frameworkName: frameworkName,
+                operatingSystem: GetOS());
+
+            bool IsMono() => Type.GetType ("Mono.Runtime") != null;
+        }
+
```
JamesNK, line 57:
```suggestion
```

src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
```diff
@@ -0,0 +1,164 @@
...
+                assemblyVersion: assemblyVersion,
+                // RuntimeInformation.FrameworkDescription is only supported for
+                // .NET Framework 4.7.1 and above. If targetting net461 or earlier,
+                // the framework description will need to be resolved manually
+                // using reflection.
+                runtimeInformation: RuntimeInformation.FrameworkDescription,
+                frameworkName: frameworkName,
+                operatingSystem: GetOS());
+
+            bool IsMono() => Type.GetType ("Mono.Runtime") != null;
```
JamesNK, line 55:
```suggestion
            static bool IsMono() => Type.GetType("Mono.Runtime") != null;
```

JamesNK: Is this the recommended way to test for mono?

captainsafia: Seems to be. It's used extensively in the dotnet repos: https://dotnet.grep.app/search?q=IsMono&words=true

JamesNK (approved):

src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
```diff
@@ -0,0 +1,163 @@
...
+            // /2.41.0-dev
+            userAgent += $"{GetGrpcDotNetVersion(assemblyVersion)} ";
+            // (.NET 5.0.7;
+            userAgent += $"({GetFrameworkDescription(runtimeInformation)}";
+            // CLR 5.0.0;
+            userAgent += GetRuntimeVersion(runtimeVersion);
+            // net6.0;
+            userAgent += GetFrameworkName(frameworkName);
+            // windows  
+            userAgent += $"{operatingSystem} ";
```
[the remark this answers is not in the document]
jtattermusch: nit: I think when OS is empty string, you'll end up with one extra space in the resulting userAgentString.

jtattermusch (commented):
Overall this looks fine, but I had a few more comments.

Also, could you copypaste the actual user agent strings that were generated on the platforms/frameworks where you tried running this manually?

test/Grpc.Net.Client.Tests/UserAgentGeneratorTests.cs
```diff
@@ -0,0 +1,134 @@
...
+        public void HandlesMonoRuntimeInfo()
+        {
+            var userAgent = UserAgentGenerator.GetUserAgentString(
+                processArch: Architecture.X64,
+                runtimeVersion: null,
+                assemblyVersion: string.Empty,
+                runtimeInformation: "Mono 6.12.0.140 (2020-02/51d876a041e Thu Apr 29 10:44:55 EDT 2021)",
+                frameworkName: string.Empty,
+                operatingSystem: "osx");
+            Assert.AreEqual($"grpc-dotnet (Mono 6.12.0.140; osx x64)", userAgent);
```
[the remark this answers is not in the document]
captainsafia: `Environment.Version` isn't actually supported on Mono from my attempts but there's an alternative way to get the CoreCLR version if we want consistency here.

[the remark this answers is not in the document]
jtattermusch: in what sense it's not supported? I'm pretty sure I've been able to retrieve its value on mono and it has returned something that seemed reasonable.
Also, it must exist on mono and return something since Environment.Version existed in .NET Framework since version 1.1 and mono has to implement the API somehow.

Issue #1406, opened 2021-09-08
NullReferenceException with channel interceptor and retry policy


DavidCMulder:
I have configured a grpc client with 
- an interceptor for setting an oauth token on the metadata 
- a retry policy configured on the ServiceConfig.
This works, but the first call results in a (handled?) null reference exception when debugging and the below flow:
```
[11:34:13 VRB] Starting retry worker.
[11:34:13 DBG] Starting attempt 1.
[11:34:13 DBG] Starting gRPC call. Method type: 'Unary', URI: 'https:...'.
[11:34:37 INF] Retrieved token for scope '...' . Token expires at 08/09/2021 10:34:36. Now is 08/09/2021 09:34:37
[11:34:37 DBG] Evaluated retry for failed gRPC call. Status code: 'Internal', Attempt: 1, Retry: False
[11:34:37 DBG] Call commited. Reason: FatalStatusCode
[11:34:37 VRB] Stopping retry worker.
[11:34:37 VRB] Serialized '...' to 8 byte message.
[11:34:37 DBG] Sending message.
[11:34:37 VRB] Message sent.
[11:34:39 VRB] Response headers received.
[11:34:39 DBG] Reading message.
[11:34:39 VRB] Deserializing 0 byte message to '...'.
[11:34:39 VRB] Received message.
[11:34:39 DBG] Finished gRPC call.
```
The exception details are:
```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Grpc.Net.Client
  StackTrace:
   at Grpc.Net.Client.Internal.Retry.RetryCall`2.<StartRetry>d__7.MoveNext()
```

Any subsequent calls flow through ok:
```
[11:54:30 VRB] Starting retry worker.
[11:54:30 DBG] Starting attempt 1.
[11:54:30 DBG] Starting gRPC call. Method type: 'Unary', URI: 'https:...'.
[11:54:30 INF] Retrieved token for scope '...' . Token expires at 08/09/2021 10:34:36. Now is 08/09/2021 09:54:30
[11:54:30 VRB] Serialized '...' to 8 byte message.
[11:54:30 VRB] Message with 13 bytes added to the buffer. There are 13 bytes buffered for this call.
[11:54:30 DBG] Sending message.
[11:54:30 VRB] Message sent.
[11:54:32 VRB] Response headers received.
[11:54:32 DBG] Reading message.
[...]

PR #1414, opened 2021-09-17
Fix combining auth and retry

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1406

Adds a `TaskCompletionSource<HttpResponseMessage>` so that the task to get the response is never null.

Best reviewed with hide whitespace enabled.

### DIFF
--- a/src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
+++ b/src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
@@ -81,6 +81,13 @@ namespace Grpc.Net.Client.Internal.Retry
                     startCallFunc(call);
 
                     SetNewActiveCallUnsynchronized(call);
+
+                    if (CommitedCallTask.IsCompletedSuccessfully())
+                    {
+                        // Call has already been commited. This could happen if written messages exceed
+                        // buffer limits, which causes the call to immediately become commited and to clear buffers.
+                        return;
+                    }
                 }
 
                 Status? responseStatus;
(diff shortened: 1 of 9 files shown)


src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
```diff
@@ -60,151 +60,157 @@ public HedgingCall(HedgingPolicyInfo hedgingPolicy, GrpcChannel channel, Method<
...
+                        if (_activeCalls.Remove(call))
+                        {
+                            call.Dispose();
+                        }
                     }
                 }
             }
-
-            if (CommitedCallTask.IsCompletedSuccessfully() && CommitedCallTask.Result == call)
+            finally
```
captainsafia, line 206: I noticed we don't do any except handling on the top-level of the `StartCall` here. Is it worthwhile to do here or is the exception handling at the call sites sufficient?

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
```diff
@@ -60,151 +60,157 @@ public HedgingCall(HedgingPolicyInfo hedgingPolicy, GrpcChannel channel, Method<
...
-            try
-            {
-                if (call._httpResponseTask == null)
+                HttpResponseMessage? httpResponse = null;
+                try
                 {
-                    // There is no response task if there was a preemptive cancel.
-                    CompatibilityHelpers.Assert(call.CancellationToken.IsCancellationRequested, "Request should have been made if call is not preemptively cancelled.");
-                    call.CancellationToken.ThrowIfCancellationRequested();
+                    httpResponse = await call.HttpResponseTask.ConfigureAwait(false);
```
captainsafia, line 91:
```suggestion
                    httpResponse = await call!.HttpResponseTask.ConfigureAwait(false);
```

`call` is guaranteed to be set here, correct? Annotation makes it a little easier to read since it's nullable.

captainsafia (approved):

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
JamesNK: Yes, a call is always set at this point.

I think an annotation makes it less clear what is going on. Since this compiles, code analysis guarantes that it's not null here. An annotation would make me wonder why this could be null.

src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
JamesNK: All logic that can throw should be inside a try/catch.

However, a global catch couldn't hurt. Also, throwing on CTS cancellation and reusing `HandleUnexpectedError` simplifies the code.

Updated.

PR #1423, opened 2021-09-23
Address followups in UserAgentGenerator code

captainsafia:
Address post-merge feedback on https://github.com/grpc/grpc-dotnet/pull/1411 and document generated examples.

### DIFF
--- a/src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
+++ b/src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
@@ -56,7 +56,7 @@ namespace Grpc.Net.Client.Internal
             
             return GetUserAgentString(
                 processArch: RuntimeInformation.ProcessArchitecture,
-                runtimeVersion: IsMono ? GetMonoRuntimeVersion(assembly) : Environment.Version, 
+                clrVersion: Environment.Version,
                 assemblyVersion: assemblyVersion,
                 // RuntimeInformation.FrameworkDescription is only supported for
                 // .NET Framework 4.7.1 and above. If targetting net461 or earlier,
@@ -65,12 +65,10 @@ namespace Grpc.Net.Client.Internal
                 runtimeInformation: RuntimeInformation.FrameworkDescription,
                 frameworkName: frameworkName,
                 operatingSystem: GetOS());
-
-            static Version GetMonoRuntimeVersion(Assembly assembly) => Version.Parse(assembly.ImageRuntimeVersion.Substring(1));
         }
 
         // Factored out for testing
-        internal static string GetUserAgentString(Architecture processArch, Version? runtimeVersion, string? assemblyVersion, string? runtimeInformation, string? frameworkName, string? operatingSystem)
+        internal static string GetUserAgentString(Architecture processArch, Version? clrVersion, string? assemblyVersion, string? runtimeInformation, string? frameworkName, string? operatingSystem)
         {
             var userAgent = "grpc-dotnet";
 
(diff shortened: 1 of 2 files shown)


JamesNK:
I think the test failure is fixed in this PR - https://github.com/grpc/grpc-dotnet/pull/1399

captainsafia:
> I think the test failure is fixed in this PR - #1399

Got it. I saw some of these test failures in my last PR but attributed it to general flakiness. Are the tests in this repo generally bullet proof and do "flaky" failures typically indicate an underlying bug?

JamesNK:
Usually, it is flakiness but the tests aren't bullet-proof. There is a lot of multithreaded code in the solution and bugs and flakiness can hide for a long time.

src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
```diff
@@ -44,7 +56,7 @@ internal static string GetUserAgentString()
             
             return GetUserAgentString(
                 processArch: RuntimeInformation.ProcessArchitecture,
-                runtimeVersion: IsMono ? null : Environment.Version, 
+                runtimeVersion: IsMono ? GetMonoRuntimeVersion(assembly) : Environment.Version,
```
jtattermusch, line 59: this gets confusing. `Environment.Version` is actually the CLR Version (see docs https://docs.microsoft.com/en-us/dotnet/api/system.environment.version?view=net-5.0) so it might be less confusing to call the argument "clrVersion" rather than "runtimeVersion" since "runtimeVersion" gets easily confused with "runtimeInformation"

jtattermusch: Environment.Version has existed since .NET Framework 1.1 so it really should exist on mono.
What doesn't necessarily exist on mono is the RuntimeInformation.FrameworkDescription (only exists on .NET Framework 4.7.1, but that's hidden by the fact that we're currently targeting netstandard2.0 and not e.g. net461)

src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
```diff
@@ -103,15 +117,15 @@ private static string GetOS()
...
             {
-                return "osx";
+                return "osx; ";
             }
             else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
             {
-                return "linux";
+                return "linux; ";
             }
             else
```
jtattermusch, line 130: nit: do we have a test where OS is empty string? I think in practice that can happen e.g. when on Xamarin Android, Xamarin iOS etc, so not just a theoretical possibility.

captainsafia: Added.

src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
```diff
@@ -79,15 +77,15 @@ internal static string GetUserAgentString(Architecture processArch, Version? run
...
+            userAgent += GetclrVersion(clrVersion);
             // net6.0;
             userAgent += GetFrameworkName(frameworkName);
             // windows  
             userAgent += $"{operatingSystem}";
             // x64)
             userAgent += $"{GetProcessArch(processArch)})";
 
-            static string GetRuntimeVersion(Version? runtimeVersion) => runtimeVersion != null ? $"CLR {runtimeVersion}; " : string.Empty;
+            static string GetclrVersion(Version? clrVersion) => clrVersion != null ? $"CLR {clrVersion}; " : string.Empty;
```
jtattermusch, line 88: nit: `GetclrVersion` -> `GetClrVersion`

Issue #1394, opened 2021-08-30
Duplex streaming problems like: The client reset the request stream.


istvanek:
Hello,
in our production application we have gRPC

- server using ASP.NET Core 5 implementation
- .NET Framework 4.8 clients using C# Grpc.Core implementation

Duplex streaming calls are used in various cases like
- notifications about data changes
- image streaming

Clients call the server, stream requests, read streamed responses and close request stream.

Sometimes we see a call ends with errors on server in requestStream.MoveNext, but has no problem on clients.
This is not problematic when we do all processing during streaming and error on server means actually a client has ended the call.
But this is problematic in cases when we do some processing on server after a client call end, but due to the error the processing is skipped and client just thinks everything is ok, which is not obviously.

To reproduce the problem, I have prepared testing service and client [attached to this post](https://github.com/grpc/grpc-dotnet/files/7076675/GrpcDuplexStreaming.zip):
- service uses ASP.NET Core 5 implementation
- client may choose .NET or Grpc.Core implementation, but both show the problem

Steps to reproduce the problem:
- start service
- start client and choose .NET or Grpc.Core implementation
- client calls the service duplex streaming call with short delay (100 ms which can be changed with +- keys) between responses
- service processes the call with short delay (100 ms which can be changed with +- keys) between responses
- to end the call on client and start another one, press any key (eg. space)
- depending on the call duration, the call ends well or with errors in the service - just press a key on client few times after various times
- to continue the service after any error, press Enter

The errors I have seen are:
- info: Grpc.AspNetCore.Server.ServerCallHandler[14]
      Error reading message.
      System.IO.IOException: The client reset the request stream.
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestPipeReader.AdvanceTo(SequencePosition consumed, SequencePosition examined)
         at Grpc.AspNetCore.Server.Internal.PipeExtensions.ReadStreamMessageAsync[T](PipeReader input, HttpContextServerCallContext serverCallContext, Func`2 deserializer, CancellationToken cancellationToken)
System.IO.IOException: The client reset the request stream.
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestPipeReader.AdvanceTo(SequencePosition consumed, SequencePosition examined)
[...]

PR #1429, opened 2021-09-29
Fix race between CompleteAsync and call dispose

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/1394


test/FunctionalTests/Client/StreamingTests.cs
```diff
@@ -996,5 +997,90 @@ async Task WaitForAllStreams(IAsyncStreamReader<DataMessage> requestStream, ISer
...
+                        var flags = state.Single(s => s.Key == "flags").Value.ToString();
+
+                        if (frameType == "DATA" && flags == "END_STREAM")
+                        {
+                            return true;
+                        }
+                    }
+
+                    return false;
+                });
```
JamesNK, line 1079: This is the only way I could figure out how to unit test this. I think if this test becomes troublesome then it could be deleted.

test/FunctionalTests/Client/StreamingTests.cs
```diff
@@ -996,5 +997,90 @@ async Task WaitForAllStreams(IAsyncStreamReader<DataMessage> requestStream, ISer
             }
         }
 #endif
+
+        [TestCase(2)]
```
captainsafia, line 1001: I might've missed the nuance in the issue text but why are we testing on iterations up to `2` and `20`?

JamesNK: The race condition happened randomly.

When I was testing, I had 2 iterations so the logs weren't too long in one test run, and 20 in another run to ensure the race condition didn't happen over many iterations.

captainsafia (approved):

JamesNK:
I hate the test. Removed.

Issue #1044, opened 2020-09-07
when using AddGrpcClient, Grpc Channel is not reused for clients of the same type?


NetGH:
Why grpc clients of the same type from the same service provider do not share the same channel? It supposes to reuse the same channel for the same address for efficiency right?

`            var services = new ServiceCollection();
            services.AddGrpcClient<AnyNamespace.AnyClient>(options =>
            {
                options.Address = new Uri("http://localhost");
            });
            var serviceProvider = services.BuildServiceProvider();

            var dummyRpcClient = serviceProvider.GetRequiredService<AnyNamespace.AnyClient>();
            var anotherDummyRpcClient = serviceProvider.GetRequiredService<AnyNamespace.AnyClient>();
            Assert.AreNotSame(dummyRpcClient, anotherDummyRpcClient);

            var callInvokerField =
                typeof(ClientBase).GetField("callInvoker", BindingFlags.Instance | BindingFlags.NonPublic);

            var callInvoker = (CallInvoker)callInvokerField.GetValue(dummyRpcClient);
            var anotherCallInvoker = (CallInvoker)callInvokerField.GetValue(anotherDummyRpcClient);

            var invokerField =
                callInvoker.GetType().GetField("invoker", BindingFlags.Instance | BindingFlags.NonPublic);

            var invoker = invokerField.GetValue(callInvoker);
            var anotherInvoker = invokerField.GetValue(anotherCallInvoker);

            var channelProperty = invoker
                .GetType()
                .GetProperty(nameof(Channel), BindingFlags.NonPublic | BindingFlags.Instance);

            var channel = (GrpcChannel) channelProperty.GetValue(invoker);
            var anotherChannel = (GrpcChannel)channelProperty.GetValue(anotherInvoker);

            Assert.AreSame(channel, anotherChannel);`

JamesNK:
At the moment the `HttpClient` is cached but a new `GrpcChannel` is created for each client. This is ok because `HttpClient` is where the heavy state, e.g. the connection, is stored.

Is not reusing the channel a problem?

d1820:
@JamesNK  that seems contradictory to the Best Practice recommendations from Microsoft  [Reuse gRPC channels](https://docs.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-5.0). Am I missing something?

JamesNK:
HttpClient is the important thing to reuse *right now*.

[...]

PR #1430, opened 2021-09-30
Refactor client factory channel and interceptor lifetime

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1264
Fixes https://github.com/grpc/grpc-dotnet/issues/842
Fixes https://github.com/grpc/grpc-dotnet/issues/1044

Client factory currently recreates the channel for every client instance that is created. This wasn't ideal, but all the heavy state for in a channel was in the HttpHandler, which was cached.

Features added in the last year mean channel now has state. Retries keep track of failed calls for throttling, and load balancing opens its own sockets. Recreating channel for each client instance is no longer an option.

This PR:
* Changes the client factory to have a singleton channel for each configured client. That means multiple instances of the client will use the same channel.
* Updates interceptors to no longer always be singletons. There is now an `InterceptorLifetime` setting. By default, interceptors have a channel lifetime, so will be created once for a channel. But configuring an interceptor to have a client lifetime means it will get resolved per-client instance.


src/Grpc.Net.ClientFactory/GrpcHttpClientBuilderExtensions.cs
```diff
@@ -127,6 +136,18 @@ public static IHttpClientBuilder AddInterceptor(this IHttpClientBuilder builder,
         /// <param name="configureInvoker">A delegate that is used to create an <see cref="Interceptor"/>.</param>
         /// <returns>An <see cref="IHttpClientBuilder"/> that can be used to configure the client.</returns>
         public static IHttpClientBuilder AddInterceptor(this IHttpClientBuilder builder, Func<Interceptor> configureInvoker)
```
captainsafia, line 138: Nit: document this has a `Channel` lifetime by default?

src/Grpc.Net.ClientFactory/Internal/DefaultGrpcClientFactory.cs
```diff
@@ -53,19 +51,26 @@ internal class DefaultGrpcClientFactory : GrpcClientFactory
...
-            var callInvoker = _callInvokerFactory.CreateCallInvoker(httpHandler, name, typeof(TClient), clientFactoryOptions);
+
+            var resolvedCallInvoker = GrpcClientFactoryOptions.BuildInterceptors(
+                callInvoker,
+                _serviceProvider,
+                clientFactoryOptions,
+                InterceptorLifetime.Client);
+
+#pragma warning disable CS0618 // Type or member is obsolete
+            if (clientFactoryOptions.Interceptors.Count != 0)
```
captainsafia, line 65: Is there a universe where we need to support a mix of interceptors registered on the obsolete `Interceptors` list and the new `InterceptorRegistrations` list?

captainsafia (approved):

src/Grpc.Net.ClientFactory/Internal/DefaultGrpcClientFactory.cs
JamesNK: No. The only way the old `Interceptors` collection could have interceptors is if someone manually added them (i.e. didn't use the interceptor extension methods).

This is for backward compatibility until the old collection gets removed.

src/Grpc.Net.ClientFactory/GrpcHttpClientBuilderExtensions.cs
JamesNK: Good idea.

I kind of want to rename `InterceptorLifetime`. It's more like a resolution scope. Someone could still register an interceptor as a singleton in DI and have a per-client interceptor be shared because it is a singleton in DI.

rizi:
> The sample below will always create a single instance because the interceptor is being configured as the channel credentials:
> 
> ```cs
> services
>     .AddGrpcClient<Greeter.GreeterClient>(o =>
>     {
>         o.Address = new Uri("https://localhost:5001");
>     })
>     .ConfigureChannel(o =>
>     {
>         var credentials = CallCredentials.FromInterceptor((context, metadata) =>
>         {
>             if (!string.IsNullOrEmpty(_token))
>             {
>                 metadata.Add("Authorization", $"Bearer {_token}");
>             }
>             return Task.CompletedTask;
>         });
> 
>         o.Credentials = ChannelCredentials.Create(new SslCredentials(), credentials);
>     });
> ```
> 
> What do you want the lifetime of an interceptor to be? Why do you want to have that lifetime? Can you share the source code of what the interceptor would look like?

I know that sample above creates a "singleton", but I want the interceptor to be resolved from the DI before every call, here was my issue I created a year ago, it's marked as fix with your merge/pull request: 
https://github.com/grpc/grpc-dotnet/issues/842

JamesNK:
Unit test example here:

https://github.com/JamesNK/grpc-dotnet/blob/4dd048f1cb3160751cae2e8b171201027e4e6a70/test/Grpc.Net.ClientFactory.Tests/GrpcHttpClientBuilderExtensionsTests.cs#L374-L420

Interceptor is registered client factory using `InterceptorScope.Client`. And the interceptor type is registered with DI as scoped. That means the interceptor is created once per-scope, and can be injected with types from the dependency injection scope.

rizi:
> Unit test example here:
> 
> https://github.com/JamesNK/grpc-dotnet/blob/4dd048f1cb3160751cae2e8b171201027e4e6a70/test/Grpc.Net.ClientFactory.Tests/GrpcHttpClientBuilderExtensionsTests.cs#L374-L420
> 
> Interceptor is registered client factory using `InterceptorScope.Client`. And the interceptor type is registered with DI as scoped. That means the interceptor is created once per-scope, and can be injected with types from the dependency injection scope.

Thx for the Infos, just to be sure, the clients will be created per scope/request?

And the channel is only created once and shared between the clients?

Br

rizi:
> * A client will be created every time it is resolved from DI, just like today.
> * Clients will use the same channel each time they are created, e.g. every instance of GreeterClient will share a channel, and every instance of CounterClient will share a different channel.
> * Interceptors can be configured to be resolved once for the channel (like today) or to be resolved each time a client is created using the current DI scope.

Thx for your patience, I think that's exactly what I need.

One last question: 
That means a channel for e.g GreeterClient is exactly created once, correct?

JamesNK:
Created once per client type. If your app has GreeterClient and CounterClient then a maximum of two will be created.

Although I might change that so a single channel is created for both if no channel settings are changed.

amsga89:
builder.Services
    .AddGrpcClient<Greeter.GreeterClient>(o =>
    {
        o.Address = new Uri("https://localhost:5001");
    })
    .AddInterceptor<LoggingInterceptor>(**InterceptorScope.Client**);

PR #1433, opened 2021-09-30
[v2.40.x] Update Google.Protobuf to 3.18.0

JamesNK:
I want to update .NET SDK template to reference 2.40.0 and to use a recent Google.Protobuf. I noticed that the Google.Protobuf version referenced here is old.

I think this is a low-risk change. The only impact is to Grpc.AspNetCore so it brings in 3.18.0 instead of the old package version.


jtattermusch (approved):
LGTM.

PR #1434, opened 2021-09-30
[v2.40.x] Remove load balancing for final release

JamesNK:


melfeqy:
Is load balancer not supported anymore ?

JamesNK:
It is still supported, but just in preview releases for now.

Issue #1219, opened 2021-03-04
Stream reader MoveNext() doesn't comply with gRPC spec, throws exception on stream abort/close


markbearden:
Using Grpc.AspNetCore v2.33.1 (in .NET Core 3.1 project).

The behavior of async stream reading for an open gRPC connection doesn't match the main gRPC API documentation in the following respect.  The [API doc for IAsyncStreamReader<T>](https://grpc.github.io/grpc/csharp/api/Grpc.Core.IAsyncStreamReader-1.html) says (with my bold emphasis):

> On the server side, MoveNext() **does not throw exceptions**. **In case of a failure,** the request stream will appear to be finished (**MoveNext will return false**) and the CancellationToken associated with the call will be cancelled to signal the failure. 

But the .NET implementation of [HttpContextStreamReader.MoveNext()](https://github.com/grpc/grpc-dotnet/blob/978200abd58a843b367a109c321347c62eef58f3/src/Grpc.AspNetCore.Server/Internal/HttpContextStreamReader.cs#L49) **does** throw an exception if the underlying TCP/HTTP2 transports are abnormally closed.

I am experiencing this, and I found another case that looks like mine [already posted publicly here](https://source.cloud.google.com/results/invocations/372f4c19-991e-41f1-8759-b6967b288b42/log), so I'll refer to that in lieu of posting my own error detail.  (Search for the `System.IO.Exception error `returned from the "FullDuplexCall" unit test, and note how `HttpContextStreamReader`1.<MoveNext>g__MoveNextAsync` is in the exception stack trace.)

I can clearly see the exception-throwing in the current source as follows:

- Within `MoveNext()`, [this call to `ReadStreamMessageAsync()`](https://github.com/grpc/grpc-dotnet/blob/978200abd58a843b367a109c321347c62eef58f3/src/Grpc.AspNetCore.Server/Internal/HttpContextStreamReader.cs#L66) is not enclosed with a try/catch.
- Within `ReadStreamMessageAsync(`), any exception that's not an `OperationCanceledException` [is rethrown within the catch](https://github.com/grpc/grpc-dotnet/blob/978200abd58a843b367a109c321347c62eef58f3/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs#L331).

[...]

PR #1435, opened 2021-09-30
Avoid throwing InvalidOpEx in server stream

erikma:
Return false instead of throwing InvalidOperationException on read of a gRPC server stream after the connection was closed.

In porting a large application from the deprecated Google gRPC server to ASP.NET, this exception path required adding an additional catch where none had been needed before. Related to #1219 but not a complete fix for that issue.

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextStreamReader.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextStreamReader.cs
@@ -58,7 +58,8 @@ namespace Grpc.AspNetCore.Server.Internal
 
             if (_completed || _serverCallContext.CancellationToken.IsCancellationRequested)
             {
-                return Task.FromException<bool>(new InvalidOperationException("Can't read messages after the request is complete."));
+                // gRPC specification indicates that MoveNext() should not throw. Simply return false.
+                return CommonGrpcProtocolHelpers.FalseTask;
             }
 
             var request = _serverCallContext.HttpContext.Request.BodyReader.ReadStreamMessageAsync(_serverCallContext, _deserializer, cancellationToken);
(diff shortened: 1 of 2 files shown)


JamesNK:
Hmmm, I disagree with what Grpc.Core is doing. Returning false instead of an error seems like an easy way for people to unknowingly ignore errors.

Perhaps there should be a setting to opt-in to this behavior. People who are porting from Grpc.Core can enable it once on the server.

@jtattermusch

erikma:
The case here in the existing client-server implementation is that a Task is kept awaiting the server stream while the client disconnects, as the server must continue to accept any late-arriving client messages. On the Google server this simply returns false once the client closes the connection.

jtattermusch:
> Hmmm, I disagree with what Grpc.Core is doing. Returning false instead of an error seems like an easy way for people to unknowingly ignore errors.
> 
> Perhaps there should be a setting to opt-in to this behavior. People who are porting from Grpc.Core can enable it once on the server.
> 
> @jtattermusch

The question is whether receiving cancellation from the client on the server side is an "error" that needs to be handled/reported. For some RPCs, the client cancelling the RPC is a "normal" mode of operation in the sense that the client announces that given streaming RPC is no longer needed (since e.g. a bidi call be be used to represent a sequence of "commands" sent by the client. Or if the RPC is used to send updates from the server, client cancelling the RPC is just a signal to stop sending updates). 
The natural consequence of client cancelling seems to be that the request stream suddenly ends (since we know that client is done sending). If the client cancels, the server won't be able to send any more responses to the client, so arguably there's not much point in throwing in the server-side handler (sending responses back from the server will fail anyway). The cancellation flag can still be detected with serverCallContext.CancellationToken (if important for the server).

Throwing in requestStream.ReadNext() just make the code more complicated in case where cancelling the RPC is part of a normal workflow (and I agree that it makes migrating code from Grpc.Core more complicated, and we don't want that).

I'd be in favor of changing the grpc-dotnet behavior or at least adding a knob to configure the behavior.

PR #1464, opened 2021-11-01
Update retrier sample

JamesNK:
* Resolve unused proto warning
* Use different status code
* Use 5 retries (max is 5 on channel so higher doesn't do anything)

### DIFF
--- a/examples/Retrier/Client/Program.cs
+++ b/examples/Retrier/Client/Program.cs
@@ -79,11 +79,11 @@ namespace Client
                 Names = { MethodName.Default },
                 RetryPolicy = new RetryPolicy
                 {
-                    MaxAttempts = 10,
+                    MaxAttempts = 5,
                     InitialBackoff = TimeSpan.FromSeconds(0.5),
                     MaxBackoff = TimeSpan.FromSeconds(0.5),
                     BackoffMultiplier = 1,
-                    RetryableStatusCodes = { StatusCode.Unavailable }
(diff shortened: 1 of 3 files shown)


examples/Retrier/Client/Program.cs
```diff
@@ -79,11 +79,11 @@ private static GrpcChannel CreateChannel()
                 Names = { MethodName.Default },
                 RetryPolicy = new RetryPolicy
                 {
-                    MaxAttempts = 10,
+                    MaxAttempts = 5,
                     InitialBackoff = TimeSpan.FromSeconds(0.5),
                     MaxBackoff = TimeSpan.FromSeconds(0.5),
                     BackoffMultiplier = 1,
-                    RetryableStatusCodes = { StatusCode.Unavailable }
+                    RetryableStatusCodes = { StatusCode.Internal }
```
captainsafia, line 86: Why the change here? This doc (https://grpc.github.io/grpc/core/md_doc_statuscodes.html) seems to indicate that Unavailable is the status code we should listen on to determine whether or not to retry. `Internal` seems to indicate a total failure that a retry cannot recover from.

JamesNK: I thought to use Internal to represent a server error, but you're right that it is more appropriate for the server to return unavailable and the client to retry on that.

Updated.

captainsafia (approved):

PR #1465, opened 2021-11-02
Fix trim warnings

JamesNK:
See https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming

### DIFF
--- a/src/Grpc.AspNetCore.Server/InterceptorRegistration.cs
+++ b/src/Grpc.AspNetCore.Server/InterceptorRegistration.cs
@@ -78,6 +78,10 @@ namespace Grpc.AspNetCore.Server
         private IGrpcInterceptorActivator? _interceptorActivator;
         private ObjectFactory? _factory;
 
+#if NET5_0_OR_GREATER
+        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+            Justification = "Type parameter members are preserved with DynamicallyAccessedMembers on InterceptorRegistration.Type property.")]
+#endif
         internal IGrpcInterceptorActivator GetActivator(IServiceProvider serviceProvider)
         {
             // Not thread safe. Side effect is resolving the service twice.
--- a/src/Grpc.Net.Client/Internal/GrpcCall.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCall.cs
@@ -762,7 +762,7 @@ namespace Grpc.Net.Client.Internal
 
                 if (diagnosticSourceEnabled)
                 {
-                    WriteDiagnosticEvent(GrpcDiagnostics.DiagnosticListener, GrpcDiagnostics.ActivityStartKey, new { Request = request });
+                    WriteDiagnosticEvent(GrpcDiagnostics.DiagnosticListener, GrpcDiagnostics.ActivityStartKey, new ActivityStartData(request));
                 }
             }
 
@@ -826,29 +826,13 @@ namespace Grpc.Net.Client.Internal
                         activity.SetEndTime(DateTime.UtcNow);
                     }
 
-                    WriteDiagnosticEvent(GrpcDiagnostics.DiagnosticListener, GrpcDiagnostics.ActivityStopKey, new { Request = request, Response = HttpResponse });
+                    WriteDiagnosticEvent(GrpcDiagnostics.DiagnosticListener, GrpcDiagnostics.ActivityStopKey, new ActivityStopData(HttpResponse, request));
                 }
 
                 activity.Stop();
             }
 
             return true;
-        }
-
-#if NET5_0_OR_GREATER
-        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
-            Justification = "The values being passed into Write have the commonly used properties being preserved with DynamicDependency.")]
-#endif
-        private static void WriteDiagnosticEvent<
-#if NET5_0_OR_GREATER
-            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)]
-#endif
-            TValue>(
-            DiagnosticSource diagnosticSource,
-            string name,
-            TValue value)
-        {
-            diagnosticSource.Write(name, value);
         }
 
         private bool IsDeadlineExceededUnsynchronized()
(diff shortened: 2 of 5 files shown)


src/Grpc.Net.Client/Internal/UserAgentGenerator.cs
```diff
@@ -26,8 +26,6 @@ namespace Grpc.Net.Client.Internal
 {
     internal static class UserAgentGenerator
     {
-        internal static bool IsMono { get; } = Type.GetType("Mono.Runtime") != null;
```
captainsafia, line 29: Good catch!

captainsafia (approved):

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -835,6 +835,22 @@ private bool FinishCall(HttpRequestMessage request, bool diagnosticSourceEnabled
             return true;
         }
 
+#if NET5_0_OR_GREATER
+        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+            Justification = "The values being passed into Write have the commonly used properties being preserved with DynamicDependency.")]
```
MichalStrehovsky, line 840: The `[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)]` annotation will keep properties on the `TValue`, but it will not do that recursively. Diagnostic source might recourse into properties of the properties. See some discussion in https://github.com/dotnet/runtime/issues/50454.

I don't think this code would still work as intended if the trimmer were to trim all properties on `HttpRequestMessage` (which is one of the property types in question in this file). There's nothing preventing that here.

JamesNK: That's true, I don't think there is anything else that can be done.

JamesNK: (btw I copied this technique from dotnet/runtime so I'm guessing the code was inspired by that issue or vice versa)

MichalStrehovsky: I think in the dotnet/runtime repo we ensure the properties on nested things are preserved through other means, if there are any. We were pretty careful not to suppress warnings that could result in different behavior after trimming.

The warnings are the only thing between "the app works as intended after trimming" and "the app is broken after I finished all my debugging, I've decided to publish a release build, and none of my unit tests caught it".

The annoying thing about trimming is that it cannot be unit tested (unit testing changes the set of things that are kept) so all trimming related failures only show up once one does E2E testing. E2E testing for many apps means "manually going through many scenarios". Trimming warnings are the only line of defense we have and an invalid suppression often ends up being a disservice to the customer.

eerhardt: Check out what we did in `System.Net.Http.DiagnosticsHandler` in https://github.com/dotnet/runtime/pull/50265.

We added `DynamicDependency` attributes for the important properties (i.e. the ones that are automatically selected by [Microsoft.Diagnostics.Monitoring.EventPipe](https://github.com/dotnet/diagnostics/blob/ffd0254da3bcc47847b1183fa5453c0877020abd/src/Microsoft.Diagnostics.Monitoring.EventPipe/Configuration/HttpRequestSourceConfiguration.cs#L36-L40)). For example:

https://github.com/dotnet/runtime/blob/395038186623985b48a1ac1ccb06a77291f2f2ed/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L218-L226

https://github.com/dotnet/runtime/blob/395038186623985b48a1ac1ccb06a77291f2f2ed/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L296-L299

This ensures the trimmer won't remove these properties from `HttpRequestMessage`, `HttpResponseMessage`, or even `Uri`.

Unfortunately DiagnosticSource's design is intrinsically incompatible with trimming. We make a "best effort" to ensure the important properties are kept. And we added to the error message that gets logged when a property cannot be found at runtime:

> "If you published the application with PublishTrimmed=true, ensure the property was not trimmed away."

So if a user is trying to get a property that was trimmed, the developer can explicitly add it to the list of properties to be kept.

src/Grpc.AspNetCore.Server/InterceptorRegistration.cs
```diff
@@ -78,6 +78,10 @@ public class InterceptorRegistration
         private IGrpcInterceptorActivator? _interceptorActivator;
         private ObjectFactory? _factory;
 
+#if NET5_0_OR_GREATER
+        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+            Justification = "Type parameter members are preserved with DynamicallyAccessedMembers on InterceptorRegistration.Type property.")]
```
eerhardt, line 83: Hmm, I thought we fixed this in the trimmer, but I guess we only fixed it for [MakeGenericMethod](https://github.com/dotnet/linker/issues/1727) and not MakeGenericType.

@vitek-karas @MichalStrehovsky - do you know why we didn't do the same for MakeGenericType? In this case, `typeof(IGrpcInterceptorActivator<>)` doesn't have any trimming annotations on it. So calling `MakeGenericType` should be safe.

Or is this the case that the Trim Analyzer blindly injects this warning and doesn't have the same smarts as the real ILLinker?

src/Grpc.Net.Client/Internal/GrpcCall.cs
JamesNK: Better? [`0218ab6` (#1465)](https://github.com/grpc/grpc-dotnet/pull/1465/commits/0218ab6c04b4de45f4abeb994474c5b8b8f83da5)

I understand it's not perfect.

eerhardt (approved):
FWIW - this change looks good to me.

src/Grpc.AspNetCore.Server/InterceptorRegistration.cs
vitek-karas: This is unfortunate. The analyzer has been fixed in https://github.com/dotnet/linker/pull/2209, but apparently this change didn't make it into the 6.0. It should be in 6.0.2xx though.

It is analyzer only problem - I would suggest we suppress this with #pragma instead of the attribute.

Issue #1470, opened 2021-11-04
`IGrpcServiceActivator` missing `DynamicallyAccessedMembers` attribute on its generic argument type


gitfool:
### What version of gRPC and what language are you using?
gRPC 2.37.0 and C#

### What operating system (Linux, Windows,...) and version?
Linux (Ubuntu 20.04) and Windows (10 21H1)

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.NET SDK 6.0.100-rc.2.21505.57

### What did you do?
See https://github.com/pulumi/pulumi/issues/8153#issuecomment-953411725 for background.

TL;DR: Simply retargeting from net5.0 to net6.0 hit an issue with linker trimming of single file executable:
```
Generic implementation type 'Grpc.AspNetCore.Server.Internal.DefaultGrpcServiceActivator`1' has a DynamicallyAccessedMembers attribute applied to a generic argument type, but the service type
'Grpc.AspNetCore.Server.IGrpcServiceActivator`1' doesn't have a matching DynamicallyAccessedMembers attribute on its generic argument type.
```

If I understand correctly, it looks like this is not fixed in the latest source as [`IGrpcServiceActivator`](https://github.com/grpc/grpc-dotnet/blob/d5682d482325a337ccaa669adbae0354a64bee22/src/Grpc.AspNetCore.Server/IGrpcServiceActivator.cs#L28) still does not have the `DynamicallyAccessedMembers` attribute as in [`DefaultGrpcServiceActivator`](https://github.com/grpc/grpc-dotnet/blob/d5682d482325a337ccaa669adbae0354a64bee22/src/Grpc.AspNetCore.Server/Internal/DefaultGrpcServiceActivator.cs#L26).

JamesNK:
Thanks for reporting this. It is fixed in the next gRPC release. A preview should be released soon.

PR #1471, opened 2021-11-05
Fix server trim error from dependency injection

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1470

FYI @MichalStrehovsky @eerhardt

### DIFF
--- a/testassets/LinkerTestsClient/LinkerTestsClient.csproj
+++ b/testassets/LinkerTestsClient/LinkerTestsClient.csproj
@@ -5,7 +5,7 @@
     <OutputType>Exe</OutputType>
     <PublishTrimmed>true</PublishTrimmed>
 
-    <!-- IL2104: Other assemblies have warnings -->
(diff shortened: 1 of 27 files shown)


testassets/LinkerTestsClient/LinkerTestsClient.csproj
```diff
@@ -1,29 +1,24 @@
...
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <OutputType>Exe</OutputType>
-    
     <PublishTrimmed>true</PublishTrimmed>
-    <TrimMode>Link</TrimMode>
+
+    <!-- IL2104: Other assemblies have warnings -->
```
captainsafia, line 8: Which assemblies are these? Might help to document in case we can eventually remove this suppression.

captainsafia (approved):

PR #1472, opened 2021-11-07
Add support for per-address attributes

JamesNK:
Feedback from load balancing proposal review.

Addresses have changed from DnsEndPoint to BalancerAddress. This allows an address to have attributes (aka metadata).

### DIFF
--- a/src/Grpc.Net.Client/Balancer/BalancerAddress.cs
+++ b/src/Grpc.Net.Client/Balancer/BalancerAddress.cs
@@ -1,0 +1,71 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+#if SUPPORT_LOAD_BALANCING
+using System.Net;
+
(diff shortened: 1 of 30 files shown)


src/Grpc.Net.Client/Balancer/BalancerAddress.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Gets the address <see cref="DnsEndPoint"/>.
+        /// </summary>
+        public DnsEndPoint EndPoint { get; }
+
+        /// <summary>
+        /// Gets the address attributes.
```
captainsafia, line 58:
```suggestion
        /// Gets the <see cref="BalancerAttributes" /> associated with the address.
```

src/Grpc.Net.Client/Balancer/BalancerAddress.cs
```diff
@@ -0,0 +1,71 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
```
captainsafia, line 3:
```suggestion
// Copyright 2021 The gRPC Authors
```

Do we need to do something to updates the copyright dates in the other files? Also, why individual headers in each source file instead of a top-level one?

captainsafia (commented):
Should usage in the samples be updated as well?

src/Grpc.Net.Client/Balancer/BalancerAddress.cs
JamesNK: I looked at other grpc repos and the standard is to leave it on the date the repo started.

JamesNK:
> Should usage of `DnsEndpoint` directly in the samples be updated as well? I think it only happens in [`master`/examples/Container/Frontend/Pages/FetchData.razor#L139](https://github.com/grpc/grpc-dotnet/blob/master/examples/Container/Frontend/Pages/FetchData.razor?rgh-link-date=2021-11-08T02%3A26%3A19Z#L139).

The samples use a NuGet package and so will wait until the package version is updated.

captainsafia (approved):

PR #1473, opened 2021-11-08
Unit test least used balancer

JamesNK:
Tests to verify the experimental least used load balancer policy.

### DIFF
--- a/test/FunctionalTests/Balancer/BalancerHelpers.cs
+++ b/test/FunctionalTests/Balancer/BalancerHelpers.cs
@@ -18,6 +18,7 @@
 
 #if SUPPORT_LOAD_BALANCING
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net;
@@ -123,6 +124,7 @@ namespace Grpc.AspNetCore.FunctionalTests.Balancer
             services.AddSingleton<ResolverFactory>(new TestResolverFactory(resolver));
             services.AddSingleton<IRandomGenerator>(new TestRandomGenerator());
             services.AddSingleton<ISubchannelTransportFactory>(new TestSubchannelTransportFactory(TimeSpan.FromSeconds(0.5)));
+            services.AddSingleton<LoadBalancerFactory>(new LeastUsedBalancerFactory());
 
             var serviceConfig = new ServiceConfig();
             if (loadBalancingConfig != null)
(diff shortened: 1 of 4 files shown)


test/FunctionalTests/Balancer/BalancerHelpers.cs
```diff
@@ -164,6 +167,53 @@ public static async Task WaitForChannelStateAsync(ILogger logger, GrpcChannel ch
             logger.LogInformation($"{channelId}: Current channel state '{currentState}' matches expected state '{state}'.");
         }
 
+        public static async Task<Subchannel[]> WaitForSubChannelsToBeReadyAsync(ILogger logger, GrpcChannel channel, int expectedCount, Func<SubchannelPicker?, Subchannel[]>? getPickerSubchannels = null)
+        {
+            if (getPickerSubchannels == null)
+            {
+                getPickerSubchannels = (picker) =>
+                {
+                    if (picker is RoundRobinPicker roundRobinPicker)
```
captainsafia, line 176: #PranavPoints?

```csharp
return picker switch
{
  RoundRobinPicker roundRobinPicker => roundRobinPicker._subchannels.ToArray(),
  PickFirstPicker pickFirstPicker => new[] { pickFirstPicker.Subchannel },
  null => Array.Empty<Subchannel>(),
  _ => throw new Exception("Unexpected picker type: " + picker.GetType().FullName)
};
```

test/FunctionalTests/Balancer/LeastUsedBalancerTests.cs
```diff
@@ -0,0 +1,121 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+#if SUPPORT_LOAD_BALANCING
+#if NET5_0_OR_GREATER
```
captainsafia, line 20: Do we need this check here if `SUPPORT_LOAD_BALANCING` is only set when the `TargetFramework` is greater than `net5.0`?

test/FunctionalTests/Balancer/BalancerHelpers.cs
```diff
@@ -122,7 +123,9 @@ public static async Task<GrpcChannel> CreateChannel(ILoggerFactory loggerFactory
             var services = new ServiceCollection();
             services.AddSingleton<ResolverFactory>(new TestResolverFactory(resolver));
             services.AddSingleton<IRandomGenerator>(new TestRandomGenerator());
+            services.AddSingleton<IRandomGenerator>(new TestRandomGenerator());
```
captainsafia, line 126: Why do we inject this twice here? It seems like we always resolve a single instance anyways?

JamesNK: Copy/paste typo

captainsafia (approved):
LGTM conflicts aside.

PR #1480, opened 2021-11-10
Update version to 2.42.0-dev (on master branch)

jtattermusch:


JamesNK:
The gRPC package version needs to be updated to 2.42.0?

https://github.com/grpc/grpc-dotnet/blob/530eb1c25c8c37260aaff11cc816adf6a20570f5/build/dependencies.props#L6

JamesNK (approved):

jtattermusch:
> The gRPC package version needs to be updated to 2.42.0?
> 
> https://github.com/grpc/grpc-dotnet/blob/530eb1c25c8c37260aaff11cc816adf6a20570f5/build/dependencies.props#L6

We only have 2.42.0-pre1 for Grpc.Core at this point. Do we need to upgrade to it or is 2.41.0 fine?

PR #1479, opened 2021-11-10
Update version to 2.41.0-pre1 (on the release branch)

jtattermusch:
I will publish the pre-release right after merging this. Let me know if there's any blockers.

PR #1485, opened 2021-11-11
Sanity check for user agent string length not useful

jtattermusch:
Length of user agent string isn't as important for performance as one could think.
The reason is that HTTP/2 uses HPACK compression for headers, so the long user agent string in actually sent only once per connection (subsequent RPCs only do a table lookup to get the user agent string value). In real life, the length of user agent string will have no effect on performance even for very high qps.

Context: https://github.com/grpc/grpc-dotnet/pull/1479#issuecomment-966065439


JamesNK:
> Length of user agent string isn't as important for performance as one could think. The reason is that HTTP/2 uses HPACK compression for headers, so the long user agent string in actually sent only once per connection (subsequent RPCs only do a table lookup to get the user agent string value). In real life, the length of user agent string will have no effect on performance even for very high qps.
> 
> Context: [#1479 (comment)](https://github.com/grpc/grpc-dotnet/pull/1479#issuecomment-966065439)

Actually, HttpClient doesn't support dynamic HPACK, so the value won't get compressed.

I don't mind removing the sanity though.

JamesNK (approved):

PR #1489, opened 2021-11-12
DNS resolver fix port usage

vasicvuk:
Fixes problem with using port in DNS resolver.

Currently Port of address property is always -1 as `dns:///` is not parsed well. This PR uses only hostname for DNS resolution and allows port to be read from Uri. 

Test **DNS_Port_Works** failed before this PR

Maybe best option is to use // instead of /// for dns resolver


JamesNK:
I've made some changes to your PR, and created a new one that builds on top of it. See https://github.com/grpc/grpc-dotnet/pull/1493

Closing this one.

PR #1491, opened 2021-11-13
Refactor client load balancing resolver

JamesNK:
The main change is `RefreshAsync` -> `Refresh` and add a protected `ResolveAsync` method. CancellationToken is based on dispose.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/DnsResolver.cs
+++ b/src/Grpc.Net.Client/Balancer/DnsResolver.cs
@@ -44,14 +44,9 @@ namespace Grpc.Net.Client.Balancer
 
         private readonly Uri _address;
         private readonly TimeSpan _refreshInterval;
-        private readonly ILogger<DnsResolver> _logger;
-        private readonly CancellationTokenSource _cts;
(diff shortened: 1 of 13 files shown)


src/Grpc.Net.Client/Balancer/DnsResolver.cs
```diff
@@ -165,33 +124,13 @@ protected override void Dispose(bool disposing)
...
-                        awaitRefresh = true;
-                    }
-                }
-
-                if (awaitRefresh)
-                {
-                    await _refreshTask.ConfigureAwait(false);
-                }
-            }
-            catch (OperationCanceledException)
```
captainsafia, line 192: Why is it safe to remove the catch for the `OperationCanceledException` here?

captainsafia (approved):

src/Grpc.Net.Client/Balancer/DnsResolver.cs
JamesNK: Refresh will never throw. However, there is a new try/catch in the base resolver that should probably ignore cancellation. I will update that code.

PR #1504, opened 2021-11-24
Log BalancerAddress instead of DnsEndPoint in load balancing

JamesNK:
Log `BalancerAddress` instead of `DnsEndPoint` because it has a more friendly ToString

Before:
> Successfully picked subchannel id '1' with address Undefined/127.0.0.1:50051.

After:
> Successfully picked subchannel id '1' with address 127.0.0.1:50051.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Internal/ConnectionManager.cs
+++ b/src/Grpc.Net.Client/Balancer/Internal/ConnectionManager.cs
@@ -302,13 +302,7 @@ namespace Grpc.Net.Client.Balancer.Internal
             }
         }
 
-        public async
-#if !NETSTANDARD2_0
-            ValueTask<(Subchannel Subchannel, BalancerAddress Address, Action<CompletionContext> OnComplete)>
-#else
-            Task<(Subchannel Subchannel, DnsEndPoint Address, Action<CompleteContext> OnComplete)>
-#endif
-            PickAsync(PickContext context, bool waitForReady, CancellationToken cancellationToken)
+        public async ValueTask<(Subchannel Subchannel, BalancerAddress Address, Action<CompletionContext> OnComplete)> PickAsync(PickContext context, bool waitForReady, CancellationToken cancellationToken)
         {
             SubchannelPicker? previousPicker = null;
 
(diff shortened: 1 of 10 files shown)


src/Grpc.Net.Client/Balancer/Internal/ConnectionManager.cs
```diff
@@ -302,13 +302,7 @@ public void UpdateState(BalancerState state)
...
         }
 
-        public async
-#if !NETSTANDARD2_0
-            ValueTask<(Subchannel Subchannel, BalancerAddress Address, Action<CompletionContext> OnComplete)>
-#else
-            Task<(Subchannel Subchannel, DnsEndPoint Address, Action<CompleteContext> OnComplete)>
-#endif
-            PickAsync(PickContext context, bool waitForReady, CancellationToken cancellationToken)
+        public async ValueTask<(Subchannel Subchannel, BalancerAddress Address, Action<CompletionContext> OnComplete)> PickAsync(PickContext context, bool waitForReady, CancellationToken cancellationToken)
```
captainsafia, line 305: Why is it OK to remove the if-def here?

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -49,15 +49,17 @@ namespace Grpc.Net.Client.Balancer.Internal
     /// </summary>
     internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDisposable
     {
+        internal record struct ActiveStream(BalancerAddress Address, Socket Socket, Stream? Stream);
```
captainsafia, line 52: Nice use of record structs here!

captainsafia (approved):
Looks good overall!

Left one comment about a removed if-def that looks awry.

src/Grpc.Net.Client/Balancer/Internal/ConnectionManager.cs
JamesNK: Originally I wrote this code to support netstandard2.0. It ended up being net5.0 or later. This is an unused ifdef that didn't get removed.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
JamesNK: Friendship ended with tuples.
Now record structs is my best friend.

PR #1527, opened 2021-12-16
Add version conditions for Microsoft.Extensions.Http and Microsoft.Extensions.Logging.Abstractions

ArminShoeibi:
This merge request adds some conditions in Grpc.Net.Client.csproj and Grpc.Net.ClientFactory.csproj.
Actually, this merge request solves this problem.

image

Thank you in advance.


JamesNK:
What does this fix? The older version is forward compatible, and a .NET 6 app can specify to use version 6 of that Microsoft.Extensions library if it wants.

ArminShoeibi:
> What does this fix? The older version is forward compatible, and a .NET 6 app can specify to use version 6 of that Microsoft.Extensions library if it wants.

Yes, you told me this before, but not everyone knows this.
They just install without attention, by the way with this merge request we don't need to install those packages explicitly.

JamesNK:
Ok. Usually referencing the lowest version is recommended. However, there are some advantages of bringing in a new version.

Let me talk with some other .NET people. I will see what the general approach is here.

JamesNK:
https://github.com/grpc/grpc-dotnet/pull/2021 updates the dependencies to .NET 6

PR #1530, opened 2021-12-19
Code clean up across solution

JamesNK:
Used editorconfig copied from dotnet and ran dotnet format.

Follow up work:
* Switch to file scoped namespaces
* Add editorconfig to repo perminantly


src/Grpc.Net.Client/Balancer/Internal/ConnectionManager.cs
```diff
@@ -468,7 +468,7 @@ void OnCancellation(object? s)
                     StateWatcher watcher = (StateWatcher)s!;
                     if (_stateWatchers.Remove(watcher))
                     {
-                        watcher.Tcs.SetCanceled();
+                        watcher.Tcs.SetCanceled(cancellationToken);
```
JamesNK, line 471: This introduces a closure. Need to refactor this method to take cancellation token as an argument. I think that is available in newer .nets

captainsafia (approved):
Did a light review but in `dotnet format` we trust.

PR #1545, opened 2021-12-29
Use IHttpActivityFeature to get host activity

JamesNK:
`IHttpActivityFeature` added in .NET 6 allows the host activity to be fetched directly.

* Simpler logic
* Slight performance improvement compared to `Activity.Current` (async local) + comparing activity name

### DIFF
--- a/testassets/FunctionalTestsWebsite/Startup.cs
+++ b/testassets/FunctionalTestsWebsite/Startup.cs
@@ -136,7 +136,7 @@ namespace FunctionalTestsWebsite
 
             app.Use(async (context, next) =>
             {
-                // Allow a call to specify a activity tags are returned as trailers.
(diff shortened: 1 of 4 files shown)


testassets/FunctionalTestsWebsite/Startup.cs
```diff
@@ -132,6 +134,33 @@ public void Configure(IApplicationBuilder app)
                 await next();
             });
 
+            app.Use(async (context, next) =>
+            {
+                // Allow a call to specify a activity tags are returned as trailers.
```
captainsafia, line 139:
```suggestion
                // Allow a call to specify activity tags are returned as trailers.
```

captainsafia (approved):

Issue #1108, opened 2020-11-09
Add support for deflate compression


JamesNK:
.NET finally added correct support for deflate - https://github.com/dotnet/runtime/pull/42717

Possibly add deflate compression to gRPC client and server.

JamesNK:
Requires .NET 6

PR #1548, opened 2021-12-30
Add deflate compression to .NET 6

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1108

### DIFF
--- a/src/Grpc.Net.Client/Internal/GrpcCallSerializationContext.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCallSerializationContext.cs
@@ -295,15 +295,6 @@ namespace Grpc.Net.Client.Internal
             return output.GetBuffer().AsSpan(0, (int)output.Length);
         }
 
-        private sealed class NonDisposableMemoryStream : MemoryStream
-        {
-            protected override void Dispose(bool disposing)
-            {
-                // Ignore dispose from wrapping compression stream.
-                // If MemoryStream is disposed then Length isn't available.
-            }
-        }
-
         public void Advance(int count)
         {
             if (_buffer != null)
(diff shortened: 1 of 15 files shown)


src/Grpc.Net.Common/Compression/DeflateCompressionProvider.cs
```diff
@@ -0,0 +1,78 @@
...
+#if NET6_0_OR_GREATER
+using System.IO;
+using System.IO.Compression;
+
+namespace Grpc.Net.Compression
+{
+    /// <summary>
+    /// Deflate compression provider.
+    /// </summary>
+    public class DeflateCompressionProvider : ICompressionProvider
```
gfoidl, line 28:
```suggestion
    public sealed class DeflateCompressionProvider : ICompressionProvider
```
With more and better support in the JIT for devirtualization `sealed` becomes important (again).

And "un-sealing" a type won't be a breaking change. The other way around, to seal an unsealed type, would be a breaking change.

src/Grpc.Net.Client/Internal/GrpcCallSerializationContext.cs
```diff
@@ -295,6 +295,15 @@ private ReadOnlySpan<byte> CompressMessage(ReadOnlySpan<byte> messageData)
             return output.GetBuffer().AsSpan(0, (int)output.Length);
         }
 
+        private sealed class NonDisposableMemoryStream : MemoryStream
```
gfoidl, line 298: This is just a "mini-type", but above on the server-side the same type is defined. Should this two definitions be merged to common place?

src/Grpc.Net.Common/Compression/DeflateCompressionProvider.cs
JamesNK: Good point.

There is one other compression provider and it isn't sealed. I'll make it sealed as well. It's breaking, but no one has a reason to inherit from it.

captainsafia (approved):

PR #1554, opened 2022-01-05
Fix load balancing subchannel compare and dispose error

JamesNK:
Regression in 2.42.0 related to not correctly comparing subchannel addresses. The regression results in subchannels being recreated each time a resolver returns new addresses. Correct behavior is to recognize that a subchannel with an address already exists and continue to use it.

Note that most changes are related to logging. I noticed this when updating container example to use latest version and added more logging to help follow logic.

Should be backported to the 2.42.0 branch before the final version is released.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
+++ b/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
@@ -93,7 +93,10 @@ namespace Grpc.Net.Client.Balancer.Internal
                 _initialSocket = null;
                 _initialSocketAddress = null;
                 _lastEndPointIndex = 0;
-                _socketConnectedTimer.Change(TimeSpan.Zero, TimeSpan.Zero);
+                if (!_disposed)
+                {
+                    _socketConnectedTimer.Change(TimeSpan.Zero, TimeSpan.Zero);
+                }
                 _currentAddress = null;
             }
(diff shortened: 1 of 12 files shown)


src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -93,7 +93,10 @@ public void Disconnect()
                 _initialSocket = null;
                 _initialSocketAddress = null;
                 _lastEndPointIndex = 0;
-                _socketConnectedTimer.Change(TimeSpan.Zero, TimeSpan.Zero);
+                if (!_disposed)
+                {
+                    _socketConnectedTimer.Change(TimeSpan.Zero, TimeSpan.Zero);
+                }
```
JamesNK, line 99: Guard against error if the transport is disconnected after it disposed.

src/Grpc.Net.Client/Balancer/SubchannelsLoadBalancer.cs
```diff
@@ -85,7 +85,7 @@ private void ResolverError(Status status)
             for (var i = 0; i < addressSubchannels.Count; i++)
             {
                 var s = addressSubchannels[i];
-                if (Equals(s.Address, address))
+                if (Equals(s.Address.EndPoint, address.EndPoint))
```
JamesNK, line 88: Compare the addresses endpoints to get address reuse.

captainsafia (approved):

PR #1559, opened 2022-01-12
Fix #1558

### LINKED ISSUE #1558: #1465 break change: not support OpenTelemetry.Instrumentation.GrpcNetClient
## Solution

Move `ActivityStartData` and `ActivityStopData` from [GrpcCall.cs](https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.Net.Client/Internal/GrpcCall.cs) to [GrpcCall.NonGeneric.cs](https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.Net.Client/Internal/GrpcCall.NonGeneric.cs)

## Reason
[GrpcClientDiagnosticListener](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.GrpcNetClient/Implementation/GrpcClientDiagnosticListener.cs#L62) use [PropertyFetcher](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/DiagnosticSourceInstrumentation/PropertyFetcher.cs#L132) read `Request` property. `PropertyFetcher` use `TypedPropertyFetch` to improve performance, but in different methods, `ActivityStartData` and `ActivityStopData` are not of the same type. so, `obj is TDeclaredObject o` is false after the first call.

## Sample
``` C#
var services = new ServiceCollection();

services.AddGrpcClient<Greeter.GreeterClient>(options => options.Address = new Uri("http://localhost"))
    .ConfigurePrimaryHttpMessageHandler(_ => new TestHandler());

using var _ = Sdk.CreateTracerProviderBuilder()
    .AddGrpcClientInstrumentation(options => options.SuppressDownstreamInstrumentation = true)
    .SetSampler(new AlwaysOnSampler())
    .Build();
await using var root = services.BuildServiceProvider();
await using var scope = root.CreateAsyncScope();

var client = scope.ServiceProvider.GetRequiredService<Greeter.GreeterClient>();

try
{
    await client.SayHelloAsync(new HelloRequest());
}
catch (RpcException ex) when (ex.Status.StatusCode == StatusCode.OK) { }

try
{
    await client.SayHello2Async(new HelloRequest2());
}
catch (RpcException ex) when (ex.Status.StatusCode == StatusCode.OK) { }

class TestHandler : DelegatingHandler
{
    protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
    {
        if (request.Headers.Contains("traceparent")) //Should never true
            throw new RpcException(Status.DefaultSuccess);

        throw new UnauthorizedAccessException();
    }
}
```

pengweiqhca:


JamesNK:
Thanks. Is there a way to unit test this? A unit test will prevent another regression.

JamesNK (approved):

JamesNK:
I updated the unit test a bit.

Thanks for raising this bug and the PR!

PR #1579, opened 2022-01-20
Fix running microbenchmarks on .NET 6

JamesNK:


captainsafia (approved):
Seems sensible. Can you clarify what the break that this is fixing was?

Seems like running a 3.1 app no longer works on the latest benchmarking package?

JamesNK:
Need to update benchmarkdotnet to support .NET versions that don't have core in the name.
Need to update the tool-chain to .NET 6 because that is what the project targets.

Disclaimer: I'm not a benchmarkdotnet expert. I try things until it works 😄

PR #1580, opened 2022-01-20
Fix running HTTP/3 on benchmark environment

JamesNK:
A couple of ifdef that should be .NET 6 or greater.


captainsafia (approved):
Approving! Seems like there's no other ifdefs like these lingering in the codebase.

PR #1592, opened 2022-01-28
Update version to 2.43.0-pre1 (on v2.43.x)

jtattermusch:


jtattermusch:
I'm going to proceed since we confirmed publishing pre1 offline.

PR #1615, opened 2022-02-19
Update interop tests to use modern package lock file

JamesNK:


src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -80,33 +80,35 @@ protected override string PeerCore
...
-                        _peer = "unknown"; // Match Grpc.Core
-                    }
+                    _peer = BuildPeer();
                 }
 
                 return _peer;
             }
         }
 
+        private string BuildPeer()
```
JamesNK, line 90: Mirror update in GrpcHttpApi. Checked in by accident but IMO not worth a separate PR

testassets/InteropTestsGrpcWebWebsite/Tests/package.json
```diff
@@ -6,8 +6,8 @@
     "test": "jest test"
   },
   "dependencies": {
-    "puppeteer": "^11.0.0",
-    "jest": "^27.3.1",
-    "jest-puppeteer": "^6.0.0"
+    "puppeteer": "^13.3.2",
```
captainsafia, line 9: Not super relevant but do we want to pin these to the patch version for more consistency in the build?

Not sure if this repo is set up with dependabot which would defeat the purpose but just a thought

JamesNK: Doesn't the lock file pin dependencies for the build?

I'm not a node expert but the ^ syntax is used in package.json files in aspnetcore repo.

testassets/InteropTestsGrpcWebWebsite/Tests/custom-environment.js
```diff
@@ -1,6 +1,9 @@
-// custom-environment.js
+// For some reason the expect-puppeteer require is returning a module rather than the function.
+// Helper function unwraps the module and returns the inner function.
+function _interopDefault(ex) { return (ex && (typeof ex === 'object') && 'default' in ex) ? ex['default'] : ex; }
+
 const PuppeteerEnvironment = require('jest-environment-puppeteer');
-const expect = require('expect-puppeteer');
+const expect = _interopDefault(require('expect-puppeteer'));
```
captainsafia, line 6:
```suggestion
import expect from 'expect-puppeteer'
```

Can we use module imports here instead?

JamesNK: I tried that but got an error along the lines of "can't import outside of a module"

JamesNK: Hmm, I didn't do exactly what you suggested: [`554f1bc` (#1615)](https://github.com/grpc/grpc-dotnet/pull/1615/commits/554f1bc48d494e95c8c42f10a2cdc28fca03453c)

I'll give it a try

JamesNK: To get import to work the package type has to be module which means lots of other changes. The current fix probably isn't better but it is simpler.

testassets/InteropTestsGrpcWebWebsite/Tests/package.json
captainsafia: Yes, it does! Missed that there were lock files in here. 😅

captainsafia (approved):

Issue #1422, opened 2021-09-22
Add IAsyncStreamWriter<T>.WriteAsync(T, CancellationToken) overload


tommyja:
### Problem

There is no way for the server to discover that a client is not consuming a stream.
For a long lived stream, this can lead to WriteAsync eventually blocking forever on the server side.

### Possible solution

Add a timeout option to IAsyncStreamWriter.WriteAsync 
If the message is not sent to the client within the timeout, terminate the stream or call a provided Action

### Workarounds considered

1. Start a timer when WriteAsync is called. If the timer triggers before WriteAsync  returns, terminate the stream. The problem with this is that WriteAsync does not block before the out buffer is full. This can take a long time for low throughput streams.

2. Using bidirectional streams with sequence number in the proto . Terminate stream if ack for a given sequence number is not received within the timeout. This works but is clumsy.

davidfowl:
@jtattermusch WriteAsync needs a CancellationToken. That's a breaking change though so we'll need to figure something out.

tommyja:
@davidfowl a cancellation token would help in many cases, but it will not solve the case with low volume streams as it can take a long time to fill the out buffer,

JamesNK:
The only way I could imagine this working is if there was a way to monitor flow control's window size - https://medium.com/coderscorner/http-2-flow-control-77e54f7fd518

As the server writes more data to the client, the available buffer space on the client decreases. When the client reads the incoming messages then it reports back that it has buffer space.

As far as I know, there is no API to notifications about flow control size changing in Kestrel. Correct me if I'm wrong @davidfowl.

Having the client report receipt of messages using a bidirectional stream will certainly work.

jtattermusch:
What exactly is the kind of scenarios you'd like to be able to detect?
- if connection is broken (or other fatal problem happens), the entire RPC is going to be cancelled and all the methods like WriteAsync are going to return immediately (which AFAIK is the current behavior).
- if the client is just reading slowly, it's basically impossible to detect whether the connection is just slow, client is processing data slowly or if client is stuck, so I don't know what value there is in "discovering that client is not consuming a stream". If you need to guarantee that the client has seen the data you've sent, on time, you need to use the approach described in workaround 2 ("explicit confirmations").

[...]

PR #1645, opened 2022-03-21
Support WriteAsync cancellation token

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1422

Turns out it was pretty complicated, especially retries and hedging. PR adds support in all places:

* WriteAsync cancellation on server
* WriteAsync cancellation on client
* WriteAsync cancellation on client for retries and hedging

Also fixed a bug when hedging and writing a large message to a client stream which causes the call to be committed because it exceeded buffer size. I found this while testing.

Also fixed a bug where the server would silently continue if it was aborted while writing a message. Now throws OperationCanceledException.

~Note: Contains some changes from https://github.com/grpc/grpc-dotnet/pull/1644 because Grpc.Core.Api needed to be updated.~

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
@@ -43,7 +43,37 @@ namespace Grpc.AspNetCore.Server.Internal
             return new Status(StatusCode.Unimplemented, $"Unsupported grpc-encoding value '{unsupportedEncoding}'. Supported encodings: {string.Join(", ", supportedEncodings)}");
         }
 
-        public static async Task WriteMessageAsync<TResponse>(this PipeWriter pipeWriter, TResponse response, HttpContextServerCallContext serverCallContext, Action<TResponse, SerializationContext> serializer, bool canFlush)
+        public static async Task WriteSingleMessageAsync<TResponse>(this PipeWriter pipeWriter, TResponse response, HttpContextServerCallContext serverCallContext, Action<TResponse, SerializationContext> serializer)
+            where TResponse : class
+        {
+            var logger = serverCallContext.Logger;
+            try
+            {
+                // Must call StartAsync before the first pipeWriter.GetSpan() in WriteHeader
+                var httpResponse = serverCallContext.HttpContext.Response;
+                if (!httpResponse.HasStarted)
+                {
+                    await httpResponse.StartAsync();
+                }
+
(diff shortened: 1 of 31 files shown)


src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -68,7 +68,14 @@ public static async Task WriteMessageAsync<TResponse>(this PipeWriter pipeWriter
...
                 if (flush)
                 {
-                    await pipeWriter.FlushAsync();
+                    var flushResult = await pipeWriter.FlushAsync();
+
+                    // Workaround bug where FlushAsync doesn't return IsCanceled = true on request abort.
+                    // https://github.com/dotnet/aspnetcore/issues/40788
+                    if (!flushResult.IsCompleted && serverCallContext.CancellationToken.IsCancellationRequested)
+                    {
+                        throw new OperationCanceledException();
```
captainsafia, line 77: Does this exception end up surfacing to the user? If so, should we add a message?

test/FunctionalTests/Client/StreamingTests.cs
```diff
@@ -35,6 +36,17 @@ namespace Grpc.AspNetCore.FunctionalTests.Client
     [TestFixture]
     public class StreamingTests : FunctionalTestBase
     {
+        private const int BigMessageSize = 1024 * 1024 * 5;
```
captainsafia, line 39: Where does this number come from?

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
JamesNK: Yes. Added message.

test/FunctionalTests/Client/StreamingTests.cs
JamesNK: Its a message big enough to hit flow control if the peer doesn't immediately read it.

Added comment.

JamesNK:
The longer I look at this PR, the more bugs I find in other places. New bug fixes:

* Client streaming hedge calls wait for WriteAsync to succeed at least once before continuing.
* Buffered hedge calls are explicitly flushed.
* Fix race between server FlushAsync exiting and request cancellation token reporting cancellation.

I want to get this merged before it grows any more.

captainsafia (approved):

PR #1651, opened 2022-03-29
Move Grpc.Core.Api nuget from grpc/grpc

jtattermusch:
Add Grpc.Core.Api and relevant tests from grpc/grpc to grpc-dotnet.

- copy the src/csharp/Grpc.Core.Api subtree from grpc/grpc
- add the tests relevant to Grpc.Core.Api from src/csharp/Grpc.Core.Tests as Grpc.Core.Api.Tests
- make necessary adjustments to fit the csproj layout in grpc-dotnet repo
- fix analyzer warnings.
- makes Grpc.Core.Api target net462 instead of net45 (net45 is past end of life). Based on my experiments, this still allows to reference newer Grpc.Core.Api versions from projects that use older versions of Grpc.Core (for which there will likely be no more major version releases).  (the net462 version of Grpc.core.Api.dll assembly can still be used by an older Grpc.Core.dll assembly that targets net45 - as long as the project including both Grpc.Core.Api and Grpc.Core uses net462+, which is a fair limitation, since net45 is past end of support so there in practice there should really be no systems that don't support at least .NET Framework 4.6.2)
- best reviewed commit-by-commit


src/Grpc.Core.Api/VersionInfo.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace Grpc.Core
+{
+    /// <summary>
+    /// Provides info about current version of gRPC.
+    /// See https://codingforsmarties.wordpress.com/2016/01/21/how-to-version-assemblies-destined-for-nuget/
+    /// for rationale about assembly versioning.
+    /// </summary>
+    public static class VersionInfo
+    {
+        // TODO(jtattermusch): how to propagate versions from https://github.com/grpc/grpc-dotnet/blob/f238855fcb77cec01e87eb9cc08680c3420c132f/Directory.Build.props#L8-L10version into
```
jtattermusch, line 28: Still need to figure out how to turn .csproj properties into  string constants.
(in grpc/grpc we simply used codegen to generate contents of VersionInfo.cs)

src/Grpc.Core.Api/.gitignore
```diff
@@ -0,0 +1,2 @@
+bin
```
JamesNK, line 1: root .gitignore should have these already

src/Grpc.Core.Api/AsyncAuthInterceptor.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// <param name="metadata">Metadata to populate with entries that will be added to outgoing call's headers.</param>
+    /// <returns></returns>
+    public delegate Task AsyncAuthInterceptor(AuthInterceptorContext context, Metadata metadata);
+
+    /// <summary>
+    /// Context for an RPC being intercepted by <see cref="AsyncAuthInterceptor"/>.
+    /// </summary>
+    public class AuthInterceptorContext
+    {
+        readonly string serviceUrl;
```
JamesNK, line 41: What do you think of eventually following grpc-dotnet standards, e.g. underscore on private variables, no this, var, etc

src/Grpc.Core.Api/ContextPropagationToken.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    /// <summary>
+    /// Token for propagating context of server side handlers to child calls.
+    /// In situations when a backend is making calls to another backend,
+    /// it makes sense to propagate properties like deadline and cancellation 
+    /// token of the server call to the child call.
+    /// Underlying gRPC implementation may provide other "opaque" contexts (like tracing context) that
+    /// are not explicitly accesible via the public C# API, but this token still allows propagating them.
+    /// </summary>
+    public abstract class ContextPropagationToken
```
JamesNK, line 29: grpc-dotnet doesn't use ContextPropagationToken and friends. Perhaps obsolete in the future?

src/Grpc.Core.Api/Internal/CodeAnalysisAttributes.cs
```diff
@@ -0,0 +1,151 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+// Content of this file is copied with permission from:
+// https://github.com/dotnet/runtime/tree/e2e43f44f1032780fa7c2bb965153c9da615c3d0/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis
+// These types are intented to be added as internalized source to libraries and apps that want to
+// use them without requiring a dependency on .NET 5.
```
JamesNK, line 22: There might already be an internal version of this file in the repo. Could bring it in via a link and remove this

src/Grpc.Core.Api/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,46 @@
...
+[assembly: InternalsVisibleTo("Grpc.IntegrationTesting,PublicKey=" +
+    "00240000048000009400000006020000002400005253413100040000010001002f5797a92c6fcde81bd4098f43" +
+    "0442bb8e12768722de0b0cb1b15e955b32a11352740ee59f2c94c48edc8e177d1052536b8ac651bce11ce5da3a" +
+    "27fc95aff3dc604a6971417453f9483c7b5e836756d5b271bf8f2403fe186e31956148c03d804487cf642f8cc0" +
+    "71394ee9672dfe5b55ea0f95dfd5a7f77d22c962ccf51320d3")]
+[assembly: InternalsVisibleTo("Grpc.Microbenchmarks,PublicKey=" +
+    "00240000048000009400000006020000002400005253413100040000010001002f5797a92c6fcde81bd4098f43" +
+    "0442bb8e12768722de0b0cb1b15e955b32a11352740ee59f2c94c48edc8e177d1052536b8ac651bce11ce5da3a" +
+    "27fc95aff3dc604a6971417453f9483c7b5e836756d5b271bf8f2403fe186e31956148c03d804487cf642f8cc0" +
+    "71394ee9672dfe5b55ea0f95dfd5a7f77d22c962ccf51320d3")]
```
JamesNK, line 46: Can be removed/replaced

src/Grpc.Core.Api/README.md
```diff
@@ -0,0 +1,3 @@
+# Grpc.Core.Api
+
+`Grpc.Core.Api` is the shared API package for gRPC C# and gRPC for .NET implementations of gRPC.
```
JamesNK, line 3: Packages in grpc-dotnet have README.md embedded in nupkg to display on nuget. Should add an issue with the list of new projects without embedded readme.

src/Grpc.Core.Api/Status.cs
```diff
@@ -0,0 +1,95 @@
...
+#endregion
+
+using System;
+
+namespace Grpc.Core
+{
+    /// <summary>
+    /// Represents RPC result, which consists of <see cref="StatusCode"/> and an optional detail string.
+    /// </summary>
+    public struct Status
```
JamesNK, line 24: Should make this readonly

src/Grpc.Core.Api/Utils/EncodingExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+#endregion
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace Grpc.Core.Api.Utils
+{
+
+    internal static class EncodingExtensions
```
JamesNK, line 24: What uses this?

JamesNK:
There are tests in Grpc.Core.Tests that relate to Grpc.Core.Api functionality. They should be extracted to a new test project.

src/Grpc.Core.Api/AsyncAuthInterceptor.cs
jtattermusch: It is something we can consider, but I think it's low priority. I understand that we'd want all the code to follow the same code-style, but I'd like us to refrain from making unnecessary changes / refactorings until the point when things settle down a bit (e.g. once we've done one or two successful releases of the new location and we have some verification that there aren't any unexpected problems).

src/Grpc.Core.Api/ContextPropagationToken.cs
jtattermusch: Perhaps but only after Grpc.Core is dead, which may still take considerable amount of time.

src/Grpc.Core.Api/Properties/AssemblyInfo.cs
jtattermusch: We can only remove once we verify that Grpc.Core.Api assembly built using the grpc-dotnet repo still works with the tests from the grpc/grpc (since those tests aren't going to be deleted, we still need to be able to run them, even though they'd live on a "frozen" branch.)

src/Grpc.Core.Api/README.md
jtattermusch: the .csproj already has `<PackageReadmeFile>README.md</PackageReadmeFile>` isn't that sufficient?

src/Grpc.Core.Api/VersionInfo.cs
jtattermusch: @JamesNK  any ideas how to do this?

JamesNK: No. I know how to easily embed properties as assembly attributes, which can then be fetched at runtime. But these are const values so that won't work.

src/Grpc.Core.Api/README.md
JamesNK: Also need https://github.com/grpc/grpc-dotnet/blob/f238855fcb77cec01e87eb9cc08680c3420c132f/src/Grpc.AspNetCore/Grpc.AspNetCore.csproj#L17

src/Grpc.Core.Api/VersionInfo.cs
JamesNK: Could write a source generator? https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/source-generators-overview

src/Grpc.Core.Api/Internal/CodeAnalysisAttributes.cs
jtattermusch: I wasn't able to find preexisting copy inside the grpc-dotnet repo.

src/Grpc.Core.Api/.gitignore
jtattermusch: removed.

src/Grpc.Core.Api/ContextPropagationToken.cs
jtattermusch: added a TODO.

src/Grpc.Core.Api/Status.cs
jtattermusch: added a TODO since I only want to make optional changes after everything is setup and works well.

src/Grpc.Core.Api/Utils/EncodingExtensions.cs
jtattermusch: added a TODO to revisit. This is probably used by Grpc.Core, so we'd have to make changes there first to avoid breaking it.

src/Grpc.Core.Api/README.md
jtattermusch: done.

jtattermusch:
> There are tests in Grpc.Core.Tests that relate to Grpc.Core.Api functionality. They should be extracted to a new test project.

Added Grpc.Core.Api.Tests with the tests from Grpc.Core.Tests that seemed relevant.

jtattermusch:
One more question: It seems feasible to add the original src/csharp/Grpc.Core.Api git subtree along with filtered commit history by using the `git subtree split` command. Do you think it would be worthwhile to import the code along with change history (the commit messages would have a `--annotate` prefix that would make it clear the the history is from a different repo) or is simply pointing to the original codebase somewhere in a readme good enough?

jtattermusch:
> One more question: It seems feasible to add the original src/csharp/Grpc.Core.Api git subtree along with filtered commit history by using the `git subtree split` command. Do you think it would be worthwhile to import the code along with change history (the commit messages would have a `--annotate` prefix that would make it clear the the history is from a different repo) or is simply pointing to the original codebase somewhere in a readme good enough?

Discussed and we decided to simply do a codedump and link to the original codebase (at a clearly documented revision) for simplicity.

jtattermusch:
@JamesNK  can you please do another pass and let me know if there's any other major concerns (besides the question of regenerating the constants in VersionInfo.cs) that would prevent this PR from accepting?
We still won't merge this, but I'd like to make sure that with some clean up this will be mergeable once we need it to.

test/Grpc.Core.Api.Tests/Grpc.Core.Api.Tests.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net45;netcoreapp3.1</TargetFrameworks>
```
JamesNK, line 3: net45 -> net462

I think eventually these new projects should test with the same frameworks as existing test projects (e.g. + net5.0 and net6.0) but can leave it unchanged for now.

src/Grpc.Core.Api/Grpc.Core.Api.csproj
```diff
@@ -0,0 +1,39 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Description>gRPC C# Surface API</Description>
+    <PackageTags>gRPC RPC HTTP/2</PackageTags>
+
+    <IsGrpcPublishedPackage>true</IsGrpcPublishedPackage>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <TargetFrameworks>net45;netstandard1.5;netstandard2.0;netstandard2.1</TargetFrameworks>
```
JamesNK, line 8: net45 isn't supported anymore. See https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-framework

Should change net45 to net462 for various Grpc.Core projects.

jtattermusch: ok, that seems doable. But there should also be a some follow-up cleanup after switching to net462 (`#if` defs and also there is some compatibility code that is no longer needed on net462)

jtattermusch:
@JamesNK  this is ready for final pass of review (even when approved, I still won't merge just yet, I'd like to synchronize removing stuff from grpc/grpc with adding the corresponding stuff to grpc-dotnet).

jtattermusch:
I'm going to merge this to allow for easier review of https://github.com/grpc/grpc-dotnet/pull/1652 and https://github.com/grpc/grpc-dotnet/pull/1708.

- the 2.47.x branch has already been cut on master, and the version 2.47 and above of Grpc.Core.Api will be released as part of the grpc-dotnet release. v2.46.x will be the last version that's going to be released from the grpc/grpc repository.

PR #1652, opened 2022-03-29
Move Grpc.Auth nuget from grpc/grpc

jtattermusch:
~~Based on https://github.com/grpc/grpc-dotnet/pull/1651~~ (rebased)

Move Grpc.Auth to the grpc-dotnet repository.

### DIFF
--- a/src/Grpc.Auth/.gitignore
+++ b/src/Grpc.Auth/.gitignore
@@ -1,3 +1,0 @@
-bin
-obj
-*.nupkg


jtattermusch:
@JamesNK this is ready review (even when approved, I still won't merge just yet, I'd like to synchronize removing stuff from grpc/grpc with adding the corresponding stuff to grpc-dotnet).

JamesNK (approved):
Looks good. I'd like to double-check after Grpc.Core.Api is merged so I can see the Grpc.Auth changes in isolation.

jtattermusch:
> Looks good. I'd like to double-check after Grpc.Core.Api is merged so I can see the Grpc.Auth changes in isolation.

Rebased. PTAL.

src/Grpc.Auth/.gitignore
```diff
@@ -0,0 +1,3 @@
+bin
+obj
+*.nupkg
```
JamesNK, line 3: Is this necessary? There is a .gitignore in the root folder

(I think this came up for the Grpc.Core.Api review)

src/Grpc.Auth/GoogleGrpcCredentials.cs
```diff
@@ -0,0 +1,81 @@
...
+            var googleCredential = await GoogleCredential.GetApplicationDefaultAsync().ConfigureAwait(false);
+            return googleCredential.ToChannelCredentials();
+        }
+
+        /// <summary>
+        /// Creates an instance of <see cref="CallCredentials"/> that will use given access token to authenticate
+        /// with a gRPC service.
+        /// </summary>
+        /// <param name="accessToken">OAuth2 access token.</param>
+        /// /// <returns>The <c>MetadataCredentials</c> instance.</returns>
```
JamesNK, line 52:
```suggestion
        /// <returns>The <c>CallCredentials</c> instance.</returns>
```

JamesNK (commented):
Minor feedback

jtattermusch:
The interop test failure is because we were grpc-dotnet interop against Grpc.Core, which has recently been removed from the master branch on grpc/grpc (see https://github.com/grpc/grpc/pull/29225).
I will fix as a followup.

src/Grpc.Auth/.gitignore
jtattermusch: Removed.

Issue #1646, opened 2022-03-22
Changing Subchannel connect backoff timeout


vadimi:
Hello, I'm using gRPC 2.44.0 client and I noticed that when grpc service is unavailable for some time it takes about 2 minutes for gRPC client to reconnect. If I understand correctly this backoff timeout is implemented on the Subchannel level:

https://github.com/grpc/grpc-dotnet/blob/d0272085d9fb42329746f04ce42b244b46e61915/src/Grpc.Net.Client/Balancer/Subchannel.cs#L272

It doesn't look like it's configurable. Is there a plan to make it configurable? It's configurable in gRPC core `new ChannelOption("grpc.max_reconnect_backoff_ms", _settings.MaxReconnectBackoffMs)` and in gRPC Go as well:
https://github.com/grpc/grpc-go/blob/6c3ccbe89a5a73a65f339b1e5d3586c2782a830c/dialoptions.go#L216

jtattermusch:
We could add an equivalent of these channel arguments:
https://github.com/grpc/grpc/blob/8a406c72df0cbf8251d4d7b56714cbcda497da86/include/grpc/impl/codegen/grpc_types.h#L258

Also see https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md

vadimi:
that would definitely be very helpful

PR #1668, opened 2022-04-05
Reconnect backoff configuration

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1646

Note that backoff behavior matches spec at https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Internal/ExponentialBackoffPolicy.cs
+++ b/src/Grpc.Net.Client/Balancer/Internal/ExponentialBackoffPolicy.cs
@@ -1,0 +1,85 @@
+﻿#region Copyright notice and license
+
+// Copyright 2019 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


src/Grpc.Net.Client/Balancer/Subchannel.cs
```diff
@@ -261,15 +266,12 @@ private async Task ConnectTransportAsync()
                         SubchannelLog.ConnectBackoffInterrupted(_logger, Id);
 
                         // Delay interrupt was triggered. Reset back-off.
-                        backoffMs = InitialBackOffMs;
+                        backoffPolicy = _manager.BackoffPolicyFactory.Create();
```
captainsafia, line 269: What's the purpose of recreating the backoff policy here?

captainsafia: Ah wait, I think I see what this is for. This will reset the backoff back to the initial values per the requirements in the spec.

captainsafia (approved):

src/Grpc.Net.Client/Balancer/Internal/ExponentialBackoffPolicy.cs
```diff
@@ -0,0 +1,85 @@
...
+        private readonly GrpcChannel _channel;
+
+        public ExponentialBackoffPolicyFactory(GrpcChannel channel)
+        {
+            _channel = channel;
+        }
+
+        public IBackoffPolicy Create()
+        {
+            // Limit ticks to Int32.Value. Task.Delay can't use larger values,
```
JamesNK, line 76:
```suggestion
            // Limit ticks to Int32.MaxValue. Task.Delay can't use larger values,
```

Issue #1654, opened 2022-03-30
Error starting gRPC call. A task was canceled.


Hovo563:
When I am doing grpc call, after some time this exception throws "Error starting gRPC call. A task was canceled."

.net 6 

Exception 

{
  "@timestamp": "2022-03-30T12:37:38.5892808+04:00",
  "level": "Error",
  "messageTemplate": "Error starting gRPC call.",
  "message": "Error starting gRPC call.",
  "exceptions": [
    {
      "Depth": 0,
      "ClassName": "System.Threading.Tasks.TaskCanceledException",
      "Message": "A task was canceled.",
      "Source": "System.Private.CoreLib",
      "StackTraceString": "   at Grpc.Net.Client.Balancer.Internal.ConnectionManager.GetNextPickerAsync(CancellationToken cancellationToken)\n   at Grpc.Net.Client.Balancer.Internal.ConnectionManager.PickAsync(PickContext context, Boolean waitForReady, CancellationToken cancellationToken)\n   at Grpc.Net.Client.Balancer.Internal.BalancerHttpHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)\n   at Grpc.Net.Client.Internal.GrpcCall`2.RunCall(HttpRequestMessage request, Nullable`1 timeout)",
      "RemoteStackTraceString": null,
      "RemoteStackIndex": 0,
      "HResult": -2146233029,
      "HelpURL": null
    }
  ],
  "fields": {
    "EventId": {
      "Id": 6,
      "Name": "ErrorStartingCall"
    },
    "SourceContext": "Grpc.Net.Client.Internal.GrpcCall",
    "GrpcMethodType": "Unary",
    "GrpcUri": "/XXXXX/Get",
  }
}

JamesNK:
Unfortunately, the error message by itself isn't enough to figure out what's going on.

Could you provide a simplified reproduction of your app that someone can debug?

Alternatively, could you provide logs for the client? See instructions here: https://docs.microsoft.com/en-us/aspnet/core/grpc/diagnostics?view=aspnetcore-6.0#grpc-client-logging

Hovo563:
here is my client registration. 

  var defaultMethodConfig = new MethodConfig
            {
                Names = { MethodName.Default },
                RetryPolicy = new RetryPolicy
                {
                    MaxAttempts = 5,
                    InitialBackoff = TimeSpan.FromSeconds(1),
[...]

PR #1688, opened 2022-04-15
Fix call cancellation impacting other calls

JamesNK:
There is a bug in the new address picker to do with cancellation.

Canceling a call also cancels other queued-up calls. The fix is to not use a share `TaskCompletionSource` between callers for cancellation. Instead, WaitAsync has its own TCS. WaitAsync is new in .NET 6 so a compatible version is added to older platforms.

This change makes picking much simpler. It hopeful addresses `An attempt was made to transition a task to a final state when it had already completed.` from https://github.com/grpc/grpc-dotnet/issues/1654

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Internal/ConnectionManager.cs
+++ b/src/Grpc.Net.Client/Balancer/Internal/ConnectionManager.cs
@@ -36,7 +36,6 @@ namespace Grpc.Net.Client.Balancer.Internal
     {
         private static readonly ServiceConfig DefaultServiceConfig = new ServiceConfig();
 
-        private readonly SemaphoreSlim _nextPickerLock;
         private readonly object _lock;
         internal readonly Resolver _resolver;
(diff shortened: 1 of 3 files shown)


src/Grpc.Net.Client/Balancer/Internal/ConnectionManager.cs
```diff
@@ -365,70 +365,21 @@ public async ValueTask<(Subchannel Subchannel, BalancerAddress Address, ISubchan
...
             lock (_lock)
             {
                 if (_picker != null && _picker != currentPicker)
                 {
-#if !NETSTANDARD2_0
-                    return new ValueTask<SubchannelPicker>(_picker);
-#else
-                    return Task.FromResult<SubchannelPicker>(_picker);
-#endif
+                    return _pickerTask!;
```
captainsafia, line 374:
```suggestion
                    Debug.Assert(_pickerTask != null);
                    return _pickerTask;
```

captainsafia (approved):

PR #1693, opened 2022-04-17
Update README.md: small edit to indicate .NET 3.0 or later.

wadepickett:
The text referring to .NET Core 3.0 made it sound like that was the only version this was available for and that gRPC for .NET has not been updated.  I simply added "or later" to make it a little clearer.   :)


wadepickett:
@JamesNK, minor edit here.

JamesNK (approved):

JamesNK:
Thanks!

PR #1696, opened 2022-04-19
Place XML comment hrefs in see elements

JamesNK:
Make links clickable:

image


captainsafia (approved):
TIL! I'll have to start using this more often in code.

PR #1700, opened 2022-04-21
Upgrade System.Commandline and add command line tests

JamesNK:
Updates System.Commandline to the latest version. It's a fairly big change in API and usage. I'm guessing it's done in the name of trimming. Annoying though.

I added unit tests for invoking via command line to ensure nothing has broken.

I also made it so that the project option can just be a path, and implicitly uses a single csproj in the path. This matches `dotnet`

### DIFF
--- a/perf/benchmarkapps/GrpcClient/ClientOptions.cs
+++ b/perf/benchmarkapps/GrpcClient/ClientOptions.cs
@@ -22,7 +22,7 @@ namespace GrpcClient
 {
     public class ClientOptions
     {
-        public string? Url { get; set; }
+        public Uri? Url { get; set; }
 #if NET5_0_OR_GREATER
         public string? UdsFileName { get; set; }
 #endif
--- a/perf/benchmarkapps/GrpcClient/Program.cs
+++ b/perf/benchmarkapps/GrpcClient/Program.cs
@@ -17,6 +17,7 @@
 #endregion
 
 using System.CommandLine;
+using System.CommandLine.Binding;
 using System.CommandLine.Invocation;
 using System.Diagnostics;
 using System.Globalization;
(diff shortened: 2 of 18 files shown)


perf/benchmarkapps/GrpcClient/ClientOptions.cs
```diff
@@ -22,7 +22,7 @@ namespace GrpcClient
 {
     public class ClientOptions
     {
-        public string? Url { get; set; }
+        public Uri? Url { get; set; }
```
JamesNK, line 25: The actual change I wanted to make to figure out why benchmarks were failing.

"Might as well update System.Commandline while I'm here. Should be easy." <- famous last words

😢

perf/benchmarkapps/GrpcClient/Program.cs
```diff
@@ -103,11 +110,39 @@ public static async Task<int> Main(string[] args)
...
+
+                foreach (var option in _options)
+                {
+                    var value = bindingContext.ParseResult.GetValueForOption(option);
+
+                    var propertyInfo = typeof(T).GetProperty(option.Name, BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance);
+                    if (propertyInfo != null)
+                    {
+                        propertyInfo.SetValue(boundValue, value);
+                    }
```
JamesNK, line 139: yolo

perf/benchmarkapps/GrpcClient/Program.cs
```diff
@@ -103,11 +108,45 @@ public static async Task<int> Main(string[] args)
...
+            {
+                var boundValue = new T();
+
+                Log($"Binding {typeof(T)}");
+
+                foreach (var option in _options)
+                {
+                    var value = bindingContext.ParseResult.GetValueForOption(option);
+
+                    var propertyInfo = typeof(T).GetProperty(option.Name, BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance);
```
captainsafia, line 137: We own the `ClientOptions` type correct? Is using reflection the only way to assign the parsed values to the `ClientOptions` object here?

JamesNK: We could store each option, then set each property individually using the option. But that would be a real pain in the butt.

It would look like this:
```cs
using System.CommandLine;
using System.CommandLine.Binding;

public class Program
{
    internal static async Task Main(string[] args)
    {
        var fileOption = new Option<FileInfo?>(
              name: "--file",
              description: "An option whose argument is parsed as a FileInfo",
              getDefaultValue: () => new FileInfo("scl.runtimeconfig.json"));

        var firstNameOption = new Option<string>(
              name: "--first-name",
              description: "Person.FirstName");

        var lastNameOption = new Option<string>(
              name: "--last-name",
              description: "Person.FirstName");

        var rootCommand = new RootCommand();
        rootCommand.Add(fileOption);
        rootCommand.Add(firstNameOption);
        rootCommand.Add(lastNameOption);

        rootCommand.SetHandler((FileInfo fileOptionValue, Person person) =>
        {
            DoRootCommand(fileOptionValue, person);
        },
        fileOption, new PersonBinder(firstNameOption, lastNameOption));

        await rootCommand.InvokeAsync(args);
    }

    public static void DoRootCommand(FileInfo aFile, Person aPerson)
    {
        Console.WriteLine($"File = {aFile?.FullName}");
        Console.WriteLine($"Person = {aPerson?.FirstName} {aPerson?.LastName}");
    }

    public class Person
    {
        public string? FirstName { get; set; }
        public string? LastName { get; set; }
    }

    public class PersonBinder : BinderBase<Person>
    {
        private readonly Option<string> _firstNameOption;
        private readonly Option<string> _lastNameOption;

        public PersonBinder(Option<string> firstNameOption, Option<string> lastNameOption)
        {
            _firstNameOption = firstNameOption;
            _lastNameOption = lastNameOption;
        }

        protected override Person GetBoundValue(BindingContext bindingContext) =>
            new Person
            {
                FirstName = bindingContext.ParseResult.GetValueForOption(_firstNameOption),
                LastName = bindingContext.ParseResult.GetValueForOption(_lastNameOption)
            };
    }
}
```
But much longer because there are many more options. I think System.Commandline is going to improve it in the future with source gen. But the current state is way too verbose.

https://docs.microsoft.com/en-us/dotnet/standard/commandline/model-binding#model-binding-more-than-16-options-and-arguments

captainsafia (approved):

Issue #1683, opened 2022-04-14
Add setting to allow AsyncAuthInterceptor over plaintext


JamesNK:
Today, by default, `AsyncAuthInterceptor` isn't used when making a gRPC call without TLS. A warning is logged instead. This behavior matches Grpc.Core. I suspect the reason is security and not sending auth headers over plain text.

There are some scenarios where only `AsyncAuthInterceptor` can be used for a certain task, e.g. inject async logic before a gRPC call. There are also some scenarios where plaintext makes sense to use, e.g. dev environments.

I think a setting could be added to GrpcChannelOptions to allow auth interceptors over plaintext. We can add `Unsafe` or `Dangerous` to the name to let people know that this isn't recommended:

```cs
options.UnsafeAllowAuthInterceptorWithInsecure = true;
```

JamesNK:
@jtattermusch Thoughts here on this divergence from Grpc.Core? The default behavior is compatible. This just adds a new setting to customize behavior.

PR #1705, opened 2022-04-25
Add call credential methods to client factory and flag to always use call credentials

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1683

@jtattermusch do you have any thoughts on this setting? The default behavior is true to Grpc.Core. This settings adds an option to use call credentials with all channels.

Also, PR adds an easy way to configure call credentials with client factory using the current DI scope. See docs PR https://github.com/dotnet/AspNetCore.Docs/pull/25620.


jtattermusch (approved):
LGTM.

PR #1708, opened 2022-04-26
Move Grpc.HealthCheck and Grpc.Reflection from grpc/grpc

jtattermusch:
~Based on https://github.com/grpc/grpc-dotnet/pull/1652.~  Rebased


test/Grpc.HealthCheck.Tests/HealthClientServerTest.cs
```diff
@@ -0,0 +1,83 @@
...
+    public class HealthClientServerTest
+    {
+        const string Host = "localhost";
+        Server server;
+        Channel channel;
+        Grpc.Health.V1.Health.HealthClient client;
+        Grpc.HealthCheck.HealthServiceImpl serviceImpl;
+
+        [OneTimeSetUp]
+        public void Init()
```
jtattermusch, line 41: @JamesNK  what would be the most straightforward way of converting this test suite (and a similar test suite for Grpc.Reflection) to use Grpc.Net.Client and Grpc.AspNetCore.Server instead of Grpc.Core?

jtattermusch:
@JamesNK  this is ready for the initial pass of review.

test/Grpc.HealthCheck.Tests/HealthClientServerTest.cs
JamesNK: There is a functional tests project that has a lot of infrastructure for hosting server and calling via client. Moving these tests to it would be simple enough.

JamesNK: I think it's fine to add a to-do here and move tests in the future.

jtattermusch: What I did for now, is that I added Grpc.Core package dependency for Grpc.HealthCheck.Tests.csproj and Grpc.Reflection.Tests.csproj. That makes the tests pass just fine and also keeps the code tested.
Added a TODO to migrate the tests to use client & server from grpc-dotnet.

jtattermusch:
@JamesNK  Rebased (on top of Grpc.Auth PR, which is small though) and ready for final review.

JamesNK (approved):

jtattermusch:
Interop test failure is a known problem (will fix as a followup).

PR #1717, opened 2022-04-28
Build nugets for recently moved packages from grpc/grpc

jtattermusch:
- add Grpc.Auth, Grpc.Core.Api, Grpc.HealthCheck and Grpc.Reflection to build_nuget.sh
- starting from 2.47.x we will be releasing new minor version of these packages with the grpc-dotnet release

### DIFF
--- a/kokoro/build_nuget.sh
+++ b/kokoro/build_nuget.sh
@@ -29,9 +29,13 @@ mkdir -p artifacts
 
 build/expand_dev_version.sh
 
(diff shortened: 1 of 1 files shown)


jtattermusch:
Let me doublecheck that the nugets were actually built correctly before merging this.

JamesNK (approved):

jtattermusch:
I checked the nugets and they seem to have been properly generated.

PR #1721, opened 2022-04-29
Run interop tests against Java instead of Grpc C# (Grpc.Core)

jtattermusch:
Grpc C# has been removed from master branch of grpc/grpc, so we need to choose some other language to run the interop tests against.
Chose java because it builds more quickly than grpc C++. golang would also work.


jtattermusch:
Trying again against C++.

JamesNK (approved):

PR #1723, opened 2022-05-02
Add AddCallCredentials overload that actually takes CallCredentials

JamesNK:
Allows integration with APIs that create CallCredentials, e.g. [Grpc.Auth](https://github.com/grpc/grpc-dotnet/blob/3f63a9ac940a3f0d62b1dfab21008b6c540ba12d/src/Grpc.Auth/GoogleGrpcCredentials.cs#L53-L56)


jtattermusch (approved):
LGTM.

PR #1724, opened 2022-05-02
Add JSON transcoding example

JamesNK:


build/dependencies.props
```diff
@@ -10,7 +10,7 @@
     <MicrosoftAspNetCoreApp31PackageVersion>3.1.3</MicrosoftAspNetCoreApp31PackageVersion>
     <MicrosoftBuildLocatorPackageVersion>1.2.2</MicrosoftBuildLocatorPackageVersion>
     <MicrosoftBuildPackageVersion>16.0.461</MicrosoftBuildPackageVersion>
-    <MicrosoftCodeAnalysisNetAnalyzersPackageVersion>6.0.0</MicrosoftCodeAnalysisNetAnalyzersPackageVersion>
+    <MicrosoftCodeAnalysisNetAnalyzersPackageVersion>7.0.0-preview1.22168.1</MicrosoftCodeAnalysisNetAnalyzersPackageVersion>
```
captainsafia, line 13: Why did we need to bump analyzer dependencies here?

examples/Transcoder/Server/Server.csproj
```diff
@@ -0,0 +1,16 @@
...
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn>1591</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Protobuf Include="..\Proto\greet.proto" GrpcServices="Server" Link="Protos\greet.proto" />
+
+    <PackageReference Include="Grpc.AspNetCore" Version="$(GrpcDotNetPackageVersion)" />
```
captainsafia, line 12: Minor nit: Since this is a sample maybe it makes sense to hardcode the version for folks who are copy-pasting?

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.100",
-    "rollForward": "latestFeature"
+    "version": "6.0.202",
```
captainsafia, line 3: Why did we need to change the `rollForward` behavior here? Is there something about `6.0.300` that we don't want?

build/dependencies.props
JamesNK: Warnings are generated on build if an older version of this package is used with a newer SDK. Our build complained that 6.x was used with .NET 7.

examples/Transcoder/Server/Server.csproj
JamesNK: Hard coding with so many samples was a nightmare to update so many references. The README.md in the examples directory talks about what you need to do if a sample is copied out of the repo.

global.json
JamesNK: latestMajor means the build server will use .NET 7 SDK which install-dotnet.sh downloads and installs. Its required because the new example uses .NET 7.

I didn't update the version number here beyond 6 because this global.json is used by all source, including some code that is tested in containers. Those containers don't support the .NET 7 version we need.

This allows everything to work with a minimal amount of hackery.

captainsafia (approved):

Issue #1729, opened 2022-05-09
Exception being thrown when using the gRPC server reflection


danielvieiravega:
### What version of gRPC and what language are you using?
gRPC: nuget packages version 2.45.0 & C#

### What operating system (Linux, Windows,...) and version?
Windows 11

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.NET SDK (reflecting any global.json):
 Version:   6.0.202
 Commit:    f8a55617d2

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22000
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.202\

Host (useful for support):
  Version: 6.0.4
  Commit:  be98e88c76

.NET SDKs installed:
  3.1.418 [C:\Program Files\dotnet\sdk]
  5.0.407 [C:\Program Files\dotnet\sdk]
  6.0.100 [C:\Program Files\dotnet\sdk]
  6.0.202 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.24 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.13 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.16 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.24 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.13 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.16 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.1 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.24 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
[...]

PR #1731, opened 2022-05-10
Validate grpc reflection required service

danielvieiravega:
This PR solves the problem related in this [issue](https://github.com/grpc/grpc-dotnet/issues/1729).

It is being checked whether the services needed to use gRPC reflection have been added. If not, an `InvalidOperationException` is thrown, asking to call the `IServiceCollection.AddGrpcReflection()` method.


danielvieiravega:
I opened this other [PR ](https://github.com/grpc/grpc-dotnet/pull/1732) in order to fix the easyCLA check.

PR #1732, opened 2022-05-10
Ensuring required services are registered when using the server reflection

danielvieiravega:
This PR solves the problem related in this [issue](https://github.com/grpc/grpc-dotnet/issues/1729).

It is being checked whether the services needed to use gRPC reflection have been added. If not, an `InvalidOperationException` is thrown, asking to call the `IServiceCollection.AddGrpcReflection()` method.

### DIFF
--- a/src/Grpc.AspNetCore.Server.Reflection/GrpcReflectionEndpointRouteBuilderExtensions.cs
+++ b/src/Grpc.AspNetCore.Server.Reflection/GrpcReflectionEndpointRouteBuilderExtensions.cs
@@ -18,6 +18,7 @@
 
 using Grpc.Reflection;
 using Microsoft.AspNetCore.Routing;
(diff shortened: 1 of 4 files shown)


JamesNK:
This is great. Exactly how I was thinking of improving the error message.

I'd like to have a unit test of this failing. There is an example of doing this here: https://github.com/grpc/grpc-dotnet/blob/713935fbf7c06b4864dc66ab023c34d5d6289677/test/Grpc.AspNetCore.Server.Tests/GrpcEndpointRouteBuilderExtensionsTests.cs#L37-L49

danielvieiravega:
I added the test @JamesNK.

JamesNK (approved):

JamesNK:
Perfect. Thanks!

Issue #1713, opened 2022-04-27
"Error starting gRPC call" - concurrency issue.


mmhet:
### What version of gRPC and what language are you using?
C#

```
    <PackageReference Include="Grpc.AspNetCore" Version="2.45.0" />
    <PackageReference Include="Grpc.Net.ClientFactory" Version="2.45.0" />
    <PackageReference Include="Google.Protobuf" Version="3.20.1" />
    <PackageReference Include="Grpc" Version="2.45.0" />
    <PackageReference Include="Grpc.Tools" Version="2.45.0" PrivateAssets="All" />
```

### What operating system (Linux, Windows,...) and version?
- Windows 10 (10.0.19044.0)
- Windows Server 2016 (10.0.14393.0)

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
Visual Studio 2022 17.1.3.

Windows 10 (10.0.19044.0) `dotnet --info`:
```
.NET SDK (reflecting any global.json):
 Version:   6.0.201
 Commit:    ef40e6aa06

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19044
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.201\

Host (useful for support):
  Version: 6.0.3
  Commit:  c24d9a9c91

.NET SDKs installed:
  1.0.0-preview2-003131 [C:\Program Files\dotnet\sdk]
  1.1.10 [C:\Program Files\dotnet\sdk]
  1.1.11 [C:\Program Files\dotnet\sdk]
  1.1.12 [C:\Program Files\dotnet\sdk]
  1.1.13 [C:\Program Files\dotnet\sdk]
  1.1.14 [C:\Program Files\dotnet\sdk]
  2.1.4 [C:\Program Files\dotnet\sdk]
  2.1.104 [C:\Program Files\dotnet\sdk]
  2.1.201 [C:\Program Files\dotnet\sdk]
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.400 [C:\Program Files\dotnet\sdk]
  2.1.402 [C:\Program Files\dotnet\sdk]
  2.1.403 [C:\Program Files\dotnet\sdk]
  2.1.500 [C:\Program Files\dotnet\sdk]
  2.1.502 [C:\Program Files\dotnet\sdk]
  2.1.503 [C:\Program Files\dotnet\sdk]
[...]

PR #1752, opened 2022-05-19
Make getting lazy properties on ServiceConfig threadsafe

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1713

### DIFF
--- a/src/Grpc.Net.Client/Internal/Configuration/ConfigProperty.cs
+++ b/src/Grpc.Net.Client/Internal/Configuration/ConfigProperty.cs
@@ -20,7 +20,7 @@ using Grpc.Net.Client.Configuration;
 
 namespace Grpc.Net.Client.Internal.Configuration
 {
-    internal struct ConfigProperty<TValue, TInner> where TValue : IConfigValue
+    internal sealed class ConfigProperty<TValue, TInner> where TValue : IConfigValue
     {
         private TValue? _value;
         private readonly Func<TInner?, TValue?> _valueFactory;
@@ -37,13 +37,23 @@ namespace Grpc.Net.Client.Internal.Configuration
         {
             if (_value == null)
             {
-                var innerValue = inner.GetValue<TInner>(_key);
-                _value = _valueFactory(innerValue);
-
-                if (_value != null && innerValue == null)
+                // Multiple threads can get a property at the same time. We want this to be safe.
+                // Because a value could be lazyily initialized, lock to ensure multiple threads
+                // don't try to update the underlying dictionary at the same time.
+                lock (this)
                 {
-                    // Set newly created value
-                    SetValue(inner, _value);
+                    // Double-check locking.
+                    if (_value == null)
+                    {
+                        var innerValue = inner.GetValue<TInner>(_key);
+                        _value = _valueFactory(innerValue);
+
+                        if (_value != null && innerValue == null)
+                        {
+                            // Set newly created value
+                            SetValue(inner, _value);
+                        }
+                    }
                 }
             }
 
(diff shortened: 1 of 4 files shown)


src/Grpc.Net.Client/Internal/Configuration/ConfigProperty.cs
```diff
@@ -37,13 +37,23 @@ public ConfigProperty(Func<TInner?, TValue?> valueFactory, string key)
         {
             if (_value == null)
             {
-                var innerValue = inner.GetValue<TInner>(_key);
-                _value = _valueFactory(innerValue);
-
-                if (_value != null && innerValue == null)
+                // Multiple threads can get a property at the same time. We want this to be safe.
+                // Because a value could be lazyily initialized, lock to ensure multiple threads
```
BrennanConroy, line 41:
```suggestion
                // Because a value could be lazily initialized, lock to ensure multiple threads
```

test/Grpc.Net.Client.Tests/ServiceConfigTests.cs
```diff
@@ -137,6 +137,38 @@ public void LoadBalancingConfig_CreateUnderlyingConfig()
...
+            for (var i = 0; i < 100; i++)
+            {
+                connectTasks.Add(Task.Run(ConnectAsync));
+            }
+
+            await Task.WhenAll(connectTasks);
+
+            void ConnectAsync()
+            {
+                // Execuse to access LoadBalancingConfigs
```
BrennanConroy, line 164:
```suggestion
                // Excuse to access LoadBalancingConfigs
```

src/Grpc.Net.Client/Internal/Configuration/ConfigProperty.cs
```diff
@@ -37,13 +37,23 @@ public ConfigProperty(Func<TInner?, TValue?> valueFactory, string key)
...
+                    // Double-check locking.
+                    if (_value == null)
+                    {
+                        var innerValue = inner.GetValue<TInner>(_key);
+                        _value = _valueFactory(innerValue);
+
+                        if (_value != null && innerValue == null)
+                        {
+                            // Set newly created value
+                            SetValue(inner, _value);
```
BrennanConroy, line 54: Should the SetValue function also lock since it's public and called by other code?

BrennanConroy (approved):

src/Grpc.Net.Client/Internal/Configuration/ConfigProperty.cs
halter73: `ConfigProperty` and `IConfigValue` are both internal so I think it's okay. If this was ever exposed anywhere, locking on `this` would be bad. But unless it's being exposed as `object`, I don't see how anyone would easily get a reference.

halter73 (approved):

src/Grpc.Net.Client/Internal/Configuration/ConfigProperty.cs
```diff
@@ -37,13 +37,23 @@ public ConfigProperty(Func<TInner?, TValue?> valueFactory, string key)
...
+                // don't try to update the underlying dictionary at the same time.
+                lock (this)
                 {
-                    // Set newly created value
-                    SetValue(inner, _value);
+                    // Double-check locking.
+                    if (_value == null)
+                    {
+                        var innerValue = inner.GetValue<TInner>(_key);
+                        _value = _valueFactory(innerValue);
```
halter73, line 49: I probably should have asked this before approving, but can `_valueFactory` call into user code? Or `inner.GetValue/SetValue`?

JamesNK: No, it's internal.

src/Grpc.Net.Client/Internal/Configuration/ConfigProperty.cs
JamesNK: > Should the SetValue function also lock since it's public and called by other code?

The goal is to make the type thread-safe for reads, e.g. multiple threads can do this without error:

```csharp
if (serviceConfig.LoadBalancingConfigs.Count > 0)
{
    // ....
}
```

If someone calls SetValue value by setting a property on another thread, e.g. `serviceConfig.LoadBalancingConfigs = new List<LoadBalancingConfig>()` then an error is fine.

PR #1763, opened 2022-06-03
Use NuGet package in Vigor example

mic-max:


JamesNK:
Thanks. Could you also remove all the Grpc.* projects from the vigor solution.

JamesNK (changes requested):
Thanks. Could you also remove all the Grpc.* projects from the vigor solution.

JamesNK:
Replaced by https://github.com/grpc/grpc-dotnet/pull/1805

Issue #1511, opened 2021-11-30
grpc-dotnet Interop tests highly flaky on grpc/grpc repository


jtattermusch:
This started flaking a few days ago.
Once of the first occurrences is 
https://source.cloud.google.com/results/invocations/29f7f9fc-47d6-4846-8ec4-368cbb688436/targets/github%2Fgrpc%2Finterop_test/tests
and it's been ~50% flaky since.

Likely caused by a recent change in grpc-dotnet?

![image](https://user-images.githubusercontent.com/9939684/144089744-dee4cb47-24ac-4197-b3e2-bba2c46fbd66.png)

JamesNK:
Error is coming from underlying HttpClient logic. Most likely culprit is updating interop tests to .NET 6 - https://github.com/grpc/grpc-dotnet/commit/b9cb8806ae24ee260648dc5f1f24ae552eaf57a7

JamesNK:
I couldn't reproduce this when calling the sandbox server from my local machine. I put server streaming and large unary in a loop and ran them both 100 times without error.

JamesNK:
Changed the interop tests to use net5.0. If tests consistently pass then that confirms the issue is a regression from 5.0 to 6.0.

jtattermusch:
Looks like after merging https://github.com/grpc/grpc-dotnet/pull/1514, the failure on the CI went away, so this really about the net5.0 vs net6.0 difference.

I also tried to reproduce with net6.0 locally and also wasn't able to do so.

jtattermusch:
Since I wasn't able to reproduce locally, I tried to find the CI log where the test is failing with the extra diagnostic info (https://github.com/grpc/grpc-dotnet/pull/1513) but before we reverted back to net5.0 https://github.com/grpc/grpc-dotnet/pull/1514.

Looks like this is the log:
https://source.cloud.google.com/results/invocations/e229d041-9371-4e16-99bb-592991ebf10b/targets/github%2Fgrpc%2Finterop_test/tests;group=tests;test=cloud_to_prod:aspnetcore:default:large_unary:tls;row=3

Is that helpful?

JamesNK:
I believe the issue is SocketsHttpHandler is sending too many pings, and the server is closing the connection. This is for a feature to measure latency between the client and server, and dynamically adjust window sizes. This is a new feature in .NET 6.

See https://github.com/dotnet/runtime/issues/62216

[...]

PR #1765, opened 2022-06-07
Revert "Change InteropTestClient to net5.0"

jtattermusch:
Reverts grpc/grpc-dotnet#1514

Switch interop tests back to net6.0, since the google backends have been updated to avoid the "too many pings" issue caused by the .NET 6 "dynamic window sizing" feature (see https://github.com/grpc/grpc-dotnet/issues/1511 and https://github.com/dotnet/runtime/issues/62216)


jtattermusch:
CC @jskeet

jtattermusch:
I'll keep an eye on the interop test results for the next few days.

PR #1773, opened 2022-06-19
Add HTTP event source logging to gRPC benchmark server

JamesNK:


perf/benchmarkapps/GrpcAspNetCoreServer/Program.cs
```diff
@@ -32,12 +33,19 @@ public static void Main(string[] args)
 
         public static IHostBuilder CreateHostBuilder(string[] args)
         {
+            var runtimeVersion = typeof(object).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "Unknown";
```
BrennanConroy, line 36: Can you use https://github.com/dotnet/crank/blob/894f93ddedc2db2de723842e41be1d31708a3f10/src/Microsoft.Crank.EventSources/BenchmarksEventSource.cs? That way it'll actually output info to eventsource which can then be displayed in Power BI.

JamesNK: This app is already doing that. I want to have this info in the console as well.

BrennanConroy (approved):

PR #1782, opened 2022-06-22
Update version to 2.47.0-pre1 (on v2.47.x branch)

jtattermusch:
I'll create the prerelease and push the nuget packages after this PR gets merged.
Note that this is going to be the first release in which we publish the nuget packages that were recently moved from the grpc/grpc repository (e.g. `Grpc.Core.Api`)

CC @apolcyn

### DIFF
--- a/src/Grpc.Core.Api/VersionInfo.cs
+++ b/src/Grpc.Core.Api/VersionInfo.cs
@@ -41,6 +41,6 @@ namespace Grpc.Core
         /// <summary>
         /// Current version of gRPC C#
         /// </summary>
-        public const string CurrentVersion = "2.47.0-dev";
+        public const string CurrentVersion = "2.47.0-pre1";
     }
 }


JamesNK (approved):

jtattermusch:
> `GrpcPackageVersion` needs to be updated for new Grpc.Tools.

Done (bumped to 2.47.0 which just got released).

jtattermusch:
Looks like we'd need a few more adjustments - while `GrpcPackageVersion` is used mostly for `Grpc.Tools` package dependency, there are a few more place where it's used as a version for other packages:
https://github.com/grpc/grpc-dotnet/search?p=1&q=GrpcPackageVersion (e.g. `Grpc.Core`, `Grpc.Core.Api` etc.).
So it seems that we'll need to split `GrpcPackageVersion` property into more separate properties.
(e.g. `GrpcToolsPackageVersion`, `GrpcCorePackageVersion` and use the existing `GrpcDotnetVersion` as version for packages that have been moved to grpc-dotnet repo - Grpc.Core.Api etc.)

jtattermusch:
Since https://github.com/grpc/grpc-dotnet/pull/1785 is now both in master and in v2.47.x (and GrpcToolsPackageVersion is now at 2.47.0), simply updating version.props should be sufficient.

jtattermusch:
I just pushed the nugets (this time includes recently moved packages Grpc.Auth, Grpc.Core.Api, Grpc.Reflection, Grpc.HealthCheck).

https://github.com/grpc/grpc-dotnet/releases/tag/v2.47.0-pre1

PR #1786, opened 2022-06-23
Additional dependency version cleanup (after moving some nuget packages to from grpc to grpc-dotnet)

jtattermusch:
Some occurrences of `PrivateAssets="All"` seems unnecessary?
Grpc.Reflection and Grpc.HealthCheck now live in grpc-dotnet repository, so their version should be GrpcDotNetPackageVersion in examples?


JamesNK:
Is it ok to close this?

JamesNK (approved):

PR #1791, opened 2022-06-27
Add more load balancing logging

JamesNK:


src/Grpc.Net.Client/Balancer/Subchannel.cs
```diff
@@ -101,25 +103,41 @@ public IDisposable OnStateChanged(Action<SubchannelState> callback)
...
                 _subchannel = subchannel;
                 _callback = callback;
+                _registrationId = subchannel.GetNextRegistrationId();
+
+                SubchannelLog.StateChangedRegistrationCreated(_subchannel._logger, _subchannel.Id, _registrationId);
             }
 
             public void Invoke(SubchannelState state)
             {
+                SubchannelLog.ExecutingStateChangedRegistration(_subchannel._logger, _subchannel.Id, _registrationId);
```
gfoidl, line 131: These are `LogLeve.Trace`, so mostly disabled (in production), but the allocation(s) from the registration-ID sill occur. If that's relevant, these allocation could be avoided by
```c#
if (_subchannel._logger.IsEnabled(LogLevel.Trace))
{
    SubchannelLog.FooBar(..., RegistrationId);
}
```
where `RegistrationId` is lazy-evaluated property that build the string from the id stored as field.

Below in the `LoggerMessage.Define` the `skipEnabledCheck` has to be set to `true`.

JamesNK: That's true, but channels are long-lived and there is probably only ever one registration. It's not worth trying to save.

src/Grpc.Net.Client/Balancer/Subchannel.cs
```diff
@@ -356,7 +378,11 @@ public IReadOnlyList<BalancerAddress> GetAddresses()
         public void Dispose()
         {
             UpdateConnectivityState(ConnectivityState.Shutdown, "Subchannel disposed.");
-            _stateChangedRegistrations.Clear();
+            for (var i = _stateChangedRegistrations.Count - 1; i >= 0; i--)
+            {
+                _stateChangedRegistrations[i].Dispose();
```
BrennanConroy, line 383: This is removing itself from the list that you are currently trying to iterate over, that is not good.

JamesNK: The for loop is reversed so it doesn't error.

Do you think a `while (_stateChangedRegistrations.Count > 0)` would be better?

src/Grpc.Net.Client/Balancer/SubchannelsLoadBalancer.cs
```diff
@@ -340,6 +344,9 @@ internal static class SubchannelsLoadBalancerLog
         private static readonly Action<ILogger, int, ConnectivityState, Exception?> _requestingConnectionForSubchannel =
             LoggerMessage.Define<int, ConnectivityState>(LogLevel.Trace, new EventId(5, "RequestingConnectionForSubchannel"), "Requesting connection for subchannel id '{SubchannelId}' because it is in state {State}.");
 
+        private static readonly Action<ILogger, int, string, Exception?> _creatingReadyPicker =
+            LoggerMessage.Define<int, string>(LogLevel.Trace, new EventId(6, "CreatingReadyPicker"), "Creating ready picker with {SubchannelCount} subchannels: {Subchannels}");
```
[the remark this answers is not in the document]
JamesNK: The client still supports 3.1 and has a min version of 3.0 of Microsoft.Extensions.Logging. That version doesn't have SkipEnabledCheck.

src/Grpc.Net.Client/Balancer/Subchannel.cs
BrennanConroy: Oh man, that's confusing. Does the remove really need to be separate? Some form of rewrite would make me happier :D

JamesNK: I want to use dispose here to remove the registrations because it has logging. It means removing registrations is always consistent (dispose) vs something inconsistent (dispose vs Clear).

halter73 (approved):

Issue #1720, opened 2022-04-29
NotSupportedException: Specified method is not supported."


pengweiqhca:
We use H2C.
```
NotSupportedException: Specified method is not supported.", DebugException="System.Net.Http.HttpRequestException: Specified method is not supported. (internalapi.betawm.beta:80)
 ---> System.NotSupportedException: Specified method is not supported.
   at Grpc.Net.Client.Balancer.Internal.PassiveSubchannelTransport.GetStreamAsync(BalancerAddress address, CancellationToken cancellationToken)
   at Grpc.Net.Client.Balancer.Internal.BalancerHttpHandler.OnConnect(SocketsHttpConnectionContext context, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.ConnectToTcpHostAsync(String host, Int32 port, HttpRequestMessage initialRequest, Boolean async, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at System.Net.Http.HttpConnectionPool.ConnectToTcpHostAsync(String host, Int32 port, HttpRequestMessage initialRequest, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.ConnectAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.AddHttp2ConnectionAsync(HttpRequestMessage request)
   at System.Threading.Tasks.TaskCompletionSourceWithCancellation`1.WaitWithCancellationAsync(CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.GetHttp2ConnectionAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.DiagnosticsHandler.SendAsyncCore(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at Microsoft.Extensions.Http.Logging.LoggingHttpMessageHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
[...]

PR #1792, opened 2022-06-27
Fix sharing SocketsHttpHandler between channels

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1720

Finally figured this error out 🥳 🎉 🍰 

Can reproduce it when:

* A SocketsHttpHandler instance is shared between channels.
* The channels share the same address.
* The first request to a server is made by any channel **_other_** than the first.


src/Grpc.Net.Client/Balancer/Internal/BalancerHttpHandler.cs
```diff
@@ -46,13 +46,17 @@ public BalancerHttpHandler(HttpMessageHandler innerHandler, HttpHandlerType http
                 var socketsHttpHandler = HttpRequestHelpers.GetHttpHandlerType<SocketsHttpHandler>(innerHandler);
                 CompatibilityHelpers.Assert(socketsHttpHandler != null, "Should have handler with this handler type.");
 
-                socketsHttpHandler.ConnectCallback = OnConnect;
+                if (!HttpRequestHelpers.IsSocketsHttpHandlerSetup(socketsHttpHandler))
```
BrennanConroy, line 49: Is thread safety a concern? I followed this back to GrpcChannel.ForAddress etc. which I'm guessing is meant to be setup once on a single thread.

PR #1796, opened 2022-06-29
Add locator example

JamesNK:


JamesNK:
Review please 🙏

examples/Locator/Server/Startup.cs
```diff
@@ -0,0 +1,49 @@
...
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapGrpcService<InternalService>().RequireHost("*:5001");
```
brunolins16, line 44: I saw in the issue, that reported the problem with `RequireHost`, this call seems to be `app.MapGrpcService<InternalService>().RequireHost` instead in the `UseEndpoints` delegate. Is that just a typo in the issue since `MapGrpcService` is an `IEndpointRouteBuilder` extension method or it is been called in a different scenario?

examples/README.md
```diff
@@ -305,4 +305,15 @@ The downloader shows how to download a file in chunks using a server streaming g
 ##### Scenarios:
 
 * Server streaming
-* Binary payload
\ No newline at end of file
+* Binary payload
+
+## [Locator](./Locator)
+
+The locator shows how to add host name constraints to gRPC services. This example adds two services:
```
brunolins16, line 312: ❓

```suggestion
The locator shows how to add host constraints to gRPC services. This example adds two services:
```

brunolins16 (approved):

examples/Locator/Server/Startup.cs
JamesNK: It doesn't make a difference. It's different syntax for the same outcome.

PR #1798, opened 2022-06-30
Don't use sum for latency

SleepyBag:
Latencies are sum among different machines now. That makes no sense.


JamesNK:
@sebastienros Is this the right thing to do?

sebastienros:
Looks good, could be mean or max here, but I also prefer max since this is what is used for a single machine with multiple values. I will need to check if I made the same mistake in the crank repos. Note that this is only fixing an issue when you use simultaneous clients which is uncommon.

JamesNK (approved):

Issue #1800, opened 2022-06-30
Way to add call credentials to insecure grpc channel without using DI factory


Blackclaws:
There is a documented way to create a gRPC client using the dependency injection framework that allows the usage of call credentials on insecure channels:
```
builder.Services.AddGrpcClient<Greeter.GreeterClient>(o =>
{
    o.Address = new Uri("http://localhost:5060");
    o.ChannelOptionsActions.Add((opt) =>
    {
        opt.UnsafeUseInsecureChannelCallCredentials = true;
    });
}).AddCallCredentials((context, metadata) =>
{
    metadata.Add("Authorization",
        "Bearer ...");
    return Task.CompletedTask;
});
```
I have not found a way to do the same if I manually create a channel. It seems the classes I would need to use are all internal/private classes.  Is this by design or is there simply a way I'm missing to do this?

Blackclaws:
Ok sometimes the solution you tried to find for hours comes right after posting a question. So here it is:

You need to create a new ChannelCredential implementation it seems. This can do the same thing as the CompositeChannelCredentials but doesn't check whether IsComposable is true.

I'm leaving this up though to propose an API change because this seems cumbersome in the extreme.

How about changing the call to ChannelCredentials.Create to also take a boolean value: AcceptUnsafeComposition

Pass that to the constructor of CompositeChannelCredentials and turn the check:
```
 if (!channelCredentials.IsComposable)
```
into
```
 if (AcceptUnsafeComposition || !channelCredentials.IsComposable)
```

JamesNK:
Does this not work?
```cs
private static GrpcChannel CreateAuthenticatedChannel(string address)
{
    var credentials = CallCredentials.FromInterceptor((context, metadata) =>
    {
        if (!string.IsNullOrEmpty(_token))
        {
            metadata.Add("Authorization", $"Bearer {_token}");
        }
        return Task.CompletedTask;
    });

    var channel = GrpcChannel.ForAddress(address, new GrpcChannelOptions
    {
        Credentials = ChannelCredentials.Create(ChannelCredentials.Insecure, credentials)
        UnsafeUseInsecureChannelCallCredentials = true
    });
    return channel;
}
```

Blackclaws:
No, because ChannelCredentials.Create fails because ChannelCredentials.Insecure is not composable. You get this error:

```
[...]

PR #1802, opened 2022-07-01
Make ChannelCredentials.Create work with allowing insecure credentials

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1800

Note that this change means Grpc.Core using a newer version of Grpc.Core.Api no longer throws an error when calling `ChannelCredentials.Create(ChannelCredentials.Insecure, callCredentials)`. That could be remedied by updating Grpc.Core to throw an error inside the configurator like Grpc.Net.Client does, but Grpc.Core is in maintenance. IMO not worth it.

I think this is safe to change because new behavior goes from error to non-error, and I don't see a situation where anyone would ever depend on the error.

@jtattermusch Thoughts?

### DIFF
--- a/src/Grpc.Core.Api/ChannelCredentials.cs
+++ b/src/Grpc.Core.Api/ChannelCredentials.cs
@@ -89,6 +89,11 @@ public abstract class ChannelCredentials
     /// <summary>
     /// Returns <c>true</c> if this credential type allows being composed by <c>CompositeCredentials</c>.
     /// </summary>
+    /// <remark>
+    /// Note: No longer used. Decision on whether composition is allowed now happens in
+    /// <see cref="ChannelCredentialsConfiguratorBase.SetCompositeCredentials(object, ChannelCredentials, CallCredentials)"/>.
+    /// Internal property left for safety because Grpc.Core has internal access to Grpc.Core.Api.
+    /// </remark>
     internal virtual bool IsComposable => false;
 
     private sealed class InsecureCredentials : ChannelCredentials
@@ -118,11 +123,6 @@ public abstract class ChannelCredentials
         {
             this.channelCredentials = GrpcPreconditions.CheckNotNull(channelCredentials);
             this.callCredentials = GrpcPreconditions.CheckNotNull(callCredentials);
-
-            if (!channelCredentials.IsComposable)
-            {
-                throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "CallCredentials can't be composed with {0}. CallCredentials must be used with secure channel credentials like SslCredentials.", channelCredentials.GetType().Name));
-            }
         }
 
         public override void InternalPopulateConfiguration(ChannelCredentialsConfiguratorBase configurator, object state)
--- a/test/Grpc.Net.Client.Tests/CallCredentialTests.cs
+++ b/test/Grpc.Net.Client.Tests/CallCredentialTests.cs
@@ -159,6 +159,147 @@ public class CallCredentialTests
     }
 
     [Test]
+    public async Task CallCredentialsWithHttp_UnsafeUseInsecureChannelCallCredentials_ChannelCredentials_MetadataOnRequest()
+    {
+        // Arrange
+        string? authorizationValue = null;
+        var httpClient = ClientTestHelpers.CreateTestClient(async request =>
+        {
+            authorizationValue = request.Headers.GetValues("authorization").Single();
+
+            var reply = new HelloReply { Message = "Hello world" };
+            var streamContent = await ClientTestHelpers.CreateResponseContent(reply).DefaultTimeout();
+            return ResponseUtils.CreateResponse(HttpStatusCode.OK, streamContent);
+        }, new Uri("http://localhost"));
+        var callCredentials = CallCredentials.FromInterceptor(async (context, metadata) =>
+        {
+            // The operation is asynchronous to ensure delegate is awaited
+            await Task.Delay(50);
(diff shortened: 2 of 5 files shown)


JamesNK:
@apolcyn Thoughts?

JamesNK:
@apolcyn ping. What do you think about this?

apolcyn:
In the core library, when someone composes a call credentials with an insecure channel creds, the call creds are just silently dropped.

The motivation for that was security: call creds typically add sensitive headers. So they wanted to make it hard for people to accidentally send them on unencrypted connections.

Personally, I think throwing an error achieves the same original goal in a more user friendly way. So the original behavior actually seems right to me.

But I'm saying all this without full background here - have we considered that security motivation here?

JamesNK:
The goal is to not throw an error if the unsecured channel is configured to allow sending call credentials.

`GrpcChannel` has an `UnsafeUseInsecureChannelCallCredentials` setting (false by default). Validation of whether a call credential is allowed is moved into the channel so it can decide whether to throw an error. Is that ok with the current Grpc.Core code?

src/Grpc.Core.Api/ChannelCredentials.cs
```diff
@@ -88,6 +88,11 @@ public static ChannelCredentials Create(ChannelCredentials channelCredentials, C
 
         /// <summary>
         /// Returns <c>true</c> if this credential type allows being composed by <c>CompositeCredentials</c>.
+        /// <para>
```
captainsafia, line 91:
```suggestion
        /// <remark>
```

Nit.

apolcyn:
Even though it's not really breaking anything for `Grpc.Core` (since the headers still won't go out), it's a regression to go from obvious error to silent error (not sending headers).

Also, I'm kind of assuming that this will simply cause headers to be silently dropped in `Grpc.Core` - I think it wouldn't be out of the question if this caused `Grpc.Core` to crash when trying to compose insecure with call creds (`Grpc.Core` has special handling for insecure channel creds by representing them with `null` in C#).

If we were to go with this, I'd probably be on the side of changing the [configurator](https://github.com/grpc/grpc/blob/2e696fa051d50db7e8a41ebef033fc84c871577a/src/csharp/Grpc.Core/Internal/DefaultChannelCredentialsConfigurator.cs#L67) in Grpc.Core, too.

JamesNK:
As far as I know Grpc.Core isn't being updated except for security issues.

I don't think we should block improving grpc-dotnet for Grpc.Core in this case as the only change is not throwing an exception. And there is already the silent behavior if credentials are specified for a gRPC call using `CallOptions.CallCredentials`.

vahpetr:
This fix need for workaround for macos. Please merge!

src/Grpc.Core.Api/ChannelCredentials.cs
```diff
@@ -88,6 +88,11 @@ public static ChannelCredentials Create(ChannelCredentials channelCredentials, C
...
+        /// Note: No longer used. Decision on whether composition is allowed now happens in
+        /// <see cref="ChannelCredentialsConfiguratorBase.SetCompositeCredentials(object, ChannelCredentials, CallCredentials)"/>.
+        /// Internal property left for safety because Grpc.Core has internal access to Grpc.Core.Api.
+        /// </para>
```
JamesNK, line 95:
```suggestion
        /// </remark>
```

jtattermusch (approved):
LGTM, but let's wait until test failures in https://github.com/grpc/grpc/pull/31376 are resolved and the PR is merged.

I'd also like @captainsafia  to approve.

test/Grpc.Net.Client.Tests/CallCredentialTests.cs
```diff
@@ -158,6 +158,147 @@ public async Task CallCredentialsWithHttp_UnsafeUseInsecureChannelCallCredential
...
+        {
+            // The operation is asynchronous to ensure delegate is awaited
+            await Task.Delay(50);
+            metadata.Add("authorization", "SECRET_TOKEN");
+        });
+
+        if (isSecure)
+        {
+            // Act && Assert
+            CreateInvoker(httpClient, isSecure, callCredentials);
```
captainsafia, line 257: Is it meaningful to add an assertion that uses the constructed invoker to issue a call and checks the token?

captainsafia (approved):

halter73 (approved):

JamesNK:
@jtattermusch When is Grpc.Core going to get a new release with the changes there? We should make sure the change in Grpc.Core.Api isn't publically available before Grpc.Core is.

I don't think there is urgency to merge this PR. No need to create a Grpc.Core release just for this.

jtattermusch:
> @jtattermusch When is Grpc.Core going to get a new release with the changes there? We should make sure the change in Grpc.Core.Api isn't publically available before Grpc.Core is.
> 
> I don't think there is urgency to merge this PR. No need to create a Grpc.Core release just for this.

We are not planning a Grpc.Core patch release from the 1.46.x branch at the moment, but if we don't happen to need one within the next month or so, I can make one.

MeysamMoghaddam:
@JamesNK  Not applied in version 2.51.0?

bugproof:
💀 can't test my app on local network without messing with certificates

danielwinkler:
@JamesNK sorry to bug you, but is there a chance to create a release with this feature soonish?

JamesNK:
This is in 2.52.0-pre1, which is on NuGet now. Non-preview release will be in a week, or maybe two.

danielwinkler:
Thanks for the info @JamesNK !

PR #1803, opened 2022-07-01
Update solution to .NET 7

JamesNK:


examples/Aggregator/Client/Client.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
```
jtattermusch, line 5: does it make sense for examples to be switched over to net7.0 which is only in preview currently?
Wouldn't it make more sense to keep examples at the latest stable (net6.0) and only update our libraries to net7?

test/Grpc.Net.Client.Tests/DeadlineTests.cs
```diff
@@ -406,36 +407,6 @@ public async Task AsyncServerStreamingCall_DeadlineDuringWrite_ResponseThrowsDea
             Assert.AreEqual(StatusCode.DeadlineExceeded, call.GetStatus().StatusCode);
         }
 
-        [Test]
```
jtattermusch, line 409: why remove this test?

testassets/InteropTestsClient/InteropTestsClient.csproj
```diff
@@ -2,8 +2,8 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks Condition="'$(LatestFramework)'!='true'">net6.0;net5.0;net472</TargetFrameworks>
-    <TargetFrameworks Condition="'$(LatestFramework)'=='true'">net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(LatestFramework)'!='true'">net7.0;net6.0;net5.0;net472</TargetFrameworks>
+    <TargetFrameworks Condition="'$(LatestFramework)'=='true'">net7.0</TargetFrameworks>
```
jtattermusch, line 6: I think updating this will break the interop tests on grpc/grpc? We'd need multiple separate changes to upgrade grpc/grpc's interop for grpc-dotnet to net7.

testassets/Shared/InteropClient.cs
```diff
@@ -168,7 +168,7 @@ private static WinHttpHandler CreateWinHttpHandler()
         {
 #pragma warning disable CA1416 // Validate platform compatibility
             var handler = new WinHttpHandler();
-            handler.SslProtocols = SslProtocols.Tls12 | SslProtocols.Tls11 | SslProtocols.Tls;
+            handler.SslProtocols = SslProtocols.Tls12;
```
jtattermusch, line 171: what's the impact of this change?

jtattermusch (commented):
Left a few comments. I'd also like @captainsafia  to review.

jtattermusch (approved):

testassets/Shared/InteropClient.cs
JamesNK: This is test only. Removed because older TLS enum values are obsolete.

testassets/InteropTestsClient/InteropTestsClient.csproj
JamesNK: Maybe. This is already passing on interop tests CI for this PR.

examples/Aggregator/Client/Client.csproj
JamesNK: It will be RC1 very soon. Changing examples back to net6.0 if someone wants is very easy.

testassets/InteropTestsWebsite/InteropTestsWebsite.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(LatestFramework)'!='true'">net6.0</TargetFrameworks>
-    <TargetFrameworks Condition="'$(LatestFramework)'=='true'">net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(LatestFramework)'!='true'">net7.0</TargetFrameworks>
```
captainsafia, line 4: Not directly related to this change, but why are two different conditions needed here?

testassets/InteropTestsClient/InteropTestsClient.csproj
```diff
@@ -44,13 +44,8 @@
     <PackageReference Include="System.Net.Http.WinHttpHandler" Version="$(SystemNetHttpWinHttpHandlerPackageVersion)" />
   </ItemGroup>
 
-  <ItemGroup Condition="$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 5.0))">
+  <ItemGroup Condition="'$(TargetFramework)' == 'net5.0' OR '$(TargetFramework)' == 'net6.0' OR '$(TargetFramework)' == 'net7.0'">
```
captainsafia, line 47: Any reason we're no longer `VersionGreaterThanOrEquals`? It's a bit more future-proof.

captainsafia (approved):
Is the inclusion of `src/Grpc.AspNetCore/lib/net7.0/_._` intentional?

Looks good other than some non-blocking comments inline.

testassets/InteropTestsClient/InteropTestsClient.csproj
JamesNK: I couldn't get it to work after adding net7.0 target 🤷

testassets/InteropTestsWebsite/InteropTestsWebsite.csproj
JamesNK: Interop tests are also run from grpc/grpc repo, and may specify a different framework.

JamesNK:
> Is the inclusion of `src/Grpc.AspNetCore/lib/net7.0/_._` intentional?

Yes. Grpc.AspNetCore is a metapackage and doesn't have a DLL itself. Adding this file keeps the build happy.

PR #1825, opened 2022-07-17
Migrate reflection api to v1 from v1alpha.

xclud:
Reflection v1 seems to be around for a while now. This PR migrates from v1alpha to v1.

https://github.com/grpc/grpc/blob/master/src/proto/grpc/reflection/v1/reflection.proto


JamesNK:
Decisions from talk with @jtattermusch :
* We can probably ship AlphaV1 and V1 in the same package. At least for some time, before deprecating AlphaV1.
* Current implementation logic could be split out into an internal class that's used by AlphaV1 and V1 implementations
* AspNetCore.Reflection package can have configuration to allow someone to specify which version to use

JamesNK:
No other languages have moved to v1 yet. I've created an issue for doing this in the future - https://github.com/grpc/grpc-dotnet/issues/1886

For now, I will close this PR until there is a cross-language plan for doing this.

Issue #1828, opened 2022-07-26
ConnectTimeout option doesn't work since version 2.44


Ilmarii:
### What version of gRPC and what language are you using?
Grpc.Net.Client 2.47.0 /C#

### What operating system (Linux, Windows,...) and version?
Linux

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.NET SDK  6.0.102

### What did you do?
I used `SocketsHttpHandler.ConnectTimeout` to limit the connection handshake time when the remote host is not responding.
Unfortunately I found that this option is ignored and the connection ends after some large default timeout.

MRE:
```
var loggerFactory = LoggerFactory.Create(logging =>
            {
                logging.AddConsole(x => x.TimestampFormat = "[HH:mm:ss] ");
                logging.SetMinimumLevel(LogLevel.Trace);
            });

            Stopwatch stopwatch = new Stopwatch();
            stopwatch.Start();

            try
            {
                var handler = new SocketsHttpHandler()
                {
                    ConnectTimeout =  TimeSpan.FromSeconds(10),
                };
                var endpoint = "http://1.1.1.1:2379"; // any endpoint that silently ignores all connections
                var channel = GrpcChannel.ForAddress(endpoint, new GrpcChannelOptions(){HttpHandler = handler, LoggerFactory = loggerFactory});

                var client = new Greeter.GreeterClient(channel);
                var reply = await client.SayHelloAsync(new HelloRequest { Name = "test" });
                Console.WriteLine("Greeting: " + reply.Message);
            }
            catch (Exception e)
            {
                Console.WriteLine($"Exception: {e}");
            }
            stopwatch.Stop();
            Console.WriteLine($"Elapsed: {stopwatch.Elapsed}");
```
### What did you expect to see?
A error after 10 seconds

### What did you see instead?
A error after ~130 seconds

### More logs
```
[17:19:04] dbug: Grpc.Net.Client.Internal.GrpcCall[1]
      Starting gRPC call. Method type: 'Unary', URI: 'http://1.1.1.1:2379/tutorial.Greeter/SayHello'.
[17:19:04] dbug: Grpc.Net.Client.Balancer.Subchannel[1]
      Subchannel id '1' created with addresses: 1.1.1.1:2379
[17:19:04] dbug: Grpc.Net.Client.Balancer.Internal.ConnectionManager[4]
      Channel picker updated.
[...]

PR #1829, opened 2022-07-27
Use SocketsHttpHandler.ConnectTimeout on connect

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1828

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Subchannel.cs
+++ b/src/Grpc.Net.Client/Balancer/Subchannel.cs
@@ -172,7 +172,7 @@ namespace Grpc.Net.Client.Balancer
 
             if (requireReconnect)
             {
-                _connectContext?.CancelConnect();
+                CancelInProgressConnect();
                 Transport.Disconnect();
                 RequestConnection();
             }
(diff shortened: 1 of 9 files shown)


test/FunctionalTests/Balancer/ConnectionTests.cs
```diff
@@ -47,6 +49,48 @@ namespace Grpc.AspNetCore.FunctionalTests.Balancer
...
+                return new HelloReply { Message = request.Name };
+            }
+
+            // Arrange
+            using var endpoint = BalancerHelpers.CreateGrpcEndpoint<HelloRequest, HelloReply>(50051, UnaryMethod, nameof(UnaryMethod));
+
+            var handler = new SocketsHttpHandler()
+            {
+                // ConnectTimeout is so small that CT will always be canceled before Socket is used.
+                ConnectTimeout = TimeSpan.FromTicks(1),
```
BrennanConroy, line 80: Doesn't TimeSpan.Zero work? We all know that using anything with real time will be flaky.
And Tick(1) actually means 15 milliseconds on windows which isn't super fast.

BrennanConroy (approved):

test/FunctionalTests/Balancer/ConnectionTests.cs
JamesNK: TimeSpan.Zero doesn't work. The server has just started up so it should take longer than 15ms to establish the first connection. If it proves flaky, I'll come up with a better idea.

src/Grpc.Net.Client/Balancer/Subchannel.cs
```diff
@@ -213,12 +213,27 @@ public void RequestConnection()
...
+        {
+            var connectContext = _connectContext;
+            if (connectContext != null)
+            {
+                lock (Lock)
+                {
+                    // Cancel connect cancellation token.
+                    connectContext.CancelConnect();
+                    connectContext.Dispose();
+                    connectContext = null;
```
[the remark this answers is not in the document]
JamesNK: Ah, yes. Although it's no longer needed now that cancel connect no-ops if the context has been disposed of.

Issue #1839, opened 2022-08-09
Return full stacktrace on service startup failures


cove:
### Is your feature request related to a problem? Please describe.
I'd like to see if service stacktraces can be bubbled up during service startup.

While writing integration tests for a service the gRPC call would fail with an unknown error and the detailed error messages and adding log interceptors didn't make it any clearer to me where the issue was. The root cause was `IConfiguration` wasn't being passed into my service on startup during the test so some values were null and causing the service to not start up, but if the stackstrace was easily available, without having to debug the pkg, it would have reduce the amount of code to look though for the cause.

### Describe the solution you'd like
Specifically [this call](https://github.com/grpc/grpc-dotnet/blob/bce585246db48c655b744976e55cb6c3819bc6c1/src/Shared/Server/UnaryServerMethodInvoker.cs#L70) was discarding the stacktrace during startup due to the `finally`, could we catch the exception and bubble it up here?

```C#
        private async Task<TResponse> ResolvedInterceptorInvoker(TRequest resolvedRequest, ServerCallContext resolvedContext)
        {
            GrpcActivatorHandle<TService> serviceHandle = default;
            try
            {
                serviceHandle = ServiceActivator.Create(resolvedContext.GetHttpContext().RequestServices);
                return await _invoker(serviceHandle.Instance, resolvedRequest, resolvedContext);
            }
            finally
            {
                if (serviceHandle.Instance != null)
                {
                    await ServiceActivator.ReleaseAsync(serviceHandle);
                }
            }
        }
```

### Describe alternatives you've considered
Log interceptors weren't that helpful since it doesn't tell you which code the error came from.

### Additional context
It would also be nice if `IConfiguration` was injected during the test, but that's likely a different ticket.

JamesNK:
A finally doesn't catch the exception. It's rethrown automatically after the finally is complete. However, if a new error is thrown in the finally, I think that could replace the original exception.

Do you have an example of this happening? I don't understand what you're doing or how IConfiguration comes into the picture.

cove:
I have a test that looks like this:
```c#
        [Test]
        public async Task LoginTest()
        {
            var client = new SesOnline.Flatbuf.AuthService.AuthServiceClient(Channel);

            var response =  await client.Login(new LoginRequest
            {
                ClientCredentialType = ClientCredentialType.DebugTrustClient,
                ClientCredential = "test cred",
                RequestedCredentialType = ServiceCredentialType.JWT
            });

[...]

PR #1840, opened 2022-08-10
add logging of exceptions to sample code test logger

cove:
I used the example gRPC test code to start but didn't realize the exceptions were getting lost, this just includes logging any exceptions from the server too.  Ref: #1839


JamesNK:
I made a similar improvement at the same time - https://github.com/grpc/grpc-dotnet/pull/1841

Do you mind if I close this PR?

PR #1844, opened 2022-08-14
Fix reconnection issues with load balancers deriving from SubchannelsLoadBalancer

mot256:
In our integration tests we have found that client side load balancing would deadlock (calls will perpetually fail with "gRPC call deadline exceeded.") in certain scenarios.

We tested with the following setup:
- a client (calling service) configured to call another service behind a Kubernetes headless service (exposing the direct IPs of the PODs of the destination service via DNS lookup). 
- The calling service uses a client side load balancing implementation (derived from SubchannelsLoadBalancer) that load balances the calls to the different IPs of the destination service.
- The calling service is configured with retry and retry back-off policy
- A destination service exposing gRPC endpoints
- The destination service's PODs are setup with a Kubernetes health check

We have found that if the destination service's POD restarts due to a health check, the POD will restart keeping the same IP as before the restart. Depending on the timing of this restart, the retry policy and the DNS polling time, we randomly experience that calls would then fail permanently (even to all instances of the destination service). Especially when the DNS was refreshed before the POD managed to restart. 
Restarting the calling service solves the problem. It sometimes worked to kill one or all of the destination service's POD(s) or introduce a new instance (scaling it out)

After allot of debugging we have found the problem to be in the SubchannelsLoadBalancer.
I have created a unit test to simulate the same issue, and consequently came up with the provided fix in this PR.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/SubchannelsLoadBalancer.cs
+++ b/src/Grpc.Net.Client/Balancer/SubchannelsLoadBalancer.cs
@@ -75,7 +75,7 @@ namespace Grpc.Net.Client.Balancer
                     {
                         RemoveSubchannel(addressSubchannel.Subchannel);
                     }
-                    Controller.UpdateState(new BalancerState(ConnectivityState.TransientFailure, new ErrorPicker(status)));
+                    UpdateBalancingState(status);
                     break;
             }
         }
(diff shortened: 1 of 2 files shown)


mot256:
We have build a local version of the Grpc.Net.Client package (that includes this fix), and we have not experience any issues with this ever since.

JamesNK:
Closing in favor of https://github.com/grpc/grpc-dotnet/pull/1853

PR #1847, opened 2022-08-16
Update Grpc.Tools dependency to 2.48.0

jtattermusch:
I'll cut the release branch for 2.48.x after merging this.


JamesNK (approved):
I think there were some transient failures

PR #1853, opened 2022-08-18
Load balancing fix to clear disposed subchannels on resolver error

JamesNK:
Builds on https://github.com/grpc/grpc-dotnet/pull/1844

@mot256 I investigated your issue, and I think the problem and solution are simpler than your original changes. I think the underlying issue is disposed subchannels are incorrectly left in `_addressSubchannels` when there is a resolver error. Fixing that issue fixes the tests.

Could you test this and see whether the fix works for you?

### DIFF
--- a/test/Grpc.Net.Client.Tests/Balancer/ConnectionManagerTests.cs
+++ b/test/Grpc.Net.Client.Tests/Balancer/ConnectionManagerTests.cs
@@ -344,11 +344,7 @@ namespace Grpc.Net.Client.Tests.Balancer
                 LoadBalancingConfigs = { new RoundRobinConfig() }
             };
 
-            var transportFactory = new TestSubchannelTransportFactory(async (_, ct) =>
-            {
-                await Task.Delay(1, ct);
-                return ConnectivityState.Ready;
-            });
+            var transportFactory = new TestSubchannelTransportFactory();
             var clientChannel = CreateConnectionManager(loggerFactory, resolver, transportFactory, new[] { new RoundRobinBalancerFactory() });
             // Configure balancer similar to how GrpcChannel constructor does it
             clientChannel.ConfigureBalancer(c => new ChildHandlerLoadBalancer(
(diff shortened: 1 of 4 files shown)


mot256:
Thanks @JamesNK (I feel honored that you, of all people, took the time to look at the PR)
It is kinda embarrassing if it is that simple... I will take it for a test run and see. 
I have 1 question:

Why would we not have consistency between a resolver returning 0 addresses vs. when it returns 1 or more addresses? e.g. If we were to write a resolver that monitors Kubernetes service endpoints and return 0 addresses for a service with no PODs, why would we not want to drop the sub-channel(s), where as if we would go from 2 to 1 addresses, the load balancer would close the sub-channel of the address that is not part of the resolved addresses.

JamesNK:
Hey, I created the bug. Your tests and changes were very helpful in narrowing down to a root cause.

I'll look into the inconsistency you mentioned.

mot256:
@JamesNK Am I correct in assuming that I build my own package with this change (as I did with my own changes)? Or am I missing some CI artifact that may hold the nuget package build? Just asking for future reference

JamesNK:
You would need to build it yourself. There is a gRPC NuGet feed but it only builds committed code - https://github.com/grpc/grpc-dotnet#grpc-nuget-feed

test/Grpc.Net.Client.Tests/Balancer/ConnectionManagerTests.cs
```diff
@@ -327,18 +327,139 @@ public async Task UpdateAddresses_ConnectIsInProgress_InProgressConnectIsCancele
...
+
+            GrpcChannelOptions channelOptions = new GrpcChannelOptions();
+            channelOptions.ServiceConfig = new ServiceConfig()
+            {
+                LoadBalancingConfigs = { new RoundRobinConfig() }
+            };
+
+            var transportFactory = new TestSubchannelTransportFactory(async (_, ct) =>
+            {
+                await Task.Delay(1, ct);
```
BrennanConroy, line 349: Does this need to be a delay? Can you just `Task.FromResult(...)` or even `Task.Yield()` if you really need to go async.

BrennanConroy (approved):
Nice simple change 😃 

The one concern I have is around thread-safety, I see that `ChildHandlerLoadBalancer` calls `UpdateChannelState` outside of a lock. This wouldn't be new in this PR, but it'd be nice to confirm this is safe.

JamesNK:
The update is initiated in a lock here: https://github.com/grpc/grpc-dotnet/blob/1d7f98834cd291955e4922d5835a949e63de04c6/src/Grpc.Net.Client/Balancer/Internal/ConnectionManager.cs#L190-L194

PR #1855, opened 2022-08-20
Change nullable GetValueOrDefault to Value

JamesNK:
`GetValueOrDefault` was used to avoid null check in [`Nullable<T>.Value`](https://github.com/dotnet/runtime/blob/5958eeb8a0ace432005c4e9fa47e14114e032688/src/libraries/System.Private.CoreLib/src/System/Nullable.cs#L36-L46). JIT in .NET 5/6/7 optimizes this away.

There probably is a very small perf hit on older platforms (.NET Framework/.NET Core 3), but IMO the easier-to-read and safer code is better.

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
@@ -19,6 +19,7 @@
 using System.Buffers;
 using System.Buffers.Binary;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.IO.Pipelines;
 using System.Runtime.CompilerServices;
 using Grpc.Core;
@@ -39,6 +40,7 @@ namespace Grpc.AspNetCore.Server.Internal
 
         public PipeWriter ResponseBufferWriter { get; set; } = default!;
 
+        [MemberNotNullWhen(true, nameof(_payloadLength))]
         private bool DirectSerializationSupported => _compressionProvider == null && _payloadLength != null;
 
         public HttpContextSerializationContext(HttpContextServerCallContext serverCallContext)
@@ -186,7 +188,7 @@ namespace Grpc.AspNetCore.Server.Internal
                     }
                     else
                     {
-                        GrpcServerLog.SerializedMessage(_serverCallContext.Logger, _serverCallContext.ResponseType, _payloadLength.GetValueOrDefault());
+                        GrpcServerLog.SerializedMessage(_serverCallContext.Logger, _serverCallContext.ResponseType, _payloadLength.Value);
                     }
                     break;
                 default:
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
@@ -19,7 +19,6 @@
 using System.Buffers;
 using System.Buffers.Binary;
 using System.Diagnostics;
-using System.Diagnostics.CodeAnalysis;
 using System.IO.Pipelines;
 using System.Runtime.CompilerServices;
 using Grpc.Core;
@@ -40,8 +39,20 @@ namespace Grpc.AspNetCore.Server.Internal
 
         public PipeWriter ResponseBufferWriter { get; set; } = default!;
 
-        [MemberNotNullWhen(true, nameof(_payloadLength))]
-        private bool DirectSerializationSupported => _compressionProvider == null && _payloadLength != null;
+        private bool IsDirectSerializationSupported(out int payloadLength)
+        {
+            // Message can be written directly to the buffer if:
+            // - Its length is known.
+            // - There is no compression.
+            if (_payloadLength != null)
+            {
+                payloadLength = _payloadLength.Value;
+                return _compressionProvider == null;
+            }
+
+            payloadLength = 0;
+            return false;
+        }
 
         public HttpContextSerializationContext(HttpContextServerCallContext serverCallContext)
         {
@@ -132,13 +143,11 @@ namespace Grpc.AspNetCore.Server.Internal
             {
                 case InternalState.Initialized:
                     // When writing directly to the buffer the header with message size needs to be written first
-                    if (DirectSerializationSupported)
+                    if (IsDirectSerializationSupported(out var payloadLength))
                     {
-                        Debug.Assert(_payloadLength != null, "A payload length is required for direct serialization.");
+                        EnsureMessageSizeAllowed(payloadLength);
 
-                        EnsureMessageSizeAllowed(_payloadLength.Value);
-
-                        WriteHeader(ResponseBufferWriter, _payloadLength.Value, compress: false);
+                        WriteHeader(ResponseBufferWriter, payloadLength, compress: false);
                     }
 
                     _state = InternalState.IncompleteBufferWriter;
@@ -153,9 +162,20 @@ namespace Grpc.AspNetCore.Server.Internal
 
         private IBufferWriter<byte> ResolveBufferWriter()
         {
-            return DirectSerializationSupported
-                ? (IBufferWriter<byte>)ResponseBufferWriter
-                : _bufferWriter ??= new ArrayBufferWriter<byte>();
+            if (IsDirectSerializationSupported(out var payloadLength))
+            {
+                return ResponseBufferWriter;
+            }
+            else if (_bufferWriter == null)
+            {
+                // Initialize buffer writer with exact length if available.
+                // ArrayBufferWriter doesn't allow zero initial length.
+                _bufferWriter = payloadLength > 0
+                    ? new ArrayBufferWriter<byte>(payloadLength)
+                    : new ArrayBufferWriter<byte>();
+            }
+
+            return _bufferWriter;
         }
 
         private void EnsureMessageSizeAllowed(int payloadLength)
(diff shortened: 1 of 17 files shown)


src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
```diff
@@ -39,6 +40,7 @@ internal sealed class HttpContextSerializationContext : SerializationContext
 
         public PipeWriter ResponseBufferWriter { get; set; } = default!;
 
+        [MemberNotNullWhen(true, nameof(_payloadLength))]
```
gfoidl, line 43:
```suggestion
```
It's for nullable _reference_ types, so for `_payloadLength` (as valuetype) it has no effect.

src/Grpc.AspNetCore.Server/Internal/PipeExtensions.cs
```diff
@@ -361,11 +361,11 @@ public static async ValueTask<T> ReadSingleMessageAsync<T>(this PipeReader input
             }
         }
 
-        private static bool TryReadMessage(ref ReadOnlySequence<byte> buffer, HttpContextServerCallContext context, [NotNullWhen(true)] out ReadOnlySequence<byte>? message)
+        private static bool TryReadMessage(ref ReadOnlySequence<byte> buffer, HttpContextServerCallContext context, [NotNullWhen(true)] out ReadOnlySequence<byte> message)
```
gfoidl, line 364:
```suggestion
        private static bool TryReadMessage(ref ReadOnlySequence<byte> buffer, HttpContextServerCallContext context, out ReadOnlySequence<byte> message)
```
a) valuetype
b) nullability got removed, so it's always non-null

src/Grpc.Net.Client/Internal/GrpcCallSerializationContext.cs
```diff
@@ -35,6 +36,7 @@ internal sealed class GrpcCallSerializationContext : SerializationContext, IBuff
         private int? _payloadLength;
         private ICompressionProvider? _compressionProvider;
 
+        [MemberNotNullWhen(true, nameof(_payloadLength))]
```
gfoidl, line 39:
```suggestion
```
See above.

src/Grpc.Net.Client/Internal/GrpcCallSerializationContext.cs
```diff
@@ -198,7 +200,7 @@ private IBufferWriter<byte> ResolveBufferWriter()
             {
                 if (_buffer == null)
                 {
-                    _buffer = ArrayPool<byte>.Shared.Rent(GrpcProtocolConstants.HeaderSize + _payloadLength.GetValueOrDefault());
+                    _buffer = ArrayPool<byte>.Shared.Rent(GrpcProtocolConstants.HeaderSize + _payloadLength.Value);
```
gfoidl, line 203: Here it looks like the [JIT can't follow](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAOgCEBXAMyZilwG4BYAKF4EsAdhjaDsAG3K4YEmABNKAJnIBxKAAcwAYQniAym34T+AL2wZ+EQVqsiEGXgG9e5V+XVR+AN3MxykQVwMciFggAkZOTY9Uz8AXnIqRW4eN3dPHxEQ4QB+cgB9dWwAT3EIbDkAGRhBAHMMAAsUtI9vX3IIYAArGDBg/Mh8DxhcXEtBAAUoCC9+KKhmt1bMv2BikQBtAF0C4GZWBZcljPaAQSgoEsYWNgB1TxEoAB41kQA+Xf27h7ZF1yPXMt2sAIBBJAARfiwPoGTzGMwWKx6BjqdTQEQKOIfAYQIawUbjKYzOZschxBKCBjiSQAMhpBSKpXKVRq9Qa5AAhBSqeI/uQAek2llOj0+uQAEojMFeGDXA73fiPAAUAEoBc5Umk3PwmOQlZDoRhYUZxKZzONkaj0fI1ZqteQNfb7Tq9fk9jcoGTudTbU6tY6/U63V9PQlzpdihNQeIXusYG8aHoGthYHIaJLhEqIhVorFyABqBklMoVap1Ro0ABqEgYMBVfL9AF9eAK/cQAOzkRr8TittzNu1amDiaQhXVK4MehWPL3kSk+vuuAOB8gAelX5AAkoJFSa8+6DuQAO4/T1HxXsmAIbBi8Ssxpj8jYHz8cTYYB3miLrXr8jhq4htOpJyBAIyCAA5MEugQEe5AmGwEDZLuEjkHe5YNF+g5+pO8qnmSRZMqW97sh8pDfn6eSCDAsH/sUcrfIqbCxu8E6MiWLLoVWNZ1uRTogHO1F/hcAFTqezHxqqDb9i2WFuB2nyiYxhyDgOaQCkCwrdL0wSSrg0qyoBp5UKq6rfi6+pQtpxrwuaSIomiUAYr6TrLoG5k4aS5Jzjyzkrq5K6uB5oZCRGUZguJCZJim8jpjUGBZpEubwQWBHsWWbI0CoMAYNW4i1gA8lA4IwEw2BUvFKr1rxrgDtVlCdt2vayTV37DqO7kHgxM5efO4i+f6dU+J6bHMulD4JIUxajcRmXZblBVFSVZXiBVUn2nVv7bshprJZ1Z54eeD5XjewRoWyj7Ptgr7vp+G0brR9FQEBnogWBkFPtSMFwQhSEWChZ0VnVQXPfhI1Eeh5CkXVaSUYJD2GUpEVKmDHFsv1AXkPxVE0cJdEI48SNVc15C1cT8nA6efIDo2QA===) -- shown by `L00f5: call 0x00007ffc9e4254e8` which is the exception being thrown.

```suggestion
                    _buffer = ArrayPool<byte>.Shared.Rent(GrpcProtocolConstants.HeaderSize + _payloadLength.GetValueOrDefault());
```

src/Grpc.Net.Client/Internal/GrpcCallSerializationContext.cs
```diff
@@ -208,7 +210,7 @@ private IBufferWriter<byte> ResolveBufferWriter()
                 // Initialize buffer writer with exact length if available.
                 // ArrayBufferWriter doesn't allow zero initial length.
                 _bufferWriter = _payloadLength > 0
-                    ? new ArrayBufferWriter<byte>(_payloadLength.GetValueOrDefault())
+                    ? new ArrayBufferWriter<byte>(_payloadLength.Value)
                     : new ArrayBufferWriter<byte>();
```
gfoidl, line 214: Here the JIT can't prove that `_payloadLength` has a non-null value, as the check for `DirectSerializationSupported` could return `false` if `__compressionProvider != null` too.
I'd write this as
```c#
            var payloadLength = _payloadLength.GetValueOrDefault();
            
            // Initialize buffer writer with exact length if available.
            // ArrayBufferWriter doesn't allow zero initial length.
            _bufferWriter = payloadLength > 0
                ? new ArrayBufferWriter<byte>(payloadLength)
                : new ArrayBufferWriter<byte>();
```

Both suggestions lead to the seconds method in the above sharplab, no exception-case here.

src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
```diff
@@ -82,7 +82,7 @@ protected RetryCallBase(GrpcChannel channel, Method<TRequest, TResponse> method,
                 _ctsRegistration = RegisterRetryCancellationToken(options.CancellationToken);
             }
 
-            var deadline = Options.Deadline.GetValueOrDefault(DateTime.MaxValue);
+            var deadline = Options.Deadline ?? DateTime.MaxValue;
```
gfoidl, line 85:
```suggestion
            var deadline = Options.Deadline.GetValueOrDefault(DateTime.MaxValue);
```
The new code [produces](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunGBLAHYYGAWQAUgjAH4GCAJQMAvAD5ZDKTJSkA3Fx70GkhgDkJQmfKWqEAOgDiMDADVsAGwCuMAPJQAIjAAzbHdXDDEtOV1qAF8gA=) more IL and machine code. I'd keep the previous state.

gfoidl (commented):
> `GetValueOrDefault` was used to avoid null check in [`Nullable<T>.Value`](https://github.com/dotnet/runtime/blob/5958eeb8a0ace432005c4e9fa47e14114e032688/src/libraries/System.Private.CoreLib/src/System/Nullable.cs#L36-L46). JIT in .NET 5/6/7 optimizes this away.

To state more precisely: only when the JIT can prove that there was a `HasValue` or `!= null` check before.

src/Grpc.Net.Client/Internal/GrpcCallSerializationContext.cs
JamesNK: if `_payloadLength` is greater than zero then it can't be null, right?

src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
JamesNK: It stops the nullable analyzer from complaining about `Nullable<T>.Value`.

src/Grpc.Net.Client/Internal/GrpcCallSerializationContext.cs
gfoidl: Yes.

If it's null, then the default value is `0`, so the ternary does the right thing.

src/Grpc.AspNetCore.Server/Internal/HttpContextSerializationContext.cs
gfoidl: Ah, correct.

I tried it with VS 17.3.1 and didn't see anything. Now tried it again, and see what you say.
Cool, so please leave it as is.

Issue #1861, opened 2022-08-26
Consider setting SocketsHttpHandler.PooledConnectionLifetime in factory


JamesNK:
The gRPC client factory now reuses channels, and they have a singleton lifetime. That means the HttpHandler used to create the factory lasts forever and isn't periodically replaced by HttpClientFactory. DNS changes won't be taken account of.

We don't want to replace the channel because it has state, and implementing logic to replace the handler underneath the channel would be complex.

Defaulting the handler created by the factory to have a pooled connection lifetime would put a max time limit on connection lifetime and they can be recreated to get DNS changes.

[SocketsHttpHandler.PooledConnectionLifetime](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.pooledconnectionlifetime?view=net-6.0)

PR #1862, opened 2022-08-26
Configure pooled connection lifetime in client factory

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1861

### DIFF
--- a/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
+++ b/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
@@ -344,7 +344,8 @@ namespace Microsoft.Extensions.DependencyInjection
 #if NET5_0_OR_GREATER
                         if (handler is SocketsHttpHandler socketsHttpHandler)
                         {
-                            // A channel is created once per client and the primary handler never changes.
+                            // A channel is created once per client, lives forever, and the primary handler never changes.
+                            // It's possible that long lived connections cause the client to miss out on DNS changes.
                             // Replicate the core benefit of a handler lifetime (periodic connection recreation)
                             // by setting PooledConnectionLifetime to handler lifetime.
                             socketsHttpHandler.PooledConnectionLifetime = options.HandlerLifetime;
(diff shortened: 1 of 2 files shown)


src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
```diff
@@ -334,13 +334,23 @@ public static IHttpClientBuilder AddGrpcClient<TClient>(this IServiceCollection
...
                         // This will throw in .NET Standard 2.0 with a prompt that a user must set a handler.
                         // Because it throws it should only be called in PostConfigure if no handler has been set.
                         var handler = HttpHandlerFactory.CreatePrimaryHandler();
+#if NET5_0_OR_GREATER
+                        if (handler is SocketsHttpHandler socketsHttpHandler)
+                        {
+                            // A channel is created once per client, lives forever, and the primary handler never changes.
+                            // It's possible that long lived connections cause the client to miss out on DNS changes.
+                            // Replicate the core benefit of a handler lifetime (periodic connection recreation)
+                            // by setting PooledConnectionLifetime to handler lifetime.
```
SamGuoMsft, line 350: Thanks @JamesNK to address this issue based on our offline discussion. Leave few comments here:
1. Since we cache channel per typed/named client (no matter the client is transient or singleton), and use PooledConnectionLifetime to control handler lifetime, we are only using httpclientfactory as configuration API, the httpclientfactory handler pool is redundant in this scenario. I think we should work with[ runtime team to make httpclientfactory just as configuration API](https://github.com/dotnet/runtime/issues/74508).
2. With caching channel and use PooledConnectionLifetime, would you mind to share more details on why we DI GrpcClientFactory as transient rather than singleton (same as DefatultHttpClientFactory)? 
` services.TryAddTransient<GrpcClientFactory, DefaultGrpcClientFactory>();`

JamesNK: > With caching channel and use PooledConnectionLifetime, would you mind to share more details on why we DI GrpcClientFactory as transient rather than singleton (same as DefatultHttpClientFactory)?
services.TryAddTransient<GrpcClientFactory, DefaultGrpcClientFactory>();

So the current `IServiceProvider` scope is injected into the factory.

captainsafia (approved):

Issue #1832, opened 2022-08-01
Grpc.Net >= 2.44.0 does not pick up HTTPS_PROXY environment variable


odin568:
### What version of gRPC and what language are you using?
Issue detected initally with 2.46.0 but started with 2.44.0. Behavior was fine until 2.43.0 (tested from 2.38.0 on)

### What operating system (Linux, Windows,...) and version?
Linux container - original Microsoft ASP.NET runtime container (latest)

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
```.NET SDK (reflecting any global.json):
 Version:   6.0.302
 Commit:    c857713418
Runtime Environment:
 OS Name:     ubuntu
 OS Version:  22.04
 OS Platform: Linux
 RID:         ubuntu.22.04-x64
 Base Path:   /usr/share/dotnet/sdk/6.0.302/
global.json file:
  Not found
Host:
  Version:      6.0.7
  Architecture: x64
  Commit:       0ec02c8c96
.NET SDKs installed:
  6.0.302 [/usr/share/dotnet/sdk]
.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.7 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.7 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
```

### What did you do?
Issue came up when I upgraded Google.Cloud.PubSub.V1 from 2.10.0 to 3.0.0.
This came with a change to replace the default Grpc.Core implementation with Grpc.Net.
Thereby I first got connection issues as the `grpc_proxy` environment variable is of course not used anymore.
I then changed it to `HTTPS_PROXY` (and `HTTP_PROXY`) and expected it to work but anyhow it was just stuck/hanging until the task was forcably canceled.

I then found a related issue in the google-cloud-dotnet repository here: https://github.com/googleapis/google-cloud-dotnet/issues/8834

The discussions here are quite lengthy as we tried out many many variants until we came to the point to see that the issue seems to be in Grpc.Net and that the environment variable was properly taken into account until 2.43.0 and ignored with 2.44.0 on.
Interestingly, folks at the google repo were not able to reproduce it, it worked on their side but there is somebody else reporting the same problem. So it seems to be a special case in some configuration - where we don't know the original reason for it.

Our testcode was mainly this at the end:
```csharp
static void TestGrpcNetClient_SocketHttpHandler() =>
[...]

PR #1863, opened 2022-08-28
Skip active connections when proxy is configured

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1832

tl;dr; HTTPS proxy tunnels a request through another request. This is different from the standard behavior to open a socket. Fallback to passive connectivity when HTTPS proxy tunnel is present.


JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/1832

jtattermusch (approved):
I think the idea is looking good.
- ideally @captainsafia would review as well
- can you please update the PR description with the summary of impact on loadbalancing when proxy is set (what will be the behavior observed by user?).

captainsafia (approved):

fmg-lydonchandra:
Thanks for this, I learnt that Grpc proxy works only when Grpc service is running as HTTPS.
I am proxying Grpc Client call through Charlesproxy/Fiddler.

Issue #1864, opened 2022-08-30
implicit behavior if continuation thrown exception synchronously


setood:
if continuation thrown exception synchronously (if deadline not UTC DateTime, for example ), then HandleResponse never called

https://github.com/grpc/grpc-dotnet/blob/2116d1efce933b1185baa1c577cc09b6a9a6c88a/examples/Interceptor/Client/ClientLoggerInterceptor.cs#L39-L50

JamesNK:
I improved the client logger interceptor to handle these errors: https://github.com/grpc/grpc-dotnet/pull/1868

PR #1868, opened 2022-08-30
Log sync errors in client logger interceptor example

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1864

### DIFF
--- a/examples/Interceptor/Client/ClientLoggerInterceptor.cs
+++ b/examples/Interceptor/Client/ClientLoggerInterceptor.cs
@@ -33,7 +33,15 @@ namespace Client
             LogCall(context.Method);
             AddCallerMetadata(ref context);
 
-            return continuation(request, context);
+            try
+            {
+                return continuation(request, context);
+            }
+            catch (Exception ex)
+            {
+                LogError(ex);
+                throw;
+            }
         }
 
         public override AsyncUnaryCall<TResponse> AsyncUnaryCall<TRequest, TResponse>(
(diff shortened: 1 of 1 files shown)


examples/Interceptor/Client/ClientLoggerInterceptor.cs
```diff
@@ -132,5 +165,16 @@ private async Task<TResponse> HandleResponse<TResponse>(Task<TResponse> t)
...
         }
+
+        private static void LogError(Exception ex)
+        {
+            // Log error to the console.
+            // Note: Configuring .NET Core logging is the recommended way to log errors
+            // https://docs.microsoft.com/aspnet/core/grpc/diagnostics#grpc-client-logging
+            var initialColor = Console.ForegroundColor;
+            Console.ForegroundColor = ConsoleColor.Red;
+            Console.WriteLine($"Call error: {ex.Message}");
```
captainsafia, line 176:
```suggestion
            Console.WriteError($"Call error: {ex.Message}");
```

Or is there no particular need to specifically map to stderr here?

captainsafia (approved):

examples/Interceptor/Client/ClientLoggerInterceptor.cs
```diff
@@ -132,5 +165,16 @@ private async Task<TResponse> HandleResponse<TResponse>(Task<TResponse> t)
             headers.Add("caller-machine", Environment.MachineName);
             headers.Add("caller-os", Environment.OSVersion.ToString());
         }
+
+        private static void LogError(Exception ex)
+        {
+            // Log error to the console.
+            // Note: Configuring .NET Core logging is the recommended way to log errors
+            // https://docs.microsoft.com/aspnet/core/grpc/diagnostics#grpc-client-logging
```
BrennanConroy, line 173: Why not configure the sample to use the logging primitives then? Looks like it should be really easy.

PR #1877, opened 2022-09-09
Change examples to use WebApplication V2

JamesNK:
Halfway done 😓


examples/Container/Frontend/Program.cs
```diff
@@ -16,32 +16,71 @@
...
+    var channel = GrpcChannel.ForAddress(backendUrl, new GrpcChannelOptions
+    {
+        Credentials = ChannelCredentials.Insecure,
+        ServiceProvider = services
+    });
+
+    return channel;
+});
+
+SetupReportingServices(builder.Services);
```
captainsafia, line 46: Super nit: I'd almost prefer (a) having another file to hold the extension method or (b) doing the registrations inline to this since it's not a common pattern. I know it's just an example but people tend to copy-paste from these a lot.

examples/Frameworker/Server/Properties/launchSettings.json
```diff
@@ -0,0 +1,12 @@
+﻿{
+  "profiles": {
+    "Server": {
+      "commandName": "Project",
+      "launchBrowser": false,
+      "applicationUrl": "https://localhost:5001",
```
captainsafia, line 6: Does this (and elsewhere) need to include a non-SSL url as well for completeness?

JamesNK: No, the client apps all use TLS.

examples/Container/Frontend/Program.cs
JamesNK: Done

captainsafia (approved):

Issue #1881, opened 2022-09-12
Response is not compressed when Grpc-Accept-Encoding header contains spaces between values


Mitch528:
### What version of gRPC and what language are you using?
Grpc.*: 2.4.7
C#

### What operating system (Linux, Windows,...) and version?
macOS

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.NET SDK (reflecting any global.json):
 Version:   6.0.400

### What did you do?
Enable gzip compression on Objective-C gRPC client

### What did you expect to see?
Response to be compressed

### What did you see instead?
Response was not compressed

### Anything else we should know about your project / environment?

The issue seems to occur due to the Objective-C client having a space between the compression algorithms in the `Grpc-Accept-Encoding` header.

Example Obj-C header:
```
Grpc-Accept-Encoding: identity, deflate, gzip
```

The following middleware appears to resolve the issue:

```cs
app.Use((context, next) =>
{
    context.Request.Headers["Grpc-Accept-Encoding"] = context.Request.Headers["Grpc-Accept-Encoding"]
        .ToString().Replace(" ", string.Empty);

    return next(context);
});
```

JamesNK:
Thanks, the fix is simple: https://github.com/grpc/grpc-dotnet/pull/1882

Continue to use your work-around until a new version is released. That should be in a couple of months.

PR #1882, opened 2022-09-13
Fix detecting message accept encoding with whitespace

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1881

### DIFF
--- a/test/FunctionalTests/Server/CompressionTests.cs
+++ b/test/FunctionalTests/Server/CompressionTests.cs
@@ -430,11 +430,12 @@ namespace Grpc.AspNetCore.FunctionalTests.Server
             AssertHasLogRpcConnectionError(StatusCode.Internal, "Request did not include grpc-encoding value with compressed message.");
         }
 
-        [TestCase("gzip", true)]
+        [TestCase("gzip", "gzip", true)]
+        [TestCase("gzip", "identity, gzip", true)]
 #if NET6_0_OR_GREATER
-        [TestCase("deflate", false)]
+        [TestCase("deflate", "deflate", false)]
 #endif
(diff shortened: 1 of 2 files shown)


test/FunctionalTests/Server/CompressionTests.cs
```diff
@@ -430,11 +430,12 @@ public async Task SendCompressedMessageWithoutEncodingHeader_ServerErrorResponse
             AssertHasLogRpcConnectionError(StatusCode.Internal, "Request did not include grpc-encoding value with compressed message.");
         }
 
-        [TestCase("gzip", true)]
+        [TestCase("gzip", "gzip", true)]
+        [TestCase("gzip", "identity, gzip", true)]
 #if NET6_0_OR_GREATER
-        [TestCase("deflate", false)]
+        [TestCase("deflate", "deflate", false)]
```
captainsafia, line 436:
```suggestion
        [TestCase("deflate ", "deflate", false)]
```

Should this test case include trailing whitespace as well just to cover the scenario where a single algo is used?

captainsafia (approved):
One small non-block comment on test coverage.

test/FunctionalTests/Server/CompressionTests.cs
```diff
@@ -430,11 +430,12 @@ public async Task SendCompressedMessageWithoutEncodingHeader_ServerErrorResponse
             AssertHasLogRpcConnectionError(StatusCode.Internal, "Request did not include grpc-encoding value with compressed message.");
         }
 
-        [TestCase("gzip", true)]
+        [TestCase("gzip", "gzip", true)]
+        [TestCase("gzip", "identity, gzip", true)]
```
JamesNK, line 434:
```suggestion
        [TestCase("gzip", "identity, gzip", true)]
        [TestCase("gzip", "gzip ", true)]
```

PR #1887, opened 2022-09-21
Update transcoder example

JamesNK:
* Server SDK version
* Transcoder package versions
* Rename some transcoder UI elements

### DIFF
--- a/build/get-dotnet.sh
+++ b/build/get-dotnet.sh
@@ -18,7 +18,7 @@ ensure_dir() {
 # main
 
 # resolve SDK version
-sdk_version="7.0.100-rc.2.22472.26"
+sdk_version="7.0.100-rc.2.22477.23"
 
 # download dotnet-install.sh
 ensure_dir $OBJDIR
--- a/build/get-dotnet.sh
+++ b/build/get-dotnet.sh
@@ -18,7 +18,7 @@ ensure_dir() {
 # main
 
 # resolve SDK version
-sdk_version="7.0.100-rc.1.22424.4"
+sdk_version="7.0.100-rtm.22470.23"
 
 # download dotnet-install.sh
 ensure_dir $OBJDIR
--- a/build/get-dotnet.sh
+++ b/build/get-dotnet.sh
@@ -18,7 +18,7 @@ ensure_dir() {
 # main
 
 # resolve SDK version
-sdk_version="7.0.100-rtm.22470.23"
+sdk_version="7.0.100-rc.2.22472.26"
 
 # download dotnet-install.sh
 ensure_dir $OBJDIR
--- a/examples/Transcoder/Proto/greet.proto
+++ b/examples/Transcoder/Proto/greet.proto
@@ -35,11 +35,13 @@ service Greeter {
   }
 }
 
+// SayHello request message.
 message HelloRequest {
   // Name to say hello to.
   string name = 1;
 }
 
+// SayHelloStream request message.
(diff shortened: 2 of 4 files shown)


captainsafia (approved):

exquirentibus:
I enjoyed your video and am trying JSON Transcoding so that I can run gRPC Server using HTTP/1 on Windows Server 2016 (which apparently can't support HTTP/2 gRPC).

I had to add

using Microsoft.Extensions.Logging;
using System;
using System.Threading.Tasks;

to GreeterService.cs  and then ran into an issue with

var builder = WebApplication.CreateBuilder(args);

in Program.cs that I couldn't easily fix as adding the VS2022 suggested

using Microsoft.AspNetCore.Builder;

Caused a problem with

builder.Services.AddGrpc().AddJsonTranscoding();

'IServiceCollection' does not contain a definition for 'AddGrpc'

But the concept of getting the changes necessary for JSON Transcoding are pretty simple and I was able to get it Almost completed in my gRPC Server application.

However .. I ran into one problem I can't figure out?

I'm using a GrpcShared shared project class library to store my Proto file between a Console client, Maui client and WPF client.

It was working perfectly before I added the changes for JSON Transcoding.

The GrpcShared class library project just had the example greet.proto file in the first level.  No Protos folder.

I added a google folder and an api folder inside and placed annotations.proto and http.proto (the latest versions I believe) inside the google/api folders.

When I try to compile the GrpcShared class project, I get an error of:

Error    CS0234    The type or namespace name 'Api' does not exist in the namespace 'Google' (are you missing an assembly reference?)    GrpcShared    G:\Projects\GrpcService\GrpcShared\obj\Debug\net7.0\Greet.cs    33    Active

which is of course the generated file.

The GrpcShared project definition is as follows:

```
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net7.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>
 
    <ItemGroup>
    <PackageReference Include="CommunityToolkit.Mvvm" Version="8.1.0" />
    <PackageReference Include="Google.Protobuf" Version="3.22.1" />
    <PackageReference Include="Grpc.Net.Client" Version="2.52.0" />
    <PackageReference Include="Grpc.Tools" Version="2.53.0">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
    </PackageReference>
    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
  </ItemGroup>
 
  <ItemGroup>
    <Compile Remove="HelloReply.cs" />
  </ItemGroup>

  <ItemGroup>
    <SupportedPlatform Include="browser" />
    <Protobuf Include="greet.proto" />
  </ItemGroup>
</Project>
```


I'm a bit stumped on how to proceed and had hoped to get some clues from your Example project.

One line in your project definition perplexes me?

`   <Protobuf Include="..\Proto\greet.proto" GrpcServices="Server" Link="Protos\greet.proto" />`

as there Is no 'Proto' folder in the project?  Only the 'Protos' folder and the above line refers to .. Both?

Any chance of a working gRPC Transcoding example that uses a Shared Project Class?  Hopefully with the proto files at the first level (as the class is Only going to contain proto files)?

Or could you point me in the right direction for what must be an obvious mental failure on my part?

PR #1890, opened 2022-09-23
Use the extending class name when throwing ObjectDisposedException

ellull:
Don't use the DnsResolver class name but the extending class name when throwing the ObjectDisposedException in the PollingResolver.Resolve() method


JamesNK (approved):

JamesNK:
Good spotting. Thanks for the PR!

ellull:
Thanks @JamesNK.

This is my first contribution to this project and I don't know if I have to do something else now that the PR has been approved.

PR #1901, opened 2022-10-05
Fix grpc benchmark cancellation

BrennanConroy:
Methods on `CallOptions` return a new instance of the options struct.


BrennanConroy:
Still didn't help even if it is correct...

JamesNK (approved):

Issue #1906, opened 2022-10-09
dotnet-grpc fails when running commands using .NET 7 SDK


UnstoppableMango:
### What version of gRPC and what language are you using?
I am using C#. I'm not actually sure what version of gRPC I'm using. Packages are 2.49.*

```shell
> dotnet grpc --version
2.49.0+2385e78df8d19f173788e16851d18db96ec047da
```

### What operating system (Linux, Windows,...) and version?
I am using Linux.
5.19.13-arch1-1 x86_64

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.NET SDK:
 Version:   7.0.100-rc.1.22431.12
 Commit:    f1cf61e1c0

### What did you do?
I ran into this after updating to the .NET 7 RC and attempted to run `dotnet grpc refresh` on a project with external proto references. It's easily reproducible with a new project.

- Create a new project. `dotnet new classlib`
- Can add a proto file and a `<Protobuf>` item to the project, but it's not necessary to reproduce the exception
- Run `dotnet grpc list`

### What did you expect to see?
A list of the `.proto` files referenced by my project

### What did you see instead?
```
Unhandled exception: Microsoft.Build.Exceptions.InvalidProjectFileException: SDK Resolver Failure: "The SDK resolver "Microsoft.DotNet.MSBuildWorkloadSdkResolver" failed while attempting to resolve the SDK "Microsoft.NET.Sdk". Exception: "System.IO.FileNotFoundException: Could not load file or assembly 'System.Runtime, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
```

```
Unhandled exception: Microsoft.Build.Exceptions.InvalidProjectFileException: SDK Resolver Failure: "The SDK resolver "Microsoft.DotNet.MSBuildWorkloadSdkResolver" failed while attempting to resolve the SDK "Microsoft.NET.Sdk". Exception: "System.IO.FileNotFoundException: Could not load file or assembly 'System.Runtime, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.

File name: 'System.Runtime, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
[...]

PR #1914, opened 2022-10-17
Upgrade dotnet-grpc dependencies

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1906

I updated the MSBuild packages used to locate and load projects, and that fixed the commend line tool being used with .NET 7 apps.

I also updated some other packages used by the tool in this PR. System.Comandline required some reactions (as usual...)


JamesNK:
A good idea but `dotnet-grpc` tool isn't high use. TBH I wouldn't want them to take up their time testing it.

PR #1921, opened 2022-10-20
Add option to full disable client balancing

kalduzov:
Hello.

Our company uses its own solution for balancing grpc requests, which is based on a custom handler. Similar to how it is now implemented in the current version of the library.

When version 2.44.0 was released, we faced the problem that our solution was completely overwritten by the current implementation of client balancing. There are several problems for us here:

1. It is impossible to disable the balancing handler in the current version of the library, but it interferes with our handler.

2. We could not make an adequate implementation based on the proposed api, taking into account our needs, because we use several sets of pickers and complex traffic balancing algorithms. We can't embed normally in the current API, because it is based on base classes, not interfaces.

3. We have balancing for HTPP and GPRC clients based on the same code - custom http handler. The only difference is how we find out which address is used as a balancing. We will find out the real address right before sending the request. We resolve the address we need directly in the overridden Send method. The current balancing implementation tries to connect to an address that may not exist in reality, and at this point the address liveness check immediately fails.

The solution that we propose is to add an option to the channel that would completely disable work with internal balancing, i.e. did not use these mechanisms at all and acted in the old style. Now there is only one option that disables the balancer itself, but does not disable address resolution and connection liveness check.

We also believe that experimental functionality should be disabled at runtime, not at compile time.

### DIFF
--- a/src/Grpc.Net.Client/GrpcChannel.cs
+++ b/src/Grpc.Net.Client/GrpcChannel.cs
@@ -41,6 +41,7 @@ public sealed class GrpcChannel : ChannelBase, IDisposable
 {
     internal const int DefaultMaxReceiveMessageSize = 1024 * 1024 * 4; // 4 MB
 #if SUPPORT_LOAD_BALANCING
+    private readonly bool _useLoadBalancing;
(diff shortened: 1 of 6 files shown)


kalduzov:
Hi @JamesNK . What do you think about my MR?

Issue #1923, opened 2022-10-20
Failed to create a gRpc duplex stream when using MAUI Android


rodrigovaras:
### What version of gRPC and what language are you using?
Using C# and version:
    <PackageReference Update="Grpc.Net.Client" Version="2.49.0" />

### What operating system (Linux, Windows,...) and version?
Running on Android 12

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.NET SDK:
 Version:   7.0.100-rc.2.22477.23
 Commit:    0a5360315a

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22000
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.100-rc.2.22477.23\

Host:
  Version:      7.0.0-rc.2.22472.3
  Architecture: x64
  Commit:       550605cc93

### What did you do?
Connect to a GRPC Servce on the cloud

### What did you expect to see?
The await call succeed

### What did you see instead?
Call get stuck here:
        var responseMetadata = await enterMapDuplex.ResponseHeadersAsync;

here are the logs until it get stuck:
04:17:39.043 msg:Starting gRPC call. Method type: 'DuplexStreaming', URI: 'https://query.sa.uksouth.mixedreality.azure.com/com.azure.mixedreality.sa.locoptquery.MapQueryPlanner/EnterMap'. level:Debug name:Grpc.Net.Client.Internal.GrpcCall [GrpcMethodType=DuplexStreaming],[GrpcUri=/com.azure.mixedreality.sa.locoptquery.MapQueryPlanner/EnterMap]
04:17:39.124 msg:Subchannel id '1' created with addresses: query.sa.uksouth.mixedreality.azure.com:443 level:Debug name:Grpc.Net.Client.Balancer.Subchannel
04:17:39.203 msg:Channel picker updated. level:Debug name:Grpc.Net.Client.Balancer.Internal.ConnectionManager
04:17:39.329 msg:Subchannel id '1' state changed to Connecting. Detail: 'Connection requested.'. level:Debug name:Grpc.Net.Client.Balancer.Subchannel
04:17:39.374 msg:Channel state updated to Connecting. level:Debug name:Grpc.Net.Client.Balancer.Internal.ConnectionManager
[...]

PR #1927, opened 2022-10-27
Detect native HttpHandler on Android

JamesNK:
Android uses a native HTTP handler if `HttpClientHandler` is specified. The native HTTP handler doesn't have enough HTTP/2 support for gRPC and can cause errors. Example: https://github.com/grpc/grpc-dotnet/issues/1923. We want Android to always use `SocketsHttpHandler`.

Detect if a native HttpHandler is being used on Android, and throw a friendly error message on Android if necessary.

Doc at https://aka.ms/aspnet/grpc/android will be added soon.

### DIFF
--- a/src/Grpc.Net.Client/GrpcChannel.cs
+++ b/src/Grpc.Net.Client/GrpcChannel.cs
@@ -92,7 +92,7 @@ public sealed class GrpcChannel : ChannelBase, IDisposable
 
     // Options that are set in unit tests
     internal ISystemClock Clock = SystemClock.Instance;
-    internal IOperatingSystem OperatingSystem = Internal.OperatingSystem.Instance;
+    internal IOperatingSystem OperatingSystem;
     internal IRandomGenerator RandomGenerator;
     internal bool DisableClientDeadline;
     internal long MaxTimerDueTime = uint.MaxValue - 1; // Max System.Threading.Timer due time
@@ -112,6 +112,7 @@ public sealed class GrpcChannel : ChannelBase, IDisposable
 
         Address = address;
         LoggerFactory = channelOptions.LoggerFactory ?? channelOptions.ResolveService<ILoggerFactory>(NullLoggerFactory.Instance);
+        OperatingSystem = channelOptions.ResolveService<IOperatingSystem>(Internal.OperatingSystem.Instance);
         RandomGenerator = channelOptions.ResolveService<IRandomGenerator>(new RandomGenerator());
         (HttpHandlerType, ConnectTimeout) = CalculateHandlerContext(channelOptions);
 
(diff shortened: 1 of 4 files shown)


src/Grpc.Net.Client/Internal/OperatingSystem.cs
```diff
@@ -23,16 +23,23 @@ namespace Grpc.Net.Client.Internal;
...
+    bool IsAndroid { get; }
 }
 
-internal class OperatingSystem : IOperatingSystem
+internal sealed class OperatingSystem : IOperatingSystem
 {
     public static readonly OperatingSystem Instance = new OperatingSystem();
 
     public bool IsBrowser { get; }
+    public bool IsAndroid { get; }
```
gfoidl, line 34: Will this kind of problem only ever exists on Android or may there be other devices (in the future) be too?

So maybe this should be named
```suggestion
    public bool HasNativeIncompatibleHttpHandler { get; }
```
to be able to cover more devices w/o adding more properties here?

JamesNK: Maybe. But because it’s an internal type we can rename it in the future.

src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -382,6 +383,31 @@ private HttpMessageInvoker CreateInternalHttpInvoker(HttpMessageHandler? handler
...
+                // GetHttpHandlerType recurses through DelegatingHandlers that may wrap the HttpClientHandler.
+                var httpClientHandler = HttpRequestHelpers.GetHttpHandlerType<HttpClientHandler>(handler);
+                
+                if (httpClientHandler != null && RuntimeHelpers.QueryRuntimeSettingSwitch("System.Net.Http.UseNativeHttpHandler", defaultValue: false))
+                {
+                    throw new InvalidOperationException("The channel configuration isn't valid on Android devices. " +
+                        "The channel is configured to use HttpClientHandler and Android's native HTTP/2 library. " +
+                        "gRPC isn't fully supported by Android's native HTTP/2 library and it can cause runtime errors. " +
+                        "To fix this problem, either configure the channel to use SocketsHttpHandler, or add " +
+                        "<UseNativeHttpHandler>true</UseNativeHttpHandler> to the app's project file. " +
```
BrennanConroy, line 406: Is this meant to be "false"?

src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -382,6 +383,31 @@ private HttpMessageInvoker CreateInternalHttpInvoker(HttpMessageHandler? handler
         {
             handler = HttpHandlerFactory.CreatePrimaryHandler();
         }
+        else
+        {
+            // Validate the user specified handler is compatible with this platform.
+            //
+            // Android's native handler doesn't fully support HTTP/2 and using it could cause hard to understand errors
```
BrennanConroy, line 390: Does this mean the native handler works for some cases? And what if the native handler improves in the future to have better support? Should there be a switch to ignore this check and allow using the native handler?

src/Grpc.Net.Client/GrpcChannel.cs
JamesNK: Opps, yes.

src/Grpc.Net.Client/GrpcChannel.cs
JamesNK: My thought is if that happens, the grpc-dotnet client can be updated to detect the Android OS version that fully supports gRPC, and not throw this error.

BrennanConroy (approved):

Issue #1922, opened 2022-10-20
Handling case where unary call contains a status of OK on success


leonardder:
I'm trying to host several GRPC services under IIS that are load balanced using Citrix ADC. Somehow, Citrix ADC adds a grpc status header with code 0 (OK) to succesful unary calls. According to the following snippet, this is not allowed:
https://github.com/grpc/grpc-dotnet/blob/45dd5d171c2ffe446a7c30d4d7b849db3c297bbe/src/Grpc.Net.Client/Internal/GrpcCall.cs?plain=1#L514-L525
And indeed, I'm getting an exception with a message of ""Failed to deserialize response message" when doing an unary call. The same call doesn't fail whenusing postman.

1. Is this check for status = 0 still appropriate?
2. Shouldn't the status have a more descriptive message of what's actually happening, e.g. "The server returned a status code unexpectedly"?

JamesNK:
Returning `grpc-status` in the header isn't unexpected. It indicates a trailers-only response. See https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses

A response with a `grpc-status` header of 0 (OK) doesn't make sense. Because there is no message sent in a trailers-only response, there is no way it could be 0 (OK) because there is no message to deserialize.

leonardder:
Thanks for pointing this out. Looks like there is definitely somethign wrong in the ADC then.
Regardless, would it be possible to change the status message into something more descriptive? I find "Failed to deserialize response message" still a bit confusing in this case.

JamesNK:
Done: https://github.com/grpc/grpc-dotnet/pull/1931

leonardder:
Thanks a lot!

PR #1931, opened 2022-11-01
Improve logging message when gRPC status OK in header

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1922

### DIFF
--- a/src/Grpc.Net.Client/Internal/GrpcCall.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCall.cs
@@ -514,8 +514,8 @@ internal sealed partial class GrpcCall<TRequest, TResponse> : GrpcCall, IGrpcCal
                         else
                         {
                             // The server should never return StatusCode.OK in the header for a unary call.
-                            // If it does then throw an error that no message was returned from the server.
-                            GrpcCallLog.MessageNotReturned(Logger);
+                            // If it does then throw an error that explains why the server response is invalid.
+                            GrpcCallLog.InvalidGrpcStatusInHeader(Logger);
 
                             // Change the status code to a more accurate status.
                             // This is consistent with Grpc.Core client behavior.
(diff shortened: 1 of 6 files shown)


src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -514,8 +514,8 @@ private async Task RunCall(HttpRequestMessage request, TimeSpan? timeout)
                         else
                         {
                             // The server should never return StatusCode.OK in the header for a unary call.
-                            // If it does then throw an error that no message was returned from the server.
-                            GrpcCallLog.MessageNotReturned(Logger);
+                            // If it does then throw an error that explains why the server response is invalid.
+                            GrpcCallLog.InvalidGrpcStatusInHeader(Logger);
```
BrennanConroy, line 518: Should the message on line 522 also be updated?

test/FunctionalTests/Client/UnaryTests.cs
```diff
@@ -36,7 +36,8 @@ public async Task ThrowErrorWithOK_ClientThrowsFailedToDeserializeError()
     {
         SetExpectedErrorsFilter(writeContext =>
         {
-            if (writeContext.State.ToString() == "Message not returned from unary or client streaming call.")
+            if (writeContext.State.ToString() == "Header contains an OK gRPC status. This is invalid for unary or client streaming calls because a status in the header indicates there is no response body. " +
```
captainsafia, line 39: Worthwhile to store this message in a constant for use in the tests and implementation?

JamesNK: This is the only place it is used in tests. We're not doing it for anything else so I'd like to leave it as is.

BrennanConroy (approved):

PR #1943, opened 2022-11-10
Update to GA .NET 7 SDK

JamesNK:


build/get-dotnet.sh
```diff
@@ -18,7 +18,7 @@ ensure_dir() {
 # main
 
 # resolve SDK version
-sdk_version="7.0.100-rc.2.22477.23"
+sdk_version=$(jq -r .sdk.version $global_json_path)
```
brunolins16, line 22: Is `jq` pre-installed in all distro?

brunolins16 (approved):

build/get-dotnet.sh
JamesNK: I'm not sure. However, this change reverts the line to the previous code. No one has brought up jq being missing before.

Issue #1969, opened 2022-12-06
Issues with Trimming - Missing annotation in GrpcClientServiceExtensions


Blackclaws:
### What version of gRPC and what language are you using?
2.50

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
7.0.100

### What did you do?
Added a GrpcClient via the service collection extensions factory methods.

### What did you expect to see?
After trimming the client should work as expected

### What did you see instead?
```
/_/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs(216,9): Trim analysis warning IL2091: Microsoft.Extensions.DependencyInjection.GrpcClientServiceExtensions.AddGrpcClient<TClient>(IServiceCollection, String, Action<GrpcClientFactoryOptions>): 'TClient' generic argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors' in 'Microsoft.Extensions.DependencyInjection.GrpcClientServiceExtensions.AddGrpcClientCore<TClient>(IServiceCollection, String)'. The generic parameter 'TClient' of 'Microsoft.Extensions.DependencyInjection.GrpcClientServiceExtensions.AddGrpcClient<TClient>(IServiceCollection, String, Action<GrpcClientFactoryOptions>)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/home/felix/spielwiese/GrpcTest/grpc-trim-issue/GrpcTest.csproj]
```

Unhandled exception. System.InvalidOperationException: A suitable constructor for type 'Test+TestClient' could not be located. Ensure the type is concrete and all parameters of a public constructor are either registered as services or passed as arguments. Also ensure no extraneous arguments are provided.

Because well.. the constructor was trimmed.

According to the docs:
https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/fixing-warnings

The dynamically accessed members annotation needs to be escalated all the way to the top of the generic chain. It unfortunately isn't enough to just add it at the bottom.

PR #1970, opened 2022-12-06
fix: Fixes missing constructor in trimmed builds

Blackclaws:
When trimming, the whole call chain for generics needs to be annotated with the same or broader scope of dynamically accessed members. See also: https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/fixing-warnings#dynamicallyaccessedmembers

This adds the required annotations up the call chain for the service extensions

Closes:  #1969

### DIFF
--- a/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
+++ b/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
@@ -53,7 +53,11 @@ public static class GrpcClientServiceExtensions
     /// <typeparamref name="TClient"/> as the service type.
     /// </para>
     /// </remarks>
-    public static IHttpClientBuilder AddGrpcClient<TClient>(this IServiceCollection services)
+    public static IHttpClientBuilder AddGrpcClient<
+#if NET5_0_OR_GREATER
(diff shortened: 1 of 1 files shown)


JamesNK (approved):

JamesNK:
Thanks. I wonder why trimming analysis didn't catch this. TClient on the core method is annotated. It should complain that the calling methods aren't annotated.

Blackclaws:
> Thanks. I wonder why trimming analysis didn't catch this?
> 
> TClient on the core method is annotated. Analysis should complain that the TClient passed from the calling methods to the core method isn't annotated with PublicConstructors.

I does catch it. The problem is you need to actually publish a self-contained trimmed app for the trim analyzer to run correctly. See also: https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming#show-all-warnings-with-sample-application

So basically you have to add a sample application that references the library as a root assembly even if it does _nothing_ and publish that to see the actual warnings.

JamesNK:
Got it. There is a sample app in the tests for this purpose, but it didn't use the client factory. I'll update it.

PR #1972, opened 2022-12-07
Add client factory to client linker test app

JamesNK:
Verify https://github.com/grpc/grpc-dotnet/pull/1970 in tests.

cc @Blackclaws


Blackclaws:
You should specify the Grpc.Net.ClientFactory assembly as a trimmer root so that it will actually traverse all methods even those not called by any test code. That's also what the official docs recommend you should do to test a library for trim problems.

BrennanConroy (approved):

PR #1985, opened 2022-12-11
Support zero-byte read in gRPC client

JamesNK:
Avoid renting and using a buffer until a response is available using a zero-byte read of the response.

### DIFF
--- a/src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
+++ b/src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
@@ -31,8 +31,6 @@ namespace Grpc.Net.Client.Web.Internal;
 /// </summary>
 internal class GrpcWebResponseStream : Stream
 {
-    // This uses C# compiler's ability to refer to static data directly. For more information see https://vcsjones.dev/2019/02/01/csharp-readonly-span-bytes-static
-    private static ReadOnlySpan<byte> BytesNewLine => new byte[] { (byte)'\r', (byte)'\n' };
     private const int HeaderLength = 5;
 
     private readonly Stream _inner;
@@ -60,26 +58,24 @@ internal class GrpcWebResponseStream : Stream
 #endif
         var headerBuffer = Memory<byte>.Empty;
 
+        if (data.Length == 0)
+        {
+            // Handle zero byte reads.
+            var read = await StreamHelpers.ReadAsync(_inner, data, cancellationToken).ConfigureAwait(false);
+            Debug.Assert(read == 0);
+            return 0;
+        }
+
         switch (_state)
         {
             case ResponseState.Ready:
                 {
-                    if (data.Length == 0)
-                    {
-                        // Handle zero byte reads.
-                        var read = await StreamHelpers.ReadAsync(_inner, data, cancellationToken).ConfigureAwait(false);
-                        Debug.Assert(read == 0);
-                        return 0;
-                    }
-                    else
-                    {
-                        // Read the header first
-                        // - 1 byte flag for compression
-                        // - 4 bytes for the content length
-                        _contentRemaining = HeaderLength;
-                        _state = ResponseState.Header;
-                        goto case ResponseState.Header;
-                    }
+                    // Read the header first
+                    // - 1 byte flag for compression
+                    // - 4 bytes for the content length
+                    _contentRemaining = HeaderLength;
+                    _state = ResponseState.Header;
+                    goto case ResponseState.Header;
                 }
             case ResponseState.Header:
                 {
@@ -151,7 +147,7 @@ internal class GrpcWebResponseStream : Stream
 
                     // The trailer needs to be completely read before returning 0 to the caller.
                     // Ensure buffer is large enough to contain the trailer header.
-                    if (headerBuffer.Length < 5)
+                    if (headerBuffer.Length < HeaderLength)
                     {
                         var newBuffer = new byte[HeaderLength];
                         headerBuffer.CopyTo(newBuffer);
@@ -225,7 +221,7 @@ internal class GrpcWebResponseStream : Stream
         {
             ReadOnlySpan<byte> line;
 
-            var lineEndIndex = remainingContent.IndexOf(BytesNewLine);
+            var lineEndIndex = remainingContent.IndexOf("\r\n"u8);
             if (lineEndIndex == -1)
             {
                 line = remainingContent;
--- a/src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
+++ b/src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
@@ -56,6 +56,17 @@ internal class GrpcWebResponseStream : Stream
         switch (_state)
         {
             case ResponseState.Ready:
+                if (data.Length == 0)
+                {
+                    // Handle zero byte reads.
+                    return 0;
+                }
+                else
+                {
+                    _state = ResponseState.Header;
+                    goto case ResponseState.Header;
+                }
+            case ResponseState.Header:
                 // Read the header first
                 // - 1 byte flag for compression
                 // - 4 bytes for the content length
@@ -88,7 +99,7 @@ internal class GrpcWebResponseStream : Stream
                 }
 
                 _contentRemaining = length;
-                // If there is no content then state is still ready
+                // If there is no content then state is reset to ready.
                 _state = _contentRemaining > 0 ? ResponseState.Content : ResponseState.Ready;
                 return 5;
             case ResponseState.Content:
@@ -260,6 +271,7 @@ internal class GrpcWebResponseStream : Stream
     private enum ResponseState
     {
         Ready,
+        Header,
         Content,
         Complete
     }
--- a/src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
+++ b/src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
@@ -83,9 +83,11 @@ internal class GrpcWebResponseStream : Stream
                 }
             case ResponseState.Header:
                 {
+                    Debug.Assert(_contentRemaining > 0);
+
                     headerBuffer = data.Length >= _contentRemaining ? data.Slice(0, _contentRemaining) : data;
                     var success = await TryReadDataAsync(_inner, headerBuffer, cancellationToken).ConfigureAwait(false);
-                    if (!success || headerBuffer.Length == 0)
+                    if (!success)
                     {
                         return 0;
                     }
--- a/src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
+++ b/src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
@@ -17,6 +17,8 @@
 #endregion
 
 using System.Buffers.Binary;
+using System.Data;
+using System.Diagnostics;
 using System.Net.Http.Headers;
 using System.Text;
 
@@ -31,11 +33,15 @@ internal class GrpcWebResponseStream : Stream
 {
     // This uses C# compiler's ability to refer to static data directly. For more information see https://vcsjones.dev/2019/02/01/csharp-readonly-span-bytes-static
     private static ReadOnlySpan<byte> BytesNewLine => new byte[] { (byte)'\r', (byte)'\n' };
+    private const int HeaderLength = 5;
 
     private readonly Stream _inner;
     private readonly HttpHeaders _responseTrailers;
-    private int _contentRemaining;
-    private ResponseState _state;
+    private byte[]? _headerBuffer;
+
+    // Internal for testing
+    internal ResponseState _state;
+    internal int _contentRemaining;
 
     public GrpcWebResponseStream(Stream inner, HttpHeaders responseTrailers)
     {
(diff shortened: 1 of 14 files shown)


src/Grpc.Net.Client/Internal/StreamExtensions.cs
```diff
@@ -62,6 +62,13 @@ private static Status CreateUnknownMessageEncodingMessageStatus(string unsupport
             GrpcCallLog.ReadingMessage(call.Logger);
             cancellationToken.ThrowIfCancellationRequested();
 
+#if NET7_0_OR_GREATER
+            // Start with zero-byte read.
+            // A zero-byte read avoids renting buffer until the response is ready. Especially useful for long running streaming calls.
+            var readCount = await responseStream.ReadAsync(Memory<byte>.Empty, cancellationToken).ConfigureAwait(false);
```
MihaZupan, line 68: Is the common case with gRPC that the whole message will be received quickly once you start receiving data?
That is, the server won't send you 10 bytes, wait for a second, then send you the rest?

JamesNK: There can be delays between messages when streaming, but a message is flushed by the server to the client all at once.

src/Grpc.Net.Client/Internal/StreamExtensions.cs
```diff
@@ -62,6 +62,13 @@ private static Status CreateUnknownMessageEncodingMessageStatus(string unsupport
             GrpcCallLog.ReadingMessage(call.Logger);
             cancellationToken.ThrowIfCancellationRequested();
 
+#if NET7_0_OR_GREATER
```
davidfowl, line 65: Why the ifdef? Something to do with the underlying stream impl?

JamesNK: Yes. Some HttpHandler stream implementations in old frameworks don't like zero byte buffers, e.g. `Microsoft.AspNetCore.TestHost.ResponseBodyReaderStream`

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
```diff
@@ -56,6 +56,17 @@ public override async ValueTask<int> ReadAsync(Memory<byte> data, CancellationTo
         switch (_state)
         {
             case ResponseState.Ready:
+                if (data.Length == 0)
+                {
+                    // Handle zero byte reads.
+                    return 0;
```
halter73, line 62: Shouldn't we try a zero-length read on `_inner` before immediately returning 0? It looks like this logic has the risk stalling in `TryReadDataAsync` during the next call to `ReadAsync` which will result in a 4 KB buffer or larger being rented even though it will be sliced to only 5 bytes for the header.

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
```diff
@@ -88,7 +99,7 @@ public override async ValueTask<int> ReadAsync(Memory<byte> data, CancellationTo
                 }
 
                 _contentRemaining = length;
-                // If there is no content then state is still ready
+                // If there is no content then state is reset to ready.
                 _state = _contentRemaining > 0 ? ResponseState.Content : ResponseState.Ready;
                 return 5;
```
halter73, line 104: I know this isn't new, but this seems really weird to me. Especially given the above `// Should never get here. Client always passes 5 to read the header` comment. If this is intended to be consumed in a special way and not as an abstract Stream where I can pass any buffer length, `GrpcWebResponseStream` should not be implementing `Stream`.

Instead, I think it would be better to implement `ReadMessageAsync` directly here in a non-Stream type rather than as an extension method on `StreamExtensions`. That way you don't end up having to do things like parse the header twice. The extension method does seem resilient enough to handle the inner stream returning a partial header just like `TryReadDataAsync` in this `GrpcWebResponseStream` does, so there seems to be a lot of redundant logic.

JamesNK: The stream is combined with a custom `HttpContent`, which intercepts the incoming HTTP response body and transforms it. Realistically it will only ever be used by Grpc.Net.Client, which always passes a buffer larger than 5 bytes (unless a zero byte read).

But you're right that it doesn't completely conform to the `Stream` spec. I've updated it to accept any size buffer.

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
```diff
@@ -52,59 +58,113 @@ public override async ValueTask<int> ReadAsync(Memory<byte> data, CancellationTo
...
-                else
+            case ResponseState.Header:
                 {
-                    // Should never get here. Client always passes 5 to read the header.
-                    throw new InvalidOperationException("Buffer is not large enough for header");
+                    headerBuffer = data.Length >= _contentRemaining ? data.Slice(0, _contentRemaining) : data;
+                    var success = await TryReadDataAsync(_inner, headerBuffer, cancellationToken).ConfigureAwait(false);
+                    if (!success || headerBuffer.Length == 0)
+                    {
+                        return 0;
```
halter73, line 90: Can `_contentRemaining` ever be zero here? If so, this seems like it could incorrectly report the end of stream. It looks like this might be impossible because you should transition to the `Ready` state first, so maybe just add an assert?

```suggestion
                        Debug.Assert(_contentRemaining > 0);
                        return 0;
```

halter73 (approved):

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
halter73, line 90: Also, doesn't `TryReadDataAsync` always return false given a zero-length header buffer?

```suggestion
                    Debug.Assert(_contentRemaining > 0);
                    if (!success)
                    {
                        return 0;
```

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
```diff
@@ -52,59 +58,115 @@ public override async ValueTask<int> ReadAsync(Memory<byte> data, CancellationTo
 #if NETSTANDARD2_0
         var data = buffer.AsMemory(offset, count);
 #endif
+        Memory<byte> headerBuffer = Memory<byte>.Empty;
```
JamesNK, line 61:
```suggestion
        var headerBuffer = Memory<byte>.Empty;
```

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
```diff
@@ -31,11 +33,15 @@ internal class GrpcWebResponseStream : Stream
 {
     // This uses C# compiler's ability to refer to static data directly. For more information see https://vcsjones.dev/2019/02/01/csharp-readonly-span-bytes-static
     private static ReadOnlySpan<byte> BytesNewLine => new byte[] { (byte)'\r', (byte)'\n' };
```
mgravell, line 35: drive-by observation: `"\r\n"u8;` (if you're happy re lang-ver)

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
```diff
@@ -52,59 +58,115 @@ public override async ValueTask<int> ReadAsync(Memory<byte> data, CancellationTo
...
+                        }
+                    }
+
+                    var read = headerBuffer.Length;
+
+                    // The buffer was less than header length either because this is the first read and the passed in buffer is small,
+                    // or it is an additonal read to finish getting header data.
+                    if (headerBuffer.Length < HeaderLength)
+                    {
+                        _headerBuffer ??= new byte[HeaderLength];
```
mgravell, line 113: this `byte[]` troubles me; it is always `5`, right? well, we can get that far more cheaply by abusing a `long` field (which is the same size as the `byte[]` *reference*, but without needing the `byte[]`) - perhaps thunking as a `Span<byte>`:

``` c#
    private long _header;
    private Span<byte> GetHeader() => MemoryMarshal.Cast<long, byte>(new Span<long>(ref _header)).Slice(0, 5);
```

similar can be done on down-level using `unsafe`, or you can just use shifts

Is it worth the evil, though? perhaps:

``` c#
#if NET7_0_OR_GREATER
    private long _header;
    private Span<byte> GetHeader() => MemoryMarshal.Cast<long, byte>(new Span<long>(ref _header)).Slice(0, 5);
#else
    private byte[]? _header;
    private Span<byte> GetHeader() => _header ??= new byte[5];
#endif
```

MihaZupan:
```c#
MemoryMarshal.CreateSpan(ref Unsafe.As<long, byte>(ref _headerBuffer), 5);
```
will likely get better codegen

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
```diff
@@ -52,59 +58,115 @@ public override async ValueTask<int> ReadAsync(Memory<byte> data, CancellationTo
...
-
-                var isTrailer = IsBitSet(compressed, pos: 7);
-                if (isTrailer)
+            case ResponseState.Trailer:
                 {
+                    Debug.Assert(headerBuffer.Length > 0);
+
+                    // The trailer needs to be completely read before returning 0 to the caller.
+                    // Ensure buffer is large enough to contain the trailer header.
+                    if (headerBuffer.Length < 5)
```
mgravell, line 154: /s/5/HeaderLength ?

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
```diff
@@ -52,59 +58,115 @@ public override async ValueTask<int> ReadAsync(Memory<byte> data, CancellationTo
...
+                    {
+                        var newBuffer = new byte[HeaderLength];
```
mgravell, line 156: as prev

mgravell (commented):
(minor comments added)

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
mgravell: yeah, I thought about that alternative - it does lose the bounds check, but on the other hand, it will never be true that `sizeof(long) < 5`, so that bounds check is redundant... but: what *really* makes that tempting is that `MemoryMarshal.CreateSpan` approach allows us to go all the way back over all netcoreapp2+ and netstandard2.1, so: much better TFM coverage 👍

JamesNK: Header buffer is passed to async methods and needs to be `Memory<byte>`.

This allocation only happens if someone uses a small buffer (less than 5 bytes) with the stream. The primary user of the stream is `Grpc.Net.Client`, which uses a much larger buffer.

Realistically, I doubt people will hit this allocation in the real world.

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
```diff
@@ -52,59 +56,113 @@ public override async ValueTask<int> ReadAsync(Memory<byte> data, CancellationTo
...
+                        if (!success)
+                        {
+                            return 0;
+                        }
+                        headerBuffer = newBuffer;
+                    }
+                    var length = (int)BinaryPrimitives.ReadUInt32BigEndian(headerBuffer.Span.Slice(1));
+
                     await ReadTrailersAsync(length, data, cancellationToken).ConfigureAwait(false);
                     return 0;
```
amcasey, line 164: What's a trailer and why do we always return "0 bytes read" for them?

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
amcasey: Could we just throw `NotSupportedException` or something if someone tries to read fewer than `HeaderCount` bytes?  This seems like a lot of complexity for code we don't expect or want people to use.

test/Grpc.Net.Client.Web.Tests/GrpcWebResponseStreamTests.cs
```diff
@@ -59,6 +59,110 @@ public async Task ReadAsync_EmptyMessage_ParseMessageAndTrailers()
         Assert.AreEqual("0", trailingHeaders.GetValues("grpc-status").Single());
     }
 
+    [Test]
+    public async Task ReadAsync_HasMessage_OneByteBuffer_ParseMessageAndTrailers()
```
amcasey, line 63: Other than the internal state checks, these tests would pass without the product change, correct?

amcasey (approved):
Seems sensible.  To confirm my understanding of the motivation, this won't _reduce_ renting of buffers but it will _defer_ renting until they're actually needed, reducing contention?

src/Grpc.Net.Client.Web/Internal/GrpcWebResponseStream.cs
mgravell: My read is that they're not pard of the content payload, and thus we're not reporting them as part of the Stream content; likewise, once we've seen a trailer we know we're at the EOF, hence: return 0 to signal that. But I'm very interested in whether my interpretation was right!

halter73: You're correct. Typically, gRPC uses HTTP/2's built-in trailing header functionality for the "grpc-status" and "grpc-message" trailers, but [gRPC Web encodes this in the response body](https://fuchsia.googlesource.com/third_party/grpc/+/refs/tags/v1.6.5/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2). Waiting for the response body to complete before sending the status allows services to report errors that occur after they start streaming a response.

[HTTP/1.1 also technically supports trailing headers for chunked responses](https://httpwg.org/specs/rfc9112.html#chunked.trailer.section), but support is spotty. I assume that's why gRPC Web doesn't rely on it.

JamesNK: > My read is that they're not pard of the content payload, and thus we're not reporting them as part of the Stream content; likewise, once we've seen a trailer we know we're at the EOF, hence: return 0 to signal that. But I'm very interested in whether my interpretation was right!

Correct.

test/Shared/SyncPointMemoryStream.cs
```diff
@@ -37,6 +39,17 @@ public SyncPointMemoryStream(bool runContinuationsAsynchronously = true)
         _awaiter = SyncPoint.Create(out _syncPoint, _runContinuationsAsynchronously);
     }
 
+    /// <summary>
+    /// End stream and wait for at least one more read that returns zero bytes.
+    /// Note that because of zero-byte reads, the stream maybe read multiple times.
```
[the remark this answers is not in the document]
JamesNK:
```suggestion
    /// Note that because of zero-byte reads, the stream may be read multiple times.
```

PR #1988, opened 2022-12-13
Added EnabledExceptionPropagation functionality

cfox-dev:
The _gRpc pipeline_ has been extended with the functionality to propagate a custom exception from a gRpc server up to the client. 

At the client side, the functionality can be enabled with the extension method `EnableExceptionPropagation` like showed by the code snippet below.
```
services.AddGrpcClient<Greeter.GreeterClient>(o =>
{
	o.Address = new Uri("http://localhost:5210");
})
.EnableExceptionPropagation();
```

At the server side, the gRpc can be configured with the `ExceptionHandlingInterceptor` interceptor.
```
builder.Services.AddGrpc(o =>
{
	o.Interceptors.Add<ExceptionHandlingInterceptor>();
});
```
The goal of this functionality is to let the client capture the exception like a normal control-flow. From its point of view, it will be totally independent of the communication used (it can catch its custom exception instead of catching generic RpcException).
Ex. If a method defined in the gRpc service has been designed to throw a custom **MyException**, the client can catch it by simply doing:
```
try
{
	var whoIs = _myServiceProxy.SayHello(name:"fox");
}
catch (MyException e)
{
	Console.WriteLine(e.Message);
}
```

### DIFF
--- a/src/Grpc.Core.Api/StatusCode.cs
+++ b/src/Grpc.Core.Api/StatusCode.cs
@@ -120,5 +120,10 @@ public enum StatusCode
     Unavailable = 14,
 
     /// <summary>Unrecoverable data loss or corruption.</summary>
-    DataLoss = 15
+    DataLoss = 15,
+
+    /// <summary>
+    /// The operation throws an application exception.
+    /// </summary>
+    Application = 16
 }
--- a/src/Grpc.AspNetCore.Server.ClientFactory/ExceptionPropagationInterceptor.cs
+++ b/src/Grpc.AspNetCore.Server.ClientFactory/ExceptionPropagationInterceptor.cs
@@ -1,0 +1,111 @@
+using System;
+using Grpc.Core;
+using Newtonsoft.Json;
+using Grpc.AspNetCore.Server;
+using Grpc.Core.Interceptors;
+using System.Threading.Tasks;
+
+namespace Grpc.AspNetCore.ClientFactory;
+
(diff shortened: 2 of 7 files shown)


cfox-dev:
Hi @JamesNK . Please could you have a look at these changes? I mean, if there is room for improvements I'm open to any feedback and I can apply the changes right now for sure.
Otherwise, if that is totally conceptually wrong, I'll close the PR. Just want to contribute on this =)
Thanks

src/Grpc.Core.Api/StatusCode.cs
```diff
@@ -120,5 +120,10 @@ public enum StatusCode
     Unavailable = 14,
 
     /// <summary>Unrecoverable data loss or corruption.</summary>
-    DataLoss = 15
+    DataLoss = 15,
+
+    /// <summary>
+    /// The operation throws an application exception.
+    /// </summary>
+    Application = 16
```
JamesNK, line 128: These values come from the grpc specification. The .NET implementation needs to be compatible with other versions

JamesNK (commented):
I think these changes in this PR are more appropriate to a third-party NuGet package, and not built into grpc-dotnet. We need to stay compatible with other implementations.

src/Grpc.Core.Api/StatusCode.cs
cfox-dev: Ah ok, good to know.
Thanks so much for your time! Closing this PR then...

PR #1992, opened 2022-12-19
RpcException API improvements

jtattermusch:
Addresses https://github.com/grpc/grpc/issues/27066 and https://github.com/grpc/grpc/issues/25740 (in a backward compatible way).

Two options for addressing the problem in Grpc.Core as well:
- if users pull the latest versions of Grpc.Core.Api (which now lives in the grpc-dotnet repo), they'll get the fix as well.  It is fine to use Grpc.Core at e.g. 2.46.5 (currently the latest patch) and Grpc.Core.Api  2.47.x, 2.48.x, ... together.
- we could potentially backport into grpc/grpc's 1.46.x branch (which seems unnecessary due to option 1 though).

### DIFF
--- a/src/Grpc.Core.Api/Status.cs
+++ b/src/Grpc.Core.Api/Status.cs
@@ -88,7 +88,8 @@ public struct Status
     {
         if (DebugException != null)
         {
-            return $"Status(StatusCode=\"{StatusCode}\", Detail=\"{Detail}\", DebugException=\"{DebugException}\")";
+            return $"Status(StatusCode=\"{StatusCode}\", Detail=\"{Detail}\"," +
+                $" DebugException=\"{DebugException.GetType().Name}: {DebugException.Message}\")";
         }
         return $"Status(StatusCode=\"{StatusCode}\", Detail=\"{Detail}\")";
     }
--- a/src/Grpc.Core.Api/Status.cs
+++ b/src/Grpc.Core.Api/Status.cs
@@ -89,7 +89,7 @@ public struct Status
         if (DebugException != null)
         {
             return $"Status(StatusCode=\"{StatusCode}\", Detail=\"{Detail}\"," +
-                $" DebugException=\"{DebugException.GetType().Name}: {DebugException.Message}\")";
+                $" DebugException=\"{DebugException.GetType()}: {DebugException.Message}\")";
         }
         return $"Status(StatusCode=\"{StatusCode}\", Detail=\"{Detail}\")";
     }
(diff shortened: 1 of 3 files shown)


jtattermusch:
CC @mgravell @tonydnewell

src/Grpc.Core.Api/Status.cs
```diff
@@ -88,7 +88,8 @@ public override string ToString()
     {
         if (DebugException != null)
         {
-            return $"Status(StatusCode=\"{StatusCode}\", Detail=\"{Detail}\", DebugException=\"{DebugException}\")";
+            return $"Status(StatusCode=\"{StatusCode}\", Detail=\"{Detail}\"," +
+                $" DebugException=\"{DebugException.GetType().Name}: {DebugException.Message}\")";
```
mgravell, line 92: a slightly odd choice to show the type name - especially as concrete exception types are often internal, but: not *invalid* - just me adding a "are you 100% sure you want this?" check

mgravell (approved):
minor thought added re text; but: I wholeheartedly support the change (and populating the inner exception), 👍

src/Grpc.Core.Api/Status.cs
JamesNK: Showing the type name is normal in `Exception.ToString()`

JamesNK, line 92: `Exception.ToString()` includes the namespace of the exception (`InvalidOperationException` vs `System.InvalidOperationException`).

The goal is to replicate ToString, but without the long stack trace. This is slightly more correct:
```suggestion
                $" DebugException=\"{DebugException.GetType()}: {DebugException.Message}\")";
```

test/Grpc.Core.Api.Tests/RpcExceptionTest.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+        catch (Exception caughtEx)
+        {
+            someExceptionWithStacktrace = caughtEx;
+        }
+        var ex = new RpcException(new Status(StatusCode.Internal, "abc", someExceptionWithStacktrace));
+        Console.WriteLine(ex.Message);
+        // Check debug exceptions's message is contained.
+        StringAssert.Contains(someExceptionWithStacktrace.Message, ex.Message);
+        StringAssert.Contains(someExceptionWithStacktrace.GetType().Name, ex.Message);
```
JamesNK, line 57:
```suggestion
        StringAssert.Contains(someExceptionWithStacktrace.GetType().FullName, ex.Message);
```

test/Grpc.Core.Api.Tests/RpcExceptionTest.cs
```diff
@@ -0,0 +1,61 @@
...
+        try
+        {
+            throw new ArgumentException("test exception");
+        }
+        catch (Exception caughtEx)
+        {
+            someExceptionWithStacktrace = caughtEx;
+        }
+        var ex = new RpcException(new Status(StatusCode.Internal, "abc", someExceptionWithStacktrace));
+        Console.WriteLine(ex.Message);
```
JamesNK, line 54:
```suggestion

```

JamesNK (approved):

PR #1993, opened 2022-12-20
Support writing to response stream with gRPC-Web middleware

JamesNK:
Addresses https://github.com/microsoft/reverse-proxy/issues/1968

tldr; gRPC-Web server middleware can run independently of Grpc.AspNetCore. However, it only supports writing to `Response.BodyWriter` (that's what Grpc.AspNetCore uses). Update to support `Response.Body`


JamesNK:
cc @tratcher

Tratcher (approved):

davidfowl (approved):

Issue #1989, opened 2022-12-13
Retrying immediately triggers a timeout


lobster2012-user:
### What version of gRPC and what language are you using?
net6.0
grpc 2.50

### What operating system (Linux, Windows,...) and version?
windows 10

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
7.0.100

### What did you do?
I am in a loop trying to send a request using the same channel. The connection timeout is set to 5 seconds in the channel settings.

### What did you expect to see?
For a non-working address, a connection error will be triggered after the timeout set in the channel settings

### What did you see instead?
Starting from the second attempt, the connection error is thrown out IMMEDIATELY.

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

13.12.2022 21:15:31
Grpc.Core.RpcException: Status(StatusCode="Unavailable", Detail="Error connecting to subchannel.", DebugException="System.TimeoutException: A connection could not be established within the configured ConnectTimeout.
 ---> System.OperationCanceledException: The operation was canceled.
   at System.Threading.CancellationToken.ThrowOperationCanceledException()
   at System.Threading.CancellationToken.ThrowIfCancellationRequested()
   at System.Net.Sockets.Socket.<ConnectAsync>g__WaitForConnectWithCancellation|277_0(AwaitableSocketAsyncEventArgs saea, ValueTask connectTask, CancellationToken cancellationToken)
   at Grpc.Net.Client.Balancer.Internal.SocketConnectivitySubchannelTransport.TryConnectAsync(ConnectContext context)
   --- End of inner exception stack trace ---")
   at Grpc.Net.Client.Balancer.Internal.ConnectionManager.PickAsync(PickContext context, Boolean waitForReady, CancellationToken cancellationToken)
   at Grpc.Net.Client.Balancer.Internal.BalancerHttpHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at Grpc.Net.Client.Internal.GrpcCall`2.RunCall(HttpRequestMessage request, Nullable`1 timeout)
   at Grpc.Net.Client.Internal.HttpClientCallInvoker.BlockingUnaryCall[TRequest,TResponse](Method`2 method, String host, CallOptions options, TRequest request)
   at Grpc.Core.Interceptors.InterceptingCallInvoker.<BlockingUnaryCall>b__3_0[TRequest,TResponse](TRequest req, ClientInterceptorContext`2 ctx)
[...]

PR #1998, opened 2022-12-27
Fix reconnecting after connection timeout

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1989


src/Grpc.Net.Client/Balancer/Subchannel.cs
```diff
@@ -273,9 +273,17 @@ private async Task ConnectTransportAsync()
                     }
                 }
 
-                if (await _transport.TryConnectAsync(connectContext).ConfigureAwait(false))
+                switch (await _transport.TryConnectAsync(connectContext).ConfigureAwait(false))
                 {
-                    return;
+                    case ConnectResult.Success:
+                        return;
+                    case ConnectResult.Timeout:
```
ttohin, line 280: I had exactly the same issue but with SocketException (for example in case of DNS resolution failure). After the SocketException the channel became unusable because every grpc call fails with the exception.

I would be good to cover this situation as well and in generic way.

JamesNK: When that error happens, the connection should continue trying to connect. If your app is configured to write logs at a trace level, you should see `StartingConnectBackoff` logs.

ttohin: Unfortunately, I don't have trace level enabled for gprc in prod. What I see is that my Grpc client became unusable for ~6 hours. Interesting that for the same application instance all other operations (CosmosDb calls, some non-grpc related HTTP calls and so on) were working fine.

From [here](https://learn.microsoft.com/en-us/aspnet/core/grpc/loadbalancing?view=aspnetcore-7.0#configure-custom-resolvers-and-load-balancers):
> By default, a DNS resolver is refreshed if a connection is interrupted. 

May be the DNS refresh wasn't triggered after the first SocketException?

JamesNK: I can't say without seeing logging. There are a lot of asynchronous actions happening inside the client, especially when connections drop and the resolver updates. That is why there is detailed logging of everything that is happening.

Are you using an up-to-date version of Grpc.Net.Client? Try 2.50.0. If the problem persists then I can take a look at the logs and try to determine what is going on.

BrennanConroy (approved):

PR #2012, opened 2023-01-08
Set WaitForDataBeforeAllocatingBuffer to false in benchmark app

JamesNK:
A small throughput boost at the cost of memory usage


mgravell (approved):
(edit: d'oh, ignore me; you already covered this in words)

LGTM

amcasey:
How does this relate to the zero-byte read change, #1985?

JamesNK:
Unrelated. This is on the server and is only for a benchmark app.

Issue #2014, opened 2023-01-11
OperationCanceledException thrown when canceling a CancelationToken does not reference the token canceled


hakenr:
### What version of gRPC and what language are you using?
Grpc.Net.Client v2.51.0, .NET C#, Blazor WebAssembly

### What operating system (Linux, Windows,...) and version?
WebAssembly (browser) - .NET

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.NET SDK 7.0.2

### What did you do?
If you cancel a CancelationToken provided to the gRPC client, the OperationCanceledException thrown does not reference the token.

Context: Many implementations use this pattern to "catch" the cancelation:
```csharp
        catch (Exception e)
        {
            if (e is OperationCanceledException oce && oce.CancellationToken == cancellationToken)
            {
                // No-op; we canceled the operation, so it's fine to suppress this exception.
            }
```
e.g. Blazor's `Virtualize` component: https://github.com/dotnet/aspnetcore/blob/dfab67748084f120472e634a975b3c45150f608b/src/Components/Web/src/Virtualization/Virtualize.cs#L378-L383

### What did you expect to see?
1) The `OperationCanceledException.CancellationToken` (in `RpcException.DebugException`) should be the original `CancellationToken` canceled.
2) In case of cancellation, the `OperationFailedException` should not be wrapped by `RpcException`. It does not make sense as there is literally no use case where you would take advantage of having a `RpcException` instead of the very specific `OperationCanceledException` being caught by the caller expecting cancellation (such as `Virtualize` component in Blazor).

### What did you see instead?
`OperationCanceledException.CancellationToken` is `null`.
`OperationCanceledException` is masked by `RpcException` which implies need for "correcting" interceptor.

### Anything else we should know about your project / environment?
I created a simple repro with Blazor-Virtualize.
https://github.com/hakenr/VirtualizeGrpcRepro

The repro includes an active interceptor which "fixes" the error (for the `AsyncUnaryCall`):
https://github.com/hakenr/VirtualizeGrpcRepro/blob/master/BlazorGrpcWebCodeFirst/Client/CancellationWorkaroundGrpcClientInterceptor.cs

If you remove the registration of `CancellationWorkaroundGrpcClientInterceptor` from `BlazorGrpcWebCodeFirst.Client/Program.cs` (line 18) you immediately see the result - the `OperationFailedException` not being caught by `Virtualize`, which breaks the app.

[...]

PR #2018, opened 2023-01-16
Report user canceled cancellation token

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2014

We want to report the user specified cancellation token via `OperationCanceledException.CancellationToken` if it is the cause of cancellation. Today, an internal cancellation token is typically provided.

### DIFF
--- a/src/Grpc.Net.Client/Internal/GrpcCall.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCall.cs
@@ -734,7 +734,8 @@ internal sealed partial class GrpcCall<TRequest, TResponse> : GrpcCall, IGrpcCal
         return false;
     }
 
-    // Report correct CancellationToken. Preserve stacktrace when possible.
+    // Report correct CancellationToken. This method creates and throws a new OperationCanceledException with a different CancellationToken.
+    // It attempts to preserve the original stack trace using ExceptionDispatchInfo where available.
 #if NET6_0_OR_GREATER
     [StackTraceHidden]
 #endif
--- a/src/Grpc.Net.Client/Internal/GrpcCall.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCall.cs
@@ -23,7 +23,7 @@ using Grpc.Core;
 using Grpc.Net.Client.Internal.Http;
 using Grpc.Shared;
 using Microsoft.Extensions.Logging;
-using System.Threading.Tasks;
+using System.Runtime.ExceptionServices;
 #if SUPPORT_LOAD_BALANCING
 using Grpc.Net.Client.Balancer.Internal;
 #endif
@@ -179,7 +179,7 @@ internal sealed partial class GrpcCall<TRequest, TResponse> : GrpcCall, IGrpcCal
         {
             Disposed = true;
 
-            Cleanup(GrpcProtocolConstants.DisposeCanceledStatus);
+            Cleanup(GrpcProtocolConstants.CreateDisposeCanceledStatus(exception: null));
 
             // If the call was disposed then observe any potential response exception.
             // Observe the task's exception to prevent TaskScheduler.UnobservedTaskException from firing.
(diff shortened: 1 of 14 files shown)


test/Grpc.Net.Client.Tests/CancellationTests.cs
```diff
@@ -90,7 +90,7 @@ public async Task AsyncClientStreamingCall_CancellationDuringSend_ThrowOperation
 
         cts.Cancel();
 
-        var ex = await ExceptionAssert.ThrowsAsync<TaskCanceledException>(() => responseTask).DefaultTimeout();
+        var ex = await ExceptionAssert.ThrowsAsync<OperationCanceledException>(() => responseTask).DefaultTimeout();
         Assert.AreEqual(StatusCode.Cancelled, call.GetStatus().StatusCode);
         Assert.AreEqual("Call canceled by the client.", call.GetStatus().Detail);
```
BrennanConroy, line 95: Assert ex.CancellationToken?

src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
```diff
@@ -69,8 +69,9 @@ internal static class GrpcProtocolConstants
 
     internal static readonly Status DeadlineExceededStatus = new Status(StatusCode.DeadlineExceeded, string.Empty);
     internal static readonly Status ThrottledStatus = new Status(StatusCode.Cancelled, "Retries stopped because retry throttling is active.");
-    internal static readonly Status ClientCanceledStatus = new Status(StatusCode.Cancelled, "Call canceled by the client.");
-    internal static readonly Status DisposeCanceledStatus = new Status(StatusCode.Cancelled, "gRPC call disposed.");
+
+    internal static Status CreateClientCanceledStatus(Exception? exception) => new Status(StatusCode.Cancelled, "Call canceled by the client.", exception);
```
BrennanConroy, line 73: nit: these two calls can use the constants in `GrpcProtocolConstants` if the exception is null.

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -723,6 +734,43 @@ internal bool ResolveException(string summary, Exception ex, [NotNull] out Statu
         return false;
     }
 
+    // Report correct CancellationToken. This method creates and throws a new OperationCanceledException with a different CancellationToken.
```
BrennanConroy, line 737:
```suggestion
    // Report correct CancellationToken. This method creates a new OperationCanceledException with a different CancellationToken.
```

test/FunctionalTests/Client/CancellationTests.cs
```diff
@@ -432,4 +434,321 @@ async Task ServerStreamingCall(DataMessage request, IServerStreamWriter<DataMess
...
+            return true;
+        });
+
+        // Arrange
+        var method = Fixture.DynamicGrpc.AddUnaryMethod<DataMessage, DataMessage>(UnaryMethod);
+        var channel = CreateChannel(throwOperationCanceledOnCancellation: true);
+        var client = TestClientFactory.Create(channel, method);
+
+        // Act
+        var cts = new CancellationTokenSource(TimeSpan.FromSeconds(0.5));
```
BrennanConroy, line 459: I hate this, can you cancel the token inside `UnaryMethod` instead?

test/FunctionalTests/Client/CancellationTests.cs
```diff
@@ -432,4 +434,321 @@ async Task ServerStreamingCall(DataMessage request, IServerStreamWriter<DataMess
...
+            return true;
+        });
+
+        // Arrange
+        var method = Fixture.DynamicGrpc.AddServerStreamingMethod<DataMessage, DataMessage>(ServerStreamingMethod);
+        var channel = CreateChannel(throwOperationCanceledOnCancellation: true);
+        var client = TestClientFactory.Create(channel, method);
+
+        // Act
+        var cts = new CancellationTokenSource(TimeSpan.FromSeconds(0.5));
```
BrennanConroy, line 521: Same

src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
JamesNK: It isn't important as Status is a struct.

test/FunctionalTests/Client/CancellationTests.cs
JamesNK: Much better idea. All the other tests do this.

BrennanConroy (approved):

JamesNK:
I remember why I used a delay. Canceling in the call doesn't work well with streaming calls. I moved the cancellation in that case to serializing the message.

Issue #2013, opened 2023-01-11
Grpc.Net.Client >= 2.50.0 does not work on Android with grpc-web


cupsos:
Related: #1927
Might be related: https://github.com/dotnet/runtime/issues/45741

`Grpc.Net.Client v2.49.0` + `GrpcWebHandler(new HttpClientHandler())` works but `Grpc.Net.Client v2.50.0` doesn't.

I think my android device cannot use `SocketsHttpHandler` because `await new HttpClient(new SocketsHttpHandler()).GetAsync("https://aka.ms")` throws an exception.

What are options for android device that cannot use `SocketsHttpHandler`?

## What I tried

| Grpc.Net.Client | GrpcChannelOptions.HttpHandler             | UseNativeHttpHandler | Result                                            |
| --------------- | ------------------------------------------ | -------------------- | ------------------------------------------------- |
| v2.49.0         | `GrpcWebHandler(new HttpClientHandler())`  | not set              | ✅ work                                            |
| v2.49.0         | `GrpcWebHandler(new HttpClientHandler())`  | `false`              | ❌ `SslException` from gRPC call                   |
| v2.49.0         | `GrpcWebHandler(new SocketsHttpHandler())` | not set              | ❌ `SslException` from gRPC call                   |
| v2.51.0         | `GrpcWebHandler(new HttpClientHandler())`  | not set              | ❌ `InvalidOperationException` from channel create |
| v2.51.0         | `GrpcWebHandler(new HttpClientHandler())`  | `false`              | ❌ `SslException` from gRPC call                   |
| v2.51.0         | `GrpcWebHandler(new SocketsHttpHandler())` | not set              | ❌ `SslException` from gRPC call                   |

## Environment

- SDK: 7.0.102
- TFM: net7.0-android
- UI: MAUI Blazor
- OS: Android 7.1.2

## Exception from `GrpcChannel.ForAddress` on Grpc.Net.Client >= 2.50.0

```log
System.InvalidOperationException: The channel configuration isn't valid on Android devices. The channel is configured to use HttpClientHandler and Android's native HTTP/2 library. gRPC isn't fully supported by Android's native HTTP/2 library and it can cause runtime errors. To fix this problem, either configure the channel to use SocketsHttpHandler, or add <UseNativeHttpHandler>false</UseNativeHttpHandler> to the app's project file. For more information, see https://aka.ms/aspnet/grpc/android.
[...]

PR #2020, opened 2023-01-16
Don't warn about Android native handler with grpc-web

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2013

Android native handler isn't compatible... unless grpc-web is configured. Suppress if the grpc-web handler is in the pipeline.


test/Grpc.Net.Client.Tests/Infrastructure/GrpcWebHandler.cs
```diff
@@ -0,0 +1,26 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+namespace Grpc.Net.Client.Web;
+
+public class GrpcWebHandler : DelegatingHandler
```
BrennanConroy, line 21: This is sketchy, why not use the already existing public GrpcWebHandler?

JamesNK:
```suggestion
// Use a GrpcWebHandler type here with the same name and namespace because Grpc.Net.Client.Tests doesn't reference Grpc.Net.Client.Web.
public class GrpcWebHandler : DelegatingHandler
```

JamesNK: I added a comment explaining the reason.

BrennanConroy: And why not add a reference to Grpc.Net.Client.Web?

JamesNK: I tried that, and it worked 😅 

If you're wondering why I did this, it's because Grpc.Net.Client.Web.Test can't reference Grpc.Net.Client because of type conflicts. I eventually moved where I added this test to the other project, and didn't attempt the opposite. Turns out it is ok.

test/Grpc.Net.Client.Tests/GrpcChannelTests.cs
```diff
@@ -443,6 +443,41 @@ public void HttpHandler_HttpClientHandlerOverNativeOnAndroid_ThrowError(bool use
...
+        {
+            var services = new ServiceCollection();
+            services.AddSingleton<IOperatingSystem>(new TestOperatingSystem { IsAndroid = true });
+
+            HttpMessageHandler handler = new HttpClientHandler();
+            if (useDelegatingHandlers)
+            {
+                handler = new TestDelegatingHandler(handler);
+            }
+            handler = new GrpcWebHandler(handler);
```
BrennanConroy, line 463: Should this be before creating the delegating handler?

BrennanConroy (approved):

PR #2026, opened 2023-01-26
Generate record instead of class

KentGrigo:
The motivation for this PR is that I'd like to make an adapter that's a record instead of a class.
But since gRPC generates a class instead of a record, it forces my hand to use a class as well, since a record can't inherit from a class.

I want to use a record in my adapter because it turns this:
```
public class XyzAdapter : XyzService.XyzServiceBase
{
	private readonly IAbc _abc;
        // x 10

	public XyzAdapter(
		IAbc abc,
                // x 10
	)
	{
		_abc = abc;
                // x 10
	}
```

into this

```
public record XyzAdapter(
	IAbc Abc,
        // x 10
) : XyzService.XyzServiceBase {
```

### DIFF
--- a/src/Grpc.Reflection/ReflectionGrpc.cs
+++ b/src/Grpc.Reflection/ReflectionGrpc.cs
@@ -81,9 +81,9 @@ namespace Grpc.Reflection.V1Alpha {
       get { return global::Grpc.Reflection.V1Alpha.ReflectionReflection.Descriptor.Services[0]; }
     }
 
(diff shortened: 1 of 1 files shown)


JamesNK:
You've changed a source-generated file. It will be automatically overridden the next time the file is regenerated.

KentGrigo:
Ah. Sorry.
I've made another PR that actually changes the generated code in [gRPC](https://github.com/grpc/grpc) now.
Thanks for the heads-up!

Issue #2025, opened 2023-01-26
Slightly more detailed exceptions, maybe?


voroninp:
Instead of:

> Key for binary valued metadata entry needs to have suffix indicating binary value.

it could be

> Key for binary valued metadata entry needs to have suffix indicating binary value. (see Metadata.BinaryHeaderSuffix)

;-)

Tette6969:
@analogrelay 
#1778

PR #2027, opened 2023-01-26
Mention Metadata.BinaryHeaderSuffix in xmldoc comments of Metadata.En…

voroninp:
…try constructor and in message of ArgumentException. (#2025)

### DIFF
--- a/src/Grpc.Core.Api/Metadata.cs
+++ b/src/Grpc.Core.Api/Metadata.cs
@@ -130,8 +130,8 @@ public sealed class Metadata : IList<Metadata.Entry>
     /// <summary>
     /// Adds a new ASCII-valued metadata entry.
     /// </summary>
-    /// <param name="key"><inheritdoc cref="Entry(string, string)" path="/param[@name='key']" /></param>
-    /// <param name="value"><inheritdoc cref="Entry(string, string)" path="/param[@name='value']" /></param>
+    /// <param name="key">Metadata key. Gets converted to lowercase. Must not use <c>-bin</c> suffix indicating a binary-valued metadata entry. Can only contain lowercase alphanumeric characters, underscores, hyphens, and dots.</param>
+    /// <param name="value">Value string. Only ASCII characters are allowed.</param>
     public void Add(string key, string value)
     {
         Add(new Entry(key, value));
@@ -140,8 +140,8 @@ public sealed class Metadata : IList<Metadata.Entry>
     /// <summary>
     /// Adds a new binary-valued metadata entry.
     /// </summary>
-    /// <param name="key"><inheritdoc cref="Entry(string, byte[])" path="/param[@name='key']" /></param>
-    /// <param name="valueBytes"><inheritdoc cref="Entry(string, byte[])" path="/param[@name='valueBytes']" /></param>
+    /// <param name="key">Metadata key. Gets converted to lowercase. Needs to have <c>-bin</c> suffix indicating a binary-valued metadata entry. The binary header suffix can be added to the key with <see cref="Metadata.BinaryHeaderSuffix"/>. Can only contain lowercase alphanumeric characters, underscores, hyphens, and dots.</param>
+    /// <param name="valueBytes">Value bytes.</param>
     public void Add(string key, byte[] valueBytes)
     {
         Add(new Entry(key, valueBytes));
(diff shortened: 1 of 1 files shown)


voroninp:
Hm, locally tests passed, but here they are failing...

src/Grpc.Core.Api/Metadata.cs
```diff
@@ -290,13 +294,13 @@ private Entry(string key, string? value, byte[]? valueBytes)
         /// <summary>
         /// Initializes a new instance of the <see cref="Grpc.Core.Metadata.Entry"/> struct with a binary value.
         /// </summary>
-        /// <param name="key">Metadata key. Gets converted to lowercase. Needs to have suffix indicating a binary valued metadata entry. Can only contain lowercase alphanumeric characters, underscores, hyphens and dots.</param>
+        /// <param name="key">Metadata key. Gets converted to lowercase. Needs to have suffix indicating a binary valued metadata entry (<see cref="Metadata.BinaryHeaderSuffix"/>). Can only contain lowercase alphanumeric characters, underscores, hyphens and dots.</param>
```
JamesNK, line 297:
```suggestion
        /// <param name="key">Metadata key. Gets converted to lowercase. Needs to have <c>-bin</c> suffix indicating a binary-valued metadata entry. Can only contain lowercase alphanumeric characters, underscores, hyphens, and dots.</param>
```

src/Grpc.Core.Api/Metadata.cs
```diff
@@ -128,16 +128,20 @@ public IEnumerable<Entry> GetAll(string key)
...
     {
         Add(new Entry(key, value));
     }
 
     /// <summary>
-    /// Adds a new binary-valued metadata entry. See <c>Metadata.Entry</c> constructor for params.
+    /// Adds a new binary-valued metadata entry.
     /// </summary>
+    /// <param name="key"><inheritdoc cref="Entry(string, byte[])" path="/param[@name='key']" /></param>
+    /// <param name="valueBytes"><inheritdoc cref="Entry(string, byte[])" path="/param[@name='valueBytes']" /></param>
```
JamesNK, line 144:
```suggestion
    /// <param name="key">Metadata key. Gets converted to lowercase. Needs to have <c>-bin</c> suffix indicating a binary-valued metadata entry. Can only contain lowercase alphanumeric characters, underscores, hyphens, and dots.</param>
    /// <param name="valueBytes">Value bytes.</param>
```

src/Grpc.Core.Api/Metadata.cs
```diff
@@ -128,16 +128,20 @@ public IEnumerable<Entry> GetAll(string key)
     }
 
     /// <summary>
-    /// Adds a new ASCII-valued metadata entry. See <c>Metadata.Entry</c> constructor for params.
+    /// Adds a new ASCII-valued metadata entry.
     /// </summary>
+    /// <param name="key"><inheritdoc cref="Entry(string, string)" path="/param[@name='key']" /></param>
+    /// <param name="value"><inheritdoc cref="Entry(string, string)" path="/param[@name='value']" /></param>
```
JamesNK, line 134:
```suggestion
    /// <param name="key">Metadata key. Gets converted to lowercase. Must not use <c>-bin</c> suffix indicating a binary-valued metadata entry. Can only contain lowercase alphanumeric characters, underscores, hyphens, and dots.</param>
    /// <param name="value">Value string. Only ASCII characters are allowed.</param>
```

src/Grpc.Core.Api/Metadata.cs
```diff
@@ -290,13 +294,13 @@ private Entry(string key, string? value, byte[]? valueBytes)
...
         /// </summary>
-        /// <param name="key">Metadata key. Gets converted to lowercase. Needs to have suffix indicating a binary valued metadata entry. Can only contain lowercase alphanumeric characters, underscores, hyphens and dots.</param>
+        /// <param name="key">Metadata key. Gets converted to lowercase. Needs to have suffix indicating a binary valued metadata entry (<see cref="Metadata.BinaryHeaderSuffix"/>). Can only contain lowercase alphanumeric characters, underscores, hyphens and dots.</param>
         /// <param name="valueBytes">Value bytes.</param>
         public Entry(string key, byte[] valueBytes)
         {
             this.key = NormalizeKey(key);
             GrpcPreconditions.CheckArgument(HasBinaryHeaderSuffix(this.key),
-                "Key for binary valued metadata entry needs to have suffix indicating binary value.");
+                $"Key for binary valued metadata entry needs to have suffix indicating binary value. (See {nameof(Metadata)}.{nameof(BinaryHeaderSuffix)})");
```
JamesNK, line 303:
```suggestion
                $"Key for binary valued metadata entry needs to have '{BinaryHeaderSuffix}' suffix indicating binary value.");
```

JamesNK:
Hi, I like the improvement. A couple of changes:

1. Just say what the `-bin` suffix is in comments.
2. Replaced inhertdoc with copy and paste. These don't change much, and it's easier to see exactly what the comment when looking at source.

voroninp:
@JamesNK  But if comments do not mention `BinaryHeaderSuffix`, won't people define the const one more time in their code?

JamesNK:
> @JamesNK But if comments do not mention `BinaryHeaderSuffix`, won't people define the const one more time in their code?

Good point. Updated comments. Thanks!

JamesNK (approved):

JamesNK:
I don't think it's necessary. Also, it's a breaking change for anyone using named parameters.

Issue #2032, opened 2023-01-30
TaskCanceledException instead of expected exception in server streaming call


lklein53:
Hi together,

I think i encountered a bug where an `RpcException` thrown in a grpc server streaming call is not thrown in the client.
Instead a `TaskCanceledException` is thrown. 
This seems to happen if the exception is thrown before any  `WriteAsync` is performed in the server.

### What version of gRPC and what language are you using?
Grpc.Net.Client 2.50.0, C# 11, Net7

### What operating system (Linux, Windows,...) and version?
Linux, Ubuntu jammy

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.NET SDK:
Version:   7.0.100
Commit:    e12b7af219

### What did you do?
Server code
```
public override Task Get(Empty request, IServerStreamWriter<Data> responseStream, ServerCallContext context) {
    throw new RpcException(new Status(StatusCode.NotFound, "NotFound"));
}
```

Client code
```
using var channel = GrpcChannel.ForAddress("http://localhost:5023");
      
var client = new ServerStreaming.ServerStreaming.ServerStreamingClient(channel);
using var streamingCall = client.Get(new Empty());
await foreach (var data in streamingCall.ResponseStream.ReadAllAsync()) {
    Console.WriteLine(data);
}
```
I realized that adding 
`_ = await streamingCall.ResponseHeadersAsync;` before the loop results in the correct exception being thrown.
Is the cient always expected to await the `ResponseHeadersAsync` ? I haven't found anything in the documentation about this. 

### What did you expect to see?
The exception thrown in the server should be thrown when reading the response stream in the client.

### What did you see instead?
A `TaskCanceledException` is thrown when reading the response stream.

Thanks for taking a look and best regards,
Lars

JamesNK:
Hi

A `TaskCanceledException` isn't expected. I tried to recreate this in a unit test and wasn't able to. An `RpcException` is always reported by the client.

https://github.com/grpc/grpc-dotnet/pull/2034

lklein53:
Hi,
thanks for taking a look. It turns out to be related to setting 
`ThrowOperationCanceledOnCancellation = true` in the `GrpcChannelOptions`. Sorry i haven't realized that it is only happening when it is set to true.
I have attached an isolated reproducer.
[reproducer.zip](https://github.com/grpc/grpc-dotnet/files/10557715/reproducer.zip)

lklein53:
Hi,
[...]

PR #2034, opened 2023-02-01
Fix reporting RpcException from server streaming error

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2032

If all of the below are true, fix throwing the status from the server in the client:

* Channel is throwing OperationCanceledException
* Call is made to server streaming method
* Server streaming method returns non-OK status before streaming result
* Client doesn't await server headers


mgravell (approved):
seems fine

Issue #2024, opened 2023-01-23
Globalization/Localization not working 


rizi:
We have implemented a custom implementation of a localization provider like it's described here: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/localization#use-a-custom-provider

Complete guide:
https://learn.microsoft.com/en-us/aspnet/core/fundamentals/localization

The provider is working fine with MVC/WebApi, but with gRPC(AsyncUnaryCall) it's not working correctly. Our custom provider is called and executed correctly, but as soon as the gRPC service(C#) is executed the default culture is used instead of the one returned by our Custom Provider.

The easiest way to check the current culture is to print the result of Thread.CurrentCulture and Thread.CurrentUICulture.

Is localization not supported in gRPC.net or do we have to configure something additional?

I can not find any details about that.

@JamesNK  Any help would be appreciated.

Fyi:
I also implemented a service interceptor that just does nothing but prints the result of Thread.CurrentCulture and Thread.CurrentUICulture and calls the continuation function --> and it seems that the correct culture is used here, but later (when the method of the service is executed) the wrong(default) culture is used, so it seems there must be something wrong in the last part of the gRPC execution chain.

Br

JamesNK:
I added a test for this and found the server method had the correct culture set:

https://github.com/grpc/grpc-dotnet/pull/2039

rizi:
> I added a test for this and found the server method had the correct culture set:
> 
> #2039

What does this mean, should it work out of the box?
Do you have any ideas why it's not working on our side?

B

JamesNK:
I'm not familiar with the feature. I expect it works with gRPC, and it works in the test I wrote.

I don't know why it isn't working for you.

rizi:
> I'm not familiar with the feature. I expect it works with gRPC, and it works in the test I wrote.
> 
> I don't know why it isn't working for you.

@JamesNK it's interesting that in the grcp interceptor the culture is still correct, can it be interceptor related? Do you know someone from the asp.net core team that could provide some information/help?

Br

JamesNK:
I think you need to create a simple reproduction of the problem and put it in a github repo. Myself or someone from the ASP.NET Core team can take a look.

rizi:
> I think you need to create a simple reproduction of the problem and put it in a github repo. Myself or someone from the ASP.NET Core team can take a look.

@JamesNK thx the unit tests you added helped me to figure out the problem.

Therefore I will close this issue/question.

PR #2039, opened 2023-02-03
Add test that uses localization

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/2024

### DIFF
--- a/test/FunctionalTests/Client/UnaryTests.cs
+++ b/test/FunctionalTests/Client/UnaryTests.cs
@@ -91,10 +91,9 @@ public class UnaryTests : FunctionalTestBase
 
         // Act
         var call = client.UnaryCall(new HelloRequest(), new CallOptions(headers: metadata));
-
-        // Assert
         await call.ResponseAsync.DefaultTimeout();
 
+        // Assert
         Assert.AreEqual(expectedServerCulture, serverCulture);
     }
 
--- a/test/FunctionalTests/Client/UnaryTests.cs
+++ b/test/FunctionalTests/Client/UnaryTests.cs
@@ -68,6 +68,36 @@ public class UnaryTests : FunctionalTestBase
         StringAssert.StartsWith("Failed to deserialize response message.", call.GetStatus().Detail);
     }
 
+    [TestCase("fr", "fr")]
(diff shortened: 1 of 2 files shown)


test/FunctionalTests/Client/UnaryTests.cs
```diff
@@ -68,6 +68,35 @@ Task<HelloReply> UnaryThrowError(HelloRequest request, ServerCallContext context
...
+        }
+
+        // Arrange
+        var method = Fixture.DynamicGrpc.AddUnaryMethod<HelloRequest, HelloReply>(UnaryThrowError);
+        var channel = CreateChannel();
+        var client = TestClientFactory.Create(channel, method);
+        var metadata = new Metadata();
+        if (clientAcceptLanguage != null)
+        {
+            metadata.Add("accept-language", clientAcceptLanguage);
```
mgravell, line 89: API thought unrelated to this PR, but more to the feature; should this be / is this already automated at the client? `metadata.AddRequestLocalization()`? an interceptor? (not sure either works perfectly; just asking the question)

test/FunctionalTests/Client/UnaryTests.cs
```diff
@@ -68,6 +68,35 @@ Task<HelloReply> UnaryThrowError(HelloRequest request, ServerCallContext context
         StringAssert.StartsWith("Failed to deserialize response message.", call.GetStatus().Detail);
     }
 
+    [TestCase("fr", "fr")]
```
mgravell, line 71: IMO we should have a test of what happens if the culture requested is not supported by the server

mgravell (approved):

test/FunctionalTests/Client/UnaryTests.cs
JamesNK: That's testing the localization feature. These tests are to double-check that gRPC doesn't lose the culture that localization sets.

test/FunctionalTests/Client/UnaryTests.cs
JamesNK: No other gRPC clients set this automatically.

Issue #1948, opened 2022-11-15
[.NET 7] Client connects first time, but hangs second time (serialization not happening?) 


Kobus-Smit:
### What version of gRPC and what language are you using?

Grpc.AspNetCore.Server 2.49.0
Grpc.Net.Client 2.49.0

I've also tried
Grpc.AspNetCore.Server 2.50.0-dev202211150802
Grpc.Net.Client 2.50.0-dev202211150802

### What operating system (Linux, Windows,...) and version?
Microsoft Windows NT 10.0.19044.0

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
VS 2022 17.4.0
.NET SDK  7.0.100
Runtime Microsoft.WindowsDesktop.App 7.0.0

### What did you do?
I've recently updated a solution to .NET 7 and deployed it.
When I start the Kestrel hosted service, and then launch the WinForms client, everything works as expected.
When I then close the WinForms client, and open it again, it hangs.

### What did you expect to see?
The client connects to the service, when launched the second time as well.

### What did you see instead?
When launched a second time, the client hangs.
Looking at the client logs, the last entry was:
`Subchannel id '1' creating stream for...`
so it is missing 
`Serialized 'System.Byte[]' to 19 byte message...` and all the other logs from the first successful client launch.

The server logs does not contain any entries for the client's second launch.

Please find attached client and service logs.

### Anything else we should know about your project / environment?
I can only reproduce this when deployed. Locally it works fine.

Any ideas why the serialization and the rest of the calls does not happen in this case?

![image](https://user-images.githubusercontent.com/11263864/201915781-9a04f658-72ed-4cf0-89e5-787f540dbf1a.png)
[Client-FirstTime.log](https://github.com/grpc/grpc-dotnet/files/10012495/Client-FirstTime.log)
[Client-SecondTime.log](https://github.com/grpc/grpc-dotnet/files/10012494/Client-SecondTime.log)
[Service.log](https://github.com/grpc/grpc-dotnet/files/10012496/Service.log)

rafikiassumani-msft:
@Kobus-Smit  can you provide an app to help us reproduce this issue?

Kobus-Smit:
@rafikiassumani-msft I'll try to share a minimal reproducible example this weekend

JamesNK:
Closing because of lack of activity

Kobus-Smit:
Hi @JamesNK and @rafikiassumani-msft ,

[...]

PR #2047, opened 2023-02-13
Add additional connection and request logging to client

JamesNK:
Additional logging for https://github.com/grpc/grpc-dotnet/issues/1948

Also, implement DisposeAsync on StreamWrapper in as there is a situation where it is called.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
+++ b/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
@@ -80,10 +80,9 @@ internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDi
         _socketConnectedTimer = new Timer(OnCheckSocketConnection, state: null, Timeout.InfiniteTimeSpan, Timeout.InfiniteTimeSpan);
     }
 
-    public object Lock => _subchannel.Lock;
+    private object Lock => _subchannel.Lock;
     public BalancerAddress? CurrentAddress => _currentAddress;
     public TimeSpan? ConnectTimeout { get; }
-    public bool HasStream { get; }
 
     // For testing. Take a copy under lock for thread-safety.
     internal IReadOnlyList<ActiveStream> GetActiveStreams()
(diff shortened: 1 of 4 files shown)


src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -300,6 +309,7 @@ public async ValueTask<Stream> GetStreamAsync(BalancerAddress address, Cancellat
         lock (Lock)
         {
             _activeStreams.Add(new ActiveStream(address, socket, stream));
+            SocketConnectivitySubchannelTransportLog.StreamCreated(_logger, _subchannel.Id, address, _activeStreams.Count);
```
mgravell, line 312: logging inside the lock troubles me; is this intentional? would it perhaps be better to snag the `Count` inside the lock and log outside?

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -331,7 +341,7 @@ private void OnStreamDisposed(Stream streamWrapper)
                     if (t.Stream == streamWrapper)
                     {
                         _activeStreams.RemoveAt(i);
-                        SocketConnectivitySubchannelTransportLog.DisposingStream(_logger, _subchannel.Id, t.Address);
+                        SocketConnectivitySubchannelTransportLog.DisposingStream(_logger, _subchannel.Id, t.Address, _activeStreams.Count);
```
mgravell, line 344: hmmm, maybe ignore my last comment about lock; I can't see a nice / clean way of avoiding this one, so... if we're in that situation *anyway*, maybe just live with it

mgravell (approved):
added a question about logging inside the lock, and whether we care about that - I can't get hugely excited either way, as long as considered - looks fine generally

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
JamesNK: I don't think this is a concern. There is logging inside locks throughout gRPC and ASP.NET Core.

PR #2052, opened 2023-02-21
Wrap EventSource calls with IsEnabled

JamesNK:
Wrap event source calls with IsEnabled to avoid `Interlocked.Increment` calls inside the methods.

Small perf improvement:
* **Before** 956,933
* **After** 966,092

The perf improvement will be much bigger with AOT + ARM: https://github.com/dotnet/runtime/issues/73246

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
@@ -23,7 +23,7 @@ using Grpc.Core;
 
 namespace Grpc.AspNetCore.Server.Internal;
 
-internal class GrpcEventSource : EventSource
+internal sealed class GrpcEventSource : EventSource
 {
     public static readonly GrpcEventSource Log = new GrpcEventSource();
 
--- a/src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
@@ -16,6 +16,7 @@
 
 #endregion
 
+using System.Diagnostics;
 using System.Diagnostics.Tracing;
 using System.Runtime.CompilerServices;
 using Grpc.Core;
@@ -68,6 +69,8 @@ internal class GrpcEventSource : EventSource
     [Event(eventId: 1, Level = EventLevel.Verbose)]
     public void CallStart(string method)
     {
+        AssertEventSourceEnabled();
+
         Interlocked.Increment(ref _totalCalls);
         Interlocked.Increment(ref _currentCalls);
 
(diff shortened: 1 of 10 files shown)


src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs
```diff
@@ -297,9 +297,15 @@ private void LogCallEnd()
         }
         if (_status.StatusCode != StatusCode.OK)
         {
-            GrpcEventSource.Log.CallFailed(_status.StatusCode);
+            if (GrpcEventSource.Log.IsEnabled())
+            {
+                GrpcEventSource.Log.CallFailed(_status.StatusCode);
+            }
+        }
+        if (GrpcEventSource.Log.IsEnabled())
```
mgravell, line 305: since you're always going to test enabled here, is it worth pulling this into a local around L297, to avoid multiple checks?

JamesNK: IsEnabled implementation directly returns a field. IsEnabled is so cheap (it's probably inlined) that I don't think it's worth it.

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -882,19 +887,26 @@ private bool FinishCall(HttpRequestMessage request, bool diagnosticSourceEnabled
...
+                            GrpcEventSource.Log.CallDeadlineExceeded();
+                        }
                         _deadline = DateTime.MaxValue;
                     }
                 }
             }
 
             GrpcCallLog.GrpcStatusError(Logger, status.StatusCode, status.Detail);
-            GrpcEventSource.Log.CallFailed(status.StatusCode);
+            if (GrpcEventSource.Log.IsEnabled())
```
mgravell, line 900: ditto hoist covering 3 in a row here

mgravell (approved):
minor nits re avoidable repeated call; 👍 in principle, and I'll let you decide whether that change is needed

src/Grpc.AspNetCore.Server/Internal/GrpcEventSource.cs
```diff
@@ -23,7 +23,7 @@
 
 namespace Grpc.AspNetCore.Server.Internal;
 
-internal class GrpcEventSource : EventSource
+internal sealed class GrpcEventSource : EventSource
```
mgravell, line 26: 👍 , although with the `static readonly Log` I would *hope* that the JIT devirtualizes anyway

EgorBo: > 👍 , although with the `static readonly Log` I would _hope_ that the JIT devirtualizes anyway

Yes, jit does that for static readonly

PR #2055, opened 2023-02-24
Add .editorconfig and fix warnings

JamesNK:
editorconfig is copied from dotnet/aspnetcore


BrennanConroy (approved):
You could instead consider using the runtimes version:
https://github.com/dotnet/runtime/blob/main/eng/CodeAnalysis.src.globalconfig

JamesNK:
I prefer sticking closer to the ASP.NET Core one because we understand it more and it's easier to keep them in sync.

PR #2066, opened 2023-03-02
Upgrade version to 2.52.0-pre1 (on v2.52.x branch)

jtattermusch:


JamesNK (approved):

JamesNK (approved):

jtattermusch:
The test run was a flake, so I'll proceed, but it should still get looked at (looks like multiple tests try to bind the same port, which is inherently flawed).

jtattermusch:
The release was published: https://github.com/grpc/grpc-dotnet/releases/tag/v2.52.0-pre1 
(and nugets uploaded)

PR #2068, opened 2023-03-09
Improve error message with invalid headers

JamesNK:
Fixes https://github.com/grpc/grpc/issues/32577

The error message makes debugging which trailer has an invalid character easier.


JamesNK:
An error from an incoming header is less likely. That happening means the HTTP client has to have allowed sending an invalid header. Also, there isn't much to do about it in gRPC because Kestrel rejects the request before gRPC runs.

Good point about invalid response header. Added same logic there.

mgravell (approved):

Issue #2075, opened 2023-03-16
Issue after update to 2.52.0


plombaer:
Dear
Since the upgrade I'm getting the following error :
System.InvalidOperationException: 'Channel is configured with an HTTP transport doesn't support client-side load balancing or connectivity state tracking. The underlying HTTP transport must be a SocketsHttpHandler with no SocketsHttpHandler.ConnectCallback configured. The HTTP transport must be configured on the channel using GrpcChannelOptions.HttpHandler.'

My service is hosted in IIS using gRPC.Web with client load balancing.
All was working fine before the update.
Here is my client code:

```cs
            var credentials = CallCredentials.FromInterceptor((c, m) =>
            {
                m.Add("Authorization", "Bearer " + ...;
                return Task.CompletedTask;
            });
            var clientHandler = new SocketsHttpHandler();
            clientHandler.UseProxy = false;
            clientHandler.EnableMultipleHttp2Connections = false;
            clientHandler.SslOptions = new SslClientAuthenticationOptions
            {
                RemoteCertificateValidationCallback = ValidateServerCertificate,
            };
            var channel = GrpcChannel.ForAddress("dns:///...", new GrpcChannelOptions
            {
                HttpHandler = new GrpcWebHandler(new SubdirectoryHandler(clientHandler, "/...")),
                Credentials = ChannelCredentials.Create(ChannelCredentials.SecureSsl, credentials),

                //UnsafeUseInsecureChannelCallCredentials = true,
                ServiceConfig = new ServiceConfig
                {
                    LoadBalancingConfigs = { new RoundRobinConfig() }
                }
            });
```

JamesNK:
That error should only happen if you're configuring SocketsHttpHandler.ConnectCallback, which you aren't doing in the sample in your issue.

Could you create a reproduction and put it in a GitHub repo? I'll take a look.

plombaer:
[TestGrpcWeb.zip](https://github.com/grpc/grpc-dotnet/files/11008584/TestGrpcWeb.zip)
Here is my test project

JamesNK:
I didn't get that error. It worked for me.

I tested with Kestrel, so I needed to modify the sample slightly not to use TLS, modify the path, and use a static load balancer instead of DNS, but that shouldn't impact the error.

![image](https://user-images.githubusercontent.com/303201/226151974-c3b4749c-d526-467c-8710-c8cb6a0afadc.png)

plombaer:
[...]

PR #2087, opened 2023-04-05
Change proxy detection to log a message

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/2075. Still looking for a good solution. In the meantime, keep the old behavior and log a message.

This reverts the functional change in https://github.com/grpc/grpc-dotnet/pull/1863

### DIFF
--- a/src/Grpc.Net.Client/GrpcChannel.cs
+++ b/src/Grpc.Net.Client/GrpcChannel.cs
@@ -269,11 +269,11 @@ public sealed class GrpcChannel : ChannelBase, IDisposable
             // - HttpClient.DefaultProxy. Set via environment variables, e.g. HTTPS_PROXY.
             if (type == HttpHandlerType.SocketsHttpHandler)
             {
-                if (IsProxied(socketsHttpHandler, address, isSecure) is { } proxyUri)
+                if (IsProxied(socketsHttpHandler, address, isSecure))
                 {
-                    logger.LogInformation("A proxy is detected. The way the gRPC client creates connections can cause unexpected behavior when a proxy is configured. " +
-                        "To ensure a proxy is correctly used by the client, configure GrpcChannelOptions.HttpHandler to use HttpClientHandler. " +
-                        "Note that HttpClientHandler isn't compatible with load balancing.", proxyUri);
+                    logger.LogInformation("Proxy configuration is detected. How the gRPC client creates connections can cause unexpected behavior when configuring a proxy. " +
+                        "To ensure the client correctly uses a proxy, configure GrpcChannelOptions.HttpHandler to use HttpClientHandler. " +
+                        "Note that HttpClientHandler isn't compatible with load balancing.");
                 }
             }
 #else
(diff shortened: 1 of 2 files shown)


JamesNK:
@jtattermusch Can start 2.53.0 release process once this is merged.

src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -267,10 +267,14 @@ private static HttpHandlerContext CalculateHandlerContext(Uri address, bool isSe
...
-            if (IsProxied(socketsHttpHandler, address, isSecure))
+            if (type == HttpHandlerType.SocketsHttpHandler)
             {
-                type = HttpHandlerType.Custom;
-                connectTimeout = null;
+                if (IsProxied(socketsHttpHandler, address, isSecure) is { } proxyUri)
+                {
+                    logger.LogInformation("A proxy is detected. The way the gRPC client creates connections can cause unexpected behavior when a proxy is configured. " +
+                        "To ensure a proxy is correctly used by the client, configure GrpcChannelOptions.HttpHandler to use HttpClientHandler. " +
+                        "Note that HttpClientHandler isn't compatible with load balancing.", proxyUri);
```
BrennanConroy, line 276: Did you mean to add a structured log value in the string somewhere?

BrennanConroy (approved):
Comment about passing `proxyUri` to the log without having a place for it in the log message.

src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -267,10 +267,14 @@ private static HttpHandlerContext CalculateHandlerContext(Uri address, bool isSe
...
+                }
```
JamesNK, line 277:
```suggestion
                if (IsProxied(socketsHttpHandler, address, isSecure))
                {
                    logger.LogInformation("Proxy configuration is detected. How the gRPC client creates connections can cause unexpected behavior when configuring a proxy. " +
                        "To ensure the client correctly uses a proxy, configure GrpcChannelOptions.HttpHandler to use HttpClientHandler. " +
                        "Note that HttpClientHandler isn't compatible with load balancing.");
                }
```

src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -267,10 +267,14 @@ private static HttpHandlerContext CalculateHandlerContext(Uri address, bool isSe
...
             // - SocketsHttpHandler.Proxy. Set via app code.
             // - HttpClient.DefaultProxy. Set via environment variables, e.g. HTTPS_PROXY.
-            if (IsProxied(socketsHttpHandler, address, isSecure))
+            if (type == HttpHandlerType.SocketsHttpHandler)
             {
-                type = HttpHandlerType.Custom;
-                connectTimeout = null;
+                if (IsProxied(socketsHttpHandler, address, isSecure))
+                {
+                    logger.LogInformation("Proxy configuration is detected. How the gRPC client creates connections can cause unexpected behavior when configuring a proxy. " +
```
JamesNK, line 274:
```suggestion
                    logger.LogInformation("Proxy configuration is detected. How the gRPC client creates connections can cause unexpected behavior when a proxy is configured. " +
```

Issue #2103, opened 2023-04-27
Unable to diagnose root cause of `StatusCode.Internal` faults


jimgolfgti:
In the course of trying to diagnose why we are seeing occasional `StatusCode.DeadlineExceeded` whilst performing unary calls we configured logging of the `RetryCall.EvaluateRetry` result, see
https://github.com/grpc/grpc-dotnet/blob/6bf44820a2ef6af5845830728cd4b3d1a8053ed4/src/Grpc.Net.Client/Internal/Retry/RetryCall.cs#L201

When evaluating the retry when the `Status.StatusCode` is `StatusCode.Internal` we are unable to see the detailed reason, it is lost once the method returns due to the status being overwritten by the DeadlineExceeded evaluation result,
https://github.com/grpc/grpc-dotnet/blob/6bf44820a2ef6af5845830728cd4b3d1a8053ed4/src/Grpc.Net.Client/Internal/Retry/RetryCall.cs#L229-L233

Is it possible to change the log message to pass the entire `Status` object?
Looking at the implementation it already serialises all the fields in its `ToString` override,
https://github.com/grpc/grpc-dotnet/blob/6bf44820a2ef6af5845830728cd4b3d1a8053ed4/src/Grpc.Core.Api/Status.cs#L87-L95

JamesNK:
This is what I see when I run the `Unary_ExceedRetryAttempts_Failure` test and look at the logs:

```
0.036s Grpc.Net.Client.Internal.RetryCall - Trace: Starting retry worker.
0.037s Grpc.Net.Client.Internal.RetryCall - Debug: Starting attempt 1.
0.048s Grpc.Net.Client.Internal.GrpcCall - Debug: Starting gRPC call. Method type: 'Unary', URI: 'http://127.0.0.1:50050/DynamicService/d866aca4-4ea5-46e7-923e-34d0cddba63d'.
0.088s Grpc.Net.Client.Internal.GrpcCall - Trace: Serialized 'Streaming.DataMessage' to 0 byte message.
0.091s Grpc.Net.Client.Internal.RetryCall - Trace: Message with 5 bytes added to the buffer. There are 5 bytes buffered for this call.
0.092s Grpc.Net.Client.Internal.GrpcCall - Debug: Sending message.
0.093s Grpc.Net.Client.Internal.GrpcCall - Trace: Message sent.
0.155s Grpc.Net.Client.Internal.GrpcCall - Trace: Response headers received.
[...]

PR #2104, opened 2023-04-27
Output detailed Status to retry debug log

jimgolfgti:
As per #2103 

This small change passes the entire struct to the debug logging delegate. If debug logging is enabled we will capture the detailed status, including cause exception if available.


JamesNK:
As discussed in https://github.com/grpc/grpc-dotnet/issues/2103, I don't think this is necessary. The PR can be reopened in the future if the situation changes.

PR #2107, opened 2023-04-28
Make CancellationToken available in call credentials interceptor

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/2108

Adds `AuthInterceptorContext.CancellationToken` property. The context is passed to call credential interceptors.

This change allows call credentials that do async work to be canceled when the gRPC call is canceled:

```csharp
var callCredentials = CallCredentials.FromInterceptor(async (context, metadata) =>
{
    metadata["authorize"] = await GetTokenAsync(context.CancellationToken);
});
```

### DIFF
--- a/src/Grpc.Net.Client/Internal/GrpcProtocolHelpers.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcProtocolHelpers.cs
@@ -1,4 +1,4 @@
-﻿#region Copyright notice and license
+#region Copyright notice and license
 
 // Copyright 2019 The gRPC Authors
 //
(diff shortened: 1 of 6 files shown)


src/Grpc.Net.Client/Internal/GrpcProtocolHelpers.cs
```diff
@@ -252,38 +273,39 @@ internal static AuthInterceptorContext CreateAuthInterceptorContext(Uri baseAddr
...
-        CallCredentials credentials)
+        CallCredentials credentials,
+        CancellationToken cancellationToken)
     {
         credentials.InternalPopulateConfiguration(configurator, null);
 
         if (configurator.Interceptor != null)
         {
-            var authInterceptorContext = GrpcProtocolHelpers.CreateAuthInterceptorContext(channel.Address, method);
+            configurator.CachedContext ??= CreateAuthInterceptorContext(channel.Address, method, cancellationToken);
```
BrennanConroy, line 291: This looked scary at first, but the call sites don't reuse the `DefaultCallCredentialsConfigurator` for different channel + method + cancellation combos.

Can we make that more obvious somehow so this code doesn't look scary?

e.g.

```
ReadCredentialMetadata(...)
{
    ReadCredentialMetadataCore(...);
    configurator.CachedContext = null;
}

// Rename current method
ReadCredentialMetadataInner(...)
{
}
```

test/Grpc.Net.Client.Tests/CallCredentialTests.cs
```diff
@@ -79,19 +80,67 @@ public async Task CallCredentialsWithHttps_MetadataOnRequest()
...
 
         // Act
+        var tcs = new TaskCompletionSource<object?>(TaskCreationOptions.RunContinuationsAsynchronously);
         var callCredentials = CallCredentials.FromInterceptor(async (context, metadata) =>
         {
             // The operation is asynchronous to ensure delegate is awaited
-            await Task.Delay(50);
+
+            // Ensure task hasn't been completed.
+            Assert.False(tcs.Task.IsCompleted);
```
BrennanConroy, line 89: This is a bit odd. I think in theory the compiler or synccontext could choose to run this asynchronously before it gets here in which case it would be a race.

Generally you would instead use 2 tcs's to ensure order.

src/Grpc.Net.Client/Internal/GrpcProtocolHelpers.cs
JamesNK: I've updated comments and renamed some fields to make it clearer what is going on.

test/Grpc.Net.Client.Tests/CallCredentialTests.cs
JamesNK: I _think_ the test was fine. The tcs couldn't be completed until after the auth interceptor awaits. The assert was to verify that behavior.

But I agree the test looks a little weird. And it's brittle. I've switched it to use a sync point.

BrennanConroy (approved):

PR #2002, opened 2022-12-30
Add host override to resolvers

JamesNK:
Fix https://github.com/grpc/grpc-dotnet/issues/2001

PR #2109, opened 2023-05-01
Remove host override from container sample

JamesNK:
https://github.com/grpc/grpc-dotnet/pull/2002 added a host override to the DNS resolver. Usually this is good. However, in the container sample we want to access the original IP address.

### DIFF
--- a/examples/Container/Frontend/Balancer/ConfigurableResolverFactory.cs
+++ b/examples/Container/Frontend/Balancer/ConfigurableResolverFactory.cs
@@ -1,4 +1,4 @@
-﻿#region Copyright notice and license
+#region Copyright notice and license
 
 // Copyright 2019 The gRPC Authors
 //
@@ -24,6 +24,8 @@ namespace Frontend.Balancer
 {
     public class ConfigurableResolverFactory : ResolverFactory
     {
+        private static readonly BalancerAttributesKey<string> HostOverrideKey = new BalancerAttributesKey<string>("HostOverride");
+
         private readonly ResolverFactory _innerResolverFactory;
         private readonly BalancerConfiguration _balancerConfiguration;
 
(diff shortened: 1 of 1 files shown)


amcasey:
Is this a pattern we expect actual users to (have to?) apply?  Seems convoluted.

JamesNK:
Users are unlikely ever to want to do this. The sample does this because it displays the host IP to make it clear that load balancing is happening.

amcasey:
> Users are unlikely ever to want to do this. The sample does this because it displays the host IP to make it clear that load balancing is happening.

Maybe include a comment to that effect so people don't start copying it into their own apps without understanding?

amcasey:
Consume #2110?

JamesNK:
The sample references a NuGet package. It needs to be released and published before the sample can use it. That is a couple of months away.

examples/Container/Frontend/Balancer/ConfigurableResolverFactory.cs
```diff
@@ -101,6 +103,11 @@ private void RaiseResult(ResolverResult result)
                         // DNS results change order between refreshes.
                         // Explicitly order by host to keep result order consistent.
                         var orderedAddresses = result.Addresses.OrderBy(a => a.EndPoint.Host).ToList();
+                        // Remove host override from addresses so the destination IP address is available.
```
JamesNK, line 106:
```suggestion
                        // Remove host override from addresses so the destination IP address is available.
                        // The sample does this because the server returns the IP address to the client.
                        // This makes it clear that gRPC calls are balanced between pods.
```

examples/Container/Frontend/Balancer/ConfigurableResolverFactory.cs
```diff
@@ -24,6 +24,8 @@ namespace Frontend.Balancer
 {
     public class ConfigurableResolverFactory : ResolverFactory
     {
+        private static readonly BalancerAttributesKey<string> HostOverrideKey = new BalancerAttributesKey<string>("HostOverride");
```
amcasey, line 27: I'm a little curious why we don't work directly with `"HostOverride"`, rather than wrapping and unwrapping it, but I'm assuming this is idiomatic.

amcasey (approved):
Seems reasonable.

examples/Container/Frontend/Balancer/ConfigurableResolverFactory.cs
JamesNK: Its a pattern copied from the `HttpRequestOptions` collection. It's designed to make the collection feel a little more strongly typed.

Issue #2119, opened 2023-05-10
Race condition when disposing of channels while cancelling calls.


amanda-tarafa:
I think there might be a race condition if a channel is disposed at the same time that one of its active calls is cancelled separately. I think this happens only for server or bidi streaming calls. The original report (https://github.com/googleapis/google-cloud-dotnet/issues/10318) comes from a Google.Cloud.PubSub.V1 user who has provided a repro and logs. A couple of us on the Google side haven't been able to reproduce but I think my description below of the race conditions is accurate, and it's also supported by the user provided logs (ignoring the Google.Cloud.PubSub.V1 part of the logs.)

- [Grpc.Net.Client.GrpcChannel.Dispose](https://github.com/grpc/grpc-dotnet/blob/38738af700a3180e87054f922f953e67d9cdd82b/src/Grpc.Net.Client/GrpcChannel.cs#L741-L754) disposes of active calls **within a lock**.
https://github.com/grpc/grpc-dotnet/blob/38738af700a3180e87054f922f953e67d9cdd82b/src/Grpc.Net.Client/GrpcChannel.cs#L741-L754

- [Grpc.Net.Client.GrpcCall.Dispose](https://github.com/grpc/grpc-dotnet/blob/38738af700a3180e87054f922f953e67d9cdd82b/src/Grpc.Net.Client/Internal/GrpcCall.cs#L176-L188) basically calls [GrpcCall.Cleanup](https://github.com/grpc/grpc-dotnet/blob/38738af700a3180e87054f922f953e67d9cdd82b/src/Grpc.Net.Client/Internal/GrpcCall.cs#L196) with a status of cancel.
- GrpcCall.Cleanup then sets the call TaskCompletionSource to cancelled.
  https://github.com/grpc/grpc-dotnet/blob/38738af700a3180e87054f922f953e67d9cdd82b/src/Grpc.Net.Client/Internal/GrpcCall.cs#L211
- But, if this GrpcCall was already being cancelled elsewhere (i.e. in a different thread) setting the TaskCompletionSource to cancelled [blocks until the cancellation callbacks have completed elsewhere](https://github.com/dotnet/runtime/blob/0743c542a0aed2928d4f65653d54958b5e861d11/src/libraries/System.Private.CoreLib/src/System/Threading/CancellationTokenRegistration.cs#L38). Note that this **blocks holding the chanell lock**.

In the meantime the thread that first cancelled the GrpcCall has been able to:

[...]

PR #2120, opened 2023-05-10
Fix bugs related to channel dispose while there are active calls

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2119

~@amanda-tarafa I tried to recreate this in a unit test without success. I went with your suggestion. Also, I modified the dispose check when starting gRPC calls to remove the chance of a new active call being added after dispose.~

**Update:** Deadlock reproduced in a unit test. Fix in PR confirmed to work.

---

This PR fixes a number of bugs related to disposing a channel while there are ongoing active calls:

* Fix potential deadlock when a call and channel are disposed at the same time.
* Fix race that could allow call to be added to the channel's active call collection after dispose.
* Fix hedge/retry calls not immediately throwing `ObjectDisposedException` after channel dispose.
* Fix hedge/retry calls not immediately canceling after channel dispose.

### DIFF
--- a/src/Grpc.Net.Client/Configuration/HedgingPolicy.cs
+++ b/src/Grpc.Net.Client/Configuration/HedgingPolicy.cs
@@ -1,4 +1,4 @@
-﻿#region Copyright notice and license
+#region Copyright notice and license
 
 // Copyright 2019 The gRPC Authors
 //
@@ -37,7 +37,7 @@ public sealed class HedgingPolicy : ConfigObject
     internal const string HedgingDelayPropertyName = "hedgingDelay";
     internal const string NonFatalStatusCodesPropertyName = "nonFatalStatusCodes";
 
-    private ConfigProperty<Values<StatusCode, object>, IList<object>> _nonFatalStatusCodes =
+    private readonly ConfigProperty<Values<StatusCode, object>, IList<object>> _nonFatalStatusCodes =
         new(i => new Values<StatusCode, object>(i ?? new List<object>(), s => ConvertHelpers.ConvertStatusCode(s), s => ConvertHelpers.ConvertStatusCode(s.ToString()!)), NonFatalStatusCodesPropertyName);
 
     /// <summary>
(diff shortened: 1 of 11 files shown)


amanda-tarafa:
The changes look good. And the test as well. I also haven't been able to reproduce, I pulled your test and played around with it even, but nothing.
I'm pretty certain the race condition is there, but it'd definetely be nice to see it happen.

amanda-tarafa:
I've continued trying to reproduce this in Grpc.Net.Client directly (no Google library involved) and I haven't been able to. The code I've been using can be see in https://github.com/amanda-tarafa/grpc-dotnet/pull/1. My test is very similar to the test in this PR, only a few tweaks here and there.

There are changes to production code, mostly logging, for the sake of keeping track of things, but I did try to mimick loosing CPU where I think mattered, without success in reproducing the deadlock.

Basically, I can't find any sensible way to trigger the Dispose thread blocking [waiting on the cancellation token registration callbacks to be executed elsewhere](https://github.com/dotnet/runtime/blob/f1819bd78f0e7506b8c3c7f90b47aa41b0e22c6f/src/libraries/System.Private.CoreLib/src/System/Threading/CancellationTokenRegistration.cs#L56), which is what we've seen in https://github.com/googleapis/google-cloud-dotnet/issues/10318. I'm a little out of my depth here, but I'll try to pick this up again at some point next week.

amanda-tarafa:
Sracth my last comment, I can now reproduce this 100% reliably (of the 20 or so times I've executed my test, we might still get the ocassional success).
See https://github.com/amanda-tarafa/grpc-dotnet/pull/1 for the unit test (no production code changes there whatsoever).

@JamesNK the main issue with your test is that the HTTP handler does not return so GrpcCall.RunCall is awaiting for the response instead of for the GrpcCall.CallTask whose continuations are the ones that block the dispose thread. Hopefully you can run this with your changes and confirm that releasing the lock earlier is indeed a fix.

JamesNK:
Good find. I've updated the test.

I confirmed the test deadlocked on my machine before the changes and succeeded after the changes.

BrennanConroy (approved):

PR #2121, opened 2023-05-11
Add IReadOnlyDictionary to BalancerAttributes collection

JamesNK:
Make this collection feel a little more idiomatic to use. Some extension methods in .NET hang off the readonly interface and assume all dictionaries implements IDictionary and IReadOnlyDictionary.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/BalancerAttributes.cs
+++ b/src/Grpc.Net.Client/Balancer/BalancerAttributes.cs
@@ -17,6 +17,7 @@
 #endregion
 
 #if SUPPORT_LOAD_BALANCING
+using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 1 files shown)


src/Grpc.Net.Client/Balancer/BalancerAttributes.cs
```diff
@@ -29,7 +30,7 @@ namespace Grpc.Net.Client.Balancer;
 /// Note: Experimental API that can change or be removed without any prior notice.
 /// </para>
 /// </summary>
-public sealed class BalancerAttributes : IDictionary<string, object?>
+public sealed class BalancerAttributes : IDictionary<string, object?>, IReadOnlyDictionary<string, object?>
```
mitchdenny, line 33: Assuming we don't need an API review for adding this interface.

mitchdenny (approved):
LGTM but I don't know whether this qualifies for needing an API review or not. But seems reasonable.

JamesNK:
This type is basically a copy of `HttpRequestOptions`. And they've already done the API review 😄 

https://github.com/dotnet/runtime/issues/68149

mitchdenny (approved):

amcasey (approved):

PR #2129, opened 2023-05-16
Fix capturing ExecutionContext by timers and background tasks

JamesNK:
Capturing `ExecutionContext` can cause a memory leak of async locals values.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Subchannel.cs
+++ b/src/Grpc.Net.Client/Balancer/Subchannel.cs
@@ -1,4 +1,4 @@
-﻿#region Copyright notice and license
+#region Copyright notice and license
 
 // Copyright 2019 The gRPC Authors
 //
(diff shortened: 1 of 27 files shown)


src/Grpc.Net.Client/Balancer/Subchannel.cs
```diff
@@ -233,7 +233,21 @@ public void RequestConnection()
...
-        _ = ConnectTransportAsync();
+        // Don't capture the current ExecutionContext and its AsyncLocals onto the connect
+        bool restoreFlow = false;
+        if (!ExecutionContext.IsFlowSuppressed())
+        {
+            ExecutionContext.SuppressFlow();
+            restoreFlow = true;
+        }
+
+        _ = Task.Run(ConnectTransportAsync);
```
ReubenBond, line 244: An alternative would be to put an `await Task.Yield()` at the top of `ConnectTransportAsync`, or to leave it as-is.
This is fine, too.

ReubenBond (approved):

Issue #2141, opened 2023-05-25
gRPC health checks in kubernetes scenario - unable to execute on demand only subset of healthchecks


banatm-abb:
### What version of gRPC and what language are you using?
2.53.0

### What operating system (Linux, Windows,...) and version?
macos

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
6.0.16

### What did you do?
Trying to configure kubernetes grpc healthchecks for my service.
I want to have different healthchecks executed on demand for startup and liveness probe.

I've tried to use the following configuration:

```c#
        services
            .AddGrpcHealthChecks(o =>
            {
                // startup probe checks dependencies
                o.Services.MapService("startup", r => r.Tags.Contains("startup"));
                // startup probe checks dependencies
                o.Services.MapService( "liveness", r => r.Tags.Contains("liveness"));
                o.Services.MapService("", r => r.Tags.Contains("liveness");
                o.UseHealthChecksCache = false;
            })
            ... //definitions here

           //auto update just liveness ones
           services.Configure<HealthCheckPublisherOptions>(options =>
           {
             options.Predicate = (r) => r.Tags.Contains("liveness");
           });
```

### What did you expect to see?
`grpc_cli call IP:PORT grpc.health.v1.Health/Check "service: 'startup'"` will run only checks tagged with `startup`
`grpc_cli call IP:PORT grpc.health.v1.Health/Check "service: 'liveness'"` will run only checks tagged with `liveness`

### What did you see instead?
`grpc_cli call IP:PORT grpc.health.v1.Health/Check "service: 'startup'"` doesnt not run `startup` tagged checks at all!

### Code hint
In the newest [code](https://github.com/grpc/grpc-dotnet/blame/6bc44e97d73244de9da410e6cccc7409b7c173df/src/Grpc.AspNetCore.HealthChecks/Internal/HealthServiceIntegration.cs#L77) I've found that the CheckHealthAsync is executed with different predicate than GetStatus. Is it by design?
```
        HealthCheckResponse.Types.ServingStatus status;
        if (_grpcHealthCheckOptions.Services.TryGetServiceMapping(service, out var serviceMapping))
        {
            var result = await _healthCheckService.CheckHealthAsync(_healthCheckOptions.Predicate, cancellationToken);
            status = HealthChecksStatusHelpers.GetStatus(result, serviceMapping.Predicate);
        }
```

JamesNK:
[...]

PR #2142, opened 2023-05-30
Refactor healthchecks service mapping to support filtering on check

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2141

* Previously, all health checks were run when `Check` was called, and then they were filtered. This PR changes the filter to happen before health checks are run. Now health checks can be selectly run depending on `Check`'s service name argument.
* Previous health result filtering is now obsolete. The new predicate is based on the information available when a health check is registered (its name and tags).
* Add some logging to the health check publisher.

API changes:

```diff
namespace Grpc.AspNetCore.HealthChecks;

+public sealed class HealthCheckMapContext
+{
+   public HealthCheckMapContext(string name, IEnumerable<string> tags);
+   public string Name { get; }
+   public IEnumerable<string> Tags { get; }
+}

+[Obsolete($"HealthResult is obsolete and will be removed in a future release. Use {nameof(HealthCheckMapContext)} instead.")]
public sealed class HealthResult
{
    public HealthResult(string name, IEnumerable<string> tags, HealthStatus status, string? description, TimeSpan duration, Exception? exception, IReadOnlyDictionary<string, object> data);
    public string Name { get; }
    public IEnumerable<string> Tags { get; }
    public HealthStatus Status { get; }
    public string? Description { get; }
    public IReadOnlyDictionary<string, object> Data { get; }
    public Exception? Exception { get; }
    public TimeSpan Duration { get; }
}

public sealed class ServiceMapping
{
+   [Obsolete("This constructor is obsolete and will be removed in the future. Use ServiceMapping(string name, Func<HealthCheckRegistration, bool> predicate) to map service names to .NET health checks.")]
    public ServiceMapping(string name, Func<HealthResult, bool> predicate);
+   public ServiceMapping(string name, Func<HealthCheckMapContext, bool> predicate);
+   public Func<HealthCheckMapContext, bool>? HealthCheckPredicate { get; }
+   [Obsolete($"This member is obsolete and will be removed in the future. Use {nameof(HealthCheckPredicate)} to map service names to .NET health checks.")]
    public Func<HealthResult, bool>? Predicate { get; }
}

-public sealed class ServiceMappingCollection : IEnumerable<ServiceMapping>
+public sealed class ServiceMappingCollection : ICollection<ServiceMapping>
{
+   [Obsolete("This method is obsolete and will be removed in the future. Use Map(string name, Func<HealthCheckRegistration, bool> predicate) to map service names to .NET health checks.")]
    public void MapService(string name, Func<HealthResult, bool> predicate);
+   public void Map(string name, Func<HealthCheckMapContext, bool> predicate);
}
```

Usage:

```csharp
builder.Services.AddGrpcHealthChecks(o =>
{
    o.Services.Map("greet.Greeter", c => c.Tags.Contains("greeter"));
    o.Services.Map("count.Counter", c => c.Tags.Contains("counter"));
});
```

### DIFF
--- a/src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksPublisher.cs
+++ b/src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksPublisher.cs
@@ -50,7 +50,7 @@ internal sealed class GrpcHealthChecksPublisher : IHealthCheckPublisher
             {
                 serviceEntries = serviceEntries.Where(entry =>
                 {
-                    var context = new HealthCheckFilterContext(entry.Key, entry.Value.Tags);
+                    var context = new HealthCheckMapContext(entry.Key, entry.Value.Tags);
                     return serviceMapping.HealthCheckPredicate(context);
                 });
             }
--- a/src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksPublisher.cs
+++ b/src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksPublisher.cs
@@ -16,6 +16,7 @@
 
 #endregion
 
+using System.Linq;
 using Grpc.Health.V1;
 using Grpc.HealthCheck;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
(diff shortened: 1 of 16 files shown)


src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksPublisher.cs
```diff
@@ -16,32 +16,86 @@
...
     public Task PublishAsync(HealthReport report, CancellationToken cancellationToken)
     {
-        foreach (var registration in _options.Services)
+        Log.EvaluatingPublishedHealthReport(_logger, report.Entries.Count, _options.Services.Count);
+
+        List<KeyValuePair<string, HealthReportEntry>>? serviceEntries = null;
+        foreach (var serviceMapping in _options.Services)
         {
-            var resolvedStatus = HealthChecksStatusHelpers.GetStatus(report, registration.Predicate);
+            serviceEntries ??= new();
```
gfoidl, line 47:
```suggestion
```
Together with the suggestion above?

src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksPublisher.cs
```diff
@@ -16,32 +16,86 @@
...
+            if (serviceMapping.HealthCheckPredicate != null)
+            {
+                for (var i = serviceEntries.Count - 1; i >= 0; i--)
+                {
+                    var entry = serviceEntries[i];
+                    var registration = new HealthCheckFilterContext(entry.Key, entry.Value.Tags);
+
+                    if (!serviceMapping.HealthCheckPredicate(registration))
+                    {
+                        serviceEntries.RemoveAt(i);
```
gfoidl, line 59: `RemoveAt` copies the elements after `i` to be contiguous again. So if you'd like to avoid these copying maybe type the list as `List<(KeyValuePair<string, HealthReportEntry> Entry, bool Publish)>` and use the `Publish`-flag is indicator -- thus in the Linq a additional `.Where(e => e.Publish)` is needed too.

JamesNK:
@gfoidl I've simplified the code in the area you commented on. The publisher runs every couple of seconds so I just want to keep it simple.

src/Grpc.AspNetCore.HealthChecks/ServiceMappingCollection.cs
```diff
@@ -25,7 +25,7 @@ namespace Grpc.AspNetCore.HealthChecks;
 /// <summary>
 /// A collection of <see cref="ServiceMapping"/> used to map health results to gRPC health checks services.
 /// </summary>
-public sealed class ServiceMappingCollection : IEnumerable<ServiceMapping>
+public sealed class ServiceMappingCollection : ICollection<ServiceMapping>
```
mgravell, line 28: nit: would `IReadOnlyCollection<T>` be a better choice here? do you need mutators?

mgravell (approved):
LGTM; my only comment was something I then deleted

PR #2148, opened 2023-06-02
Call Complete on SubchannelCallTracker after HttpContent has been disposed

JamesNK:
Update on https://github.com/grpc/grpc-dotnet/pull/2139

* Adds unit tests.
* Fixes scenario a streaming call is never used (the response stream isn't read from and the call isn't disposed) and the server aborts the call.

FYI @Khazuar

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Internal/BalancerHttpHandler.cs
+++ b/src/Grpc.Net.Client/Balancer/Internal/BalancerHttpHandler.cs
@@ -149,13 +149,18 @@ internal class BalancerHttpHandler : DelegatingHandler
         {
             var responseMessage = await responseMessageTask.ConfigureAwait(false);
 
-            // TODO(JamesNK): This doesn't take into account long running streams.
-            // If there is response content then we need to wait until it is read to the end
-            // or the request is disposed.
-            result.SubchannelCallTracker?.Complete(new CompletionContext
+            if (result.SubchannelCallTracker is not null)
             {
-                Address = address
-            });
+                if (responseMessage.Content is not null)
+                {
+                    responseMessage.Content = new HttpContentWrapper(responseMessage.Content,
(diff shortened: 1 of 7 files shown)


JamesNK:
I looked at this more, and unfortunately, I don't think it is possible to accurately track when the response ends in all situations in .NET.

Work does work is getting notifications when:

1. The client aborts the response.
2. The client reads to the end of the response.

The problem is there isn't a way to get a notification from the server. For example, there isn't a way to get a notification that the server has aborted the request if the client abandons using it, such as never disposes the response, or stops reading the response stream before the end.

The only way I can see how to get that information is if the client has a pending read call on the response stream. A way to get the notification could be always to have a `Stream.ReadAsync` pending on the real response stream. If it succeeds, then write the data to a buffer that the client then reads from with another stream. If it fails, then the response is in a bad state so report it is finished. However, this idea doesn't work if the server is constantly sending data and the client isn't reading it. The buffer grows indefinitely.

Khazuar:
Mhh .. that's suboptimal. I'd have argued that clients are supposed to call `Dispose` on the response whenever they are done with it, including the streamed data. As far as I understand there would cover server-side aborts as well, correct? I do understand though, that this is not super obvious to users.

JamesNK:
I don't know of one. I think a new feature is needed in HttpClient.

Khazuar:
As far as I can tell, there is no finalizer/destructor defined for `HttpContent` so far (though the infrastructure for a proper Dispose-Pattern is in place, even calling `GC.SuppressFinalize`). Couldn't we just add a destructor to the `HttpContentWrapper`? It would only fire if the user doesn't call `Dispose` and would guarantee that `SubchannelCallTracker.Complete` is called eventually. And for those who properly dispose of their response it will even work quite timely.

Issue #2158, opened 2023-06-12
InvalidOperationException when configuring gRPC client (Grpc.Net.ClientFactory)


jviau:
### What version of gRPC and what language are you using?

C# / .NET: `Grpc.Net.ClientFactory/2.53.0`

### What operating system (Linux, Windows,...) and version?

Applicable to all OS/versions

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)

dotnet 7

### What did you do?
If possible, provide a recipe for reproducing the error. Try being specific and include code snippets if helpful.

It appears any extension method from [GrpcHttpClientBuilderExtensions](https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.Net.ClientFactory/GrpcHttpClientBuilderExtensions.cs) do not work when using `.AddGrpcClient<T>()` (and **NOT** `.AddGrpcClient<T>(Action<GrpcClientFactoryOptions>)`). These methods will always throw `InvalidOperationException` from `ValidateGrpcClient`.

The reason being without using the overload which includes `Action<GrpcClientFactoryOptions>`, the service `ConfigureNamedOptions<GrpcClientFactoryOptions>` is never added, and so the validation check will always fail.

The fix should be as simple as changing `.AddGrpcClient<T>()` to call the action-accepting overload. IE: `.AddGrpcClient(_ => { })`.

### What did you expect to see?

Able to use `.AddGrpcClient<T>().AddInterceptor(...)`

### What did you see instead?

`InvalidOperationException` thrown from `AddInterceptor` call.

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

See [TROUBLESHOOTING.md](https://github.com/grpc/grpc-dotnet/blob/master/TROUBLESHOOTING.md) for how to diagnose problems better.

### Anything else we should know about your project / environment?

This is easily worked around for end users by using the other overload. However, my situation is I am adding an SDK extension to the Azure Functions isolated dotnet worker, and so this workaround needs to be passed on to our customers, which is not ideal.

### Side note

Also consider updating `ValidateGrpcClient` to use `[CallerMemberName] string? caller = null` instead of hardcoding `nameof(AddInteceptor)`

PR #2159, opened 2023-06-13
Fix client factory extension method validation issues

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2158

* Extension methods now work when AddGrpcClient is called without configuration
* Extension methods now report the failing method name correctly (I made it explicit instead of an attribute to avoid someone unknowingly accidentally regressing the error name)

### DIFF
--- a/src/Grpc.Net.ClientFactory/GrpcHttpClientBuilderExtensions.cs
+++ b/src/Grpc.Net.ClientFactory/GrpcHttpClientBuilderExtensions.cs
@@ -1,4 +1,4 @@
-﻿#region Copyright notice and license
+#region Copyright notice and license
 
 // Copyright 2019 The gRPC Authors
 //
@@ -47,7 +47,7 @@ public static class GrpcHttpClientBuilderExtensions
             throw new ArgumentNullException(nameof(configureChannel));
         }
 
-        ValidateGrpcClient(builder);
+        ValidateGrpcClient(builder, nameof(ConfigureChannel));
 
         builder.Services.AddTransient<IConfigureOptions<GrpcClientFactoryOptions>>(services =>
         {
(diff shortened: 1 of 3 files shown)


JamesNK:
FYI @jviau

You can temporarily work around the issue by using an AddGrpcClient overload that includes a configuration parameter and provide an empty one.

jviau (approved):
Thanks for the quick fix! And yes, I am using the `.AddGrpcClient(_ => { })` for now.

src/Grpc.Net.ClientFactory/GrpcHttpClientBuilderExtensions.cs
```diff
@@ -379,7 +379,7 @@ public static IHttpClientBuilder ConfigureGrpcClientCreator(this IHttpClientBuil
         return builder;
     }
 
-    private static void ValidateGrpcClient(IHttpClientBuilder builder)
+    private static void ValidateGrpcClient(IHttpClientBuilder builder, string caller)
```
mgravell, line 382: would this be cleaner as `[CallerMemberName] string caller = null!` ?

mgravell (approved):
approved with one optional suggestion

edit: I just noticed the "(I made it explicit instead of an attribute to avoid someone unknowingly accidentally regressing the error name)"  (sorry, words are hard!) - noted, so: you considered it, dismissed it, and gave a reason; fine

src/Grpc.Net.ClientFactory/GrpcHttpClientBuilderExtensions.cs
JamesNK: I didn’t because there aren’t tests for each name, and someone could unknowingly move this method to a child method and have the wrong name.

Issue #2122, opened 2023-05-11
Streams sporadically throw unexpected exceptions


Schaeri:
In our testing, we have found that when using GRPC streams, the expected exceptions are not being delivered to the client. This occurs sporadically and very infrequently.  

The following attached code example demonstrates the problem with a simple console application. The program can be run either via VS, Rider, run.bat or dotnet run. 

[GrpcServiceStreamExceptionIssue.zip](https://github.com/grpc/grpc-dotnet/files/11451622/GrpcServiceStreamExceptionIssue.zip)

What does the example do:

- Starts a GRPC service
- Creates a client
- And calls in a loop a method that receives a GRPC stream
- This service method throws an exception with the status code 'PreConditionFailed'.
- If an error other than the expected one is found, the program stops execution and writes the result to the console.

To reproduce the error I need 20000 to 400000 loops. So the example runs for some time until the error occurs. 

Instead of the expected exception the following error occurs: IncompleteSendRpcException with the following message A message could not be sent because the server had already terminated the connection.
The example then gives the following output:

`Start GRPC Service
Wait until service is ready
Start test: Try to get wrong exception
Wrong exception found
Expected: (StatusCode FailedPrecondition + Detail TEST)
Found exception: IncompleteSendRpcException A message could not be sent because the server had already terminated the connection; this exception can be suppressed by specifying the IgnoreStreamTermination flag when creating the
CallContext
Number of test runs: 117211
Stop service`

Version Infos:

- protobuf-net.Grpc 1.1.1
- protobuf-net.Grpc.AspNetCore 1.1.1 
- Grpc.Net.Client 2.53.0
- .Net SDK 7.0.203
- Test System Windows 11 sowie Windows Server 2016

JamesNK:
The message says the connection is unexpectedly closed. Are you gracefully closing the gRPC stream? I have seen errors where streams aren't gracefully closed on the client and server, leading to a bad state that causes the server to abort the connection.

See https://learn.microsoft.com/en-us/aspnet/core/grpc/client?view=aspnetcore-7.0#bi-directional-streaming-call

Schaeri:
An example project demonstrates the problem. In the service method nothing more is done than throwing an exception.  

`
public Task RunAsync(IAsyncEnumerable<TestRequest> request, CallContext context)
{
      throw new RpcException(new(StatusCode.FailedPrecondition, "TEST"));
} 
`

The problem occurs only sporadically after many attempts.

[...]

PR #2170, opened 2023-06-21
WriteAsync cancellation throws an error with the calls completed status if possible

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2122

If a call is completed during `Stream.WriteAsync`, the cancellation error was thrown. This has been changed to throw the call's completed status if available.

### DIFF
--- a/src/Grpc.Net.Client/Internal/StreamExtensions.cs
+++ b/src/Grpc.Net.Client/Internal/StreamExtensions.cs
@@ -1,4 +1,4 @@
-﻿#region Copyright notice and license
+#region Copyright notice and license
 
 // Copyright 2019 The gRPC Authors
 //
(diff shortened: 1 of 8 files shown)


mgravell (approved):

src/Grpc.Net.Client/Internal/StreamExtensions.cs
```diff
@@ -342,24 +342,41 @@ private static bool ReadCompressedFlag(byte flag)
...
         }
     }
+
+    private static bool TryCreateCallCompleteException(Exception originalException, GrpcCall call, [NotNullWhen(true)] out Exception? exception)
+    {
+        // The call may have been completed while WriteAsync was running and caused WriteAsync to throw.
+        // In this situation, report the call's completed status.
+        //
+        // Replace exception with the status error if:
+        // 1. The exception type is thrown Stream.WriteAsync if the call was completed during a write, and
```
JamesNK, line 370:
```suggestion
        // 1. The original exception is one Stream.WriteAsync throws if the call was completed during a write, and
```

PR #2186, opened 2023-07-08
Remove System.Memory dependency on .NET Standard 2.1 in Grpc.Core.Api.

teo-tsirpanis:
It's not needed; the types this package provides are inbox.

### DIFF
--- a/src/Grpc.Core.Api/Grpc.Core.Api.csproj
+++ b/src/Grpc.Core.Api/Grpc.Core.Api.csproj
@@ -19,8 +19,8 @@
     <Compile Include="..\Shared\CodeAnalysisAttributes.cs" Link="Internal\CodeAnalysisAttributes.cs" />
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="System.Memory" Condition="'$(TargetFramework)' != 'netstandard2.1'" Version="$(SystemMemoryPackageVersion)" />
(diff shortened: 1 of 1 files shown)


src/Grpc.Core.Api/Grpc.Core.Api.csproj
```diff
@@ -20,7 +20,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Memory" Version="$(SystemMemoryPackageVersion)" />
+    <PackageReference Include="System.Memory" Condition="'$(TargetFramework)' != 'netstandard2.1'" Version="$(SystemMemoryPackageVersion)" />
```
JamesNK, line 23:
```suggestion
    <PackageReference Include="System.Memory" Version="$(SystemMemoryPackageVersion)" />
```

src/Grpc.Core.Api/Grpc.Core.Api.csproj
```diff
@@ -20,7 +20,7 @@
   </ItemGroup>
 
   <ItemGroup>
```
JamesNK, line 22:
```suggestion
  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.1'">
```

JamesNK (approved):
Other projects have the condition on the ItemGroup. Change PR to be consistent.

JamesNK:
Thanks

PR #2196, opened 2023-07-18
Improve gRPC channel and client debugging

JamesNK:
## GrpcChannel

Before:

image

After:

image

## Generated clients

Before:

image

After:

image

image

### DIFF
--- a/src/Grpc.Core.Api/ClientBase.cs
+++ b/src/Grpc.Core.Api/ClientBase.cs
@@ -86,7 +86,8 @@ public abstract class ClientBase<T> : ClientBase
 /// <summary>
 /// Base class for client-side stubs.
 /// </summary>
-// The call invoker's debug information is provided by DebuggerDisplayAttribute. It isn't available in ToString.
+// The call invoker's debug information is specified in DebuggerDisplayAttribute.
+// It can't be concatenated inside ServiceNameDebuggerToString() because it isn't available in ToString.
 [DebuggerDisplay("{ServiceNameDebuggerToString(),nq}{CallInvoker}")]
 [DebuggerTypeProxy(typeof(ClientBaseDebugType))]
 public abstract class ClientBase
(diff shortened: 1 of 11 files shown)


JamesNK:
@mgravell

mgravell (approved):

src/Grpc.Core.Api/ClientBase.cs
```diff
@@ -84,6 +86,9 @@ public T WithHost(string host)
 /// <summary>
 /// Base class for client-side stubs.
 /// </summary>
+// The call invoker's debug information is provided by DebuggerDisplayAttribute. It isn't available in ToString.
```
JamesNK, line 89:
```suggestion
// The call invoker's debug information is specified in DebuggerDisplayAttribute.
// It can't be concatenated inside ServiceNameDebuggerToString() because it isn't available in ToString.
```

src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -845,6 +846,22 @@ public ISubchannelTransport Create(Subchannel subchannel)
...
 #endif
 
+    internal string DebuggerToString()
+    {
+        var debugText = $@"Address = ""{Address.OriginalString}""";
+        if (!IsHttpOrHttpsAddress(Address))
+        {
+            // It is easy to tell whether a channel is secured when the address contains http/https.
+            // Load balancing use custom schemes. Include IsSecure in debug text for custom schemes.
+            debugText += $", IsSecure = {(_isSecure ? "true" : "false")}";
```
JamesNK, line 856:
```suggestion
            // It is easy to tell whether a channel is secured when the address contains http/https.
            // Load balancing use custom schemes. Include IsSecure in debug text for custom schemes.
            // For example: Address = "dns:///my-dns-server, IsSecure = false
            debugText += $", IsSecure = {(_isSecure ? "true" : "false")}";
```

PR #2201, opened 2023-07-21
fix: concurrent execution restarts the timer

Wolffy101:


JamesNK:
CLA needs to be signed before this can be merged. If you're unable to sign then let me know, and I can apply this change.

Wolffy101:
> CLA needs to be signed before this can be merged. If you're unable to sign then let me know, and I can apply this change.

I can't sign,thanks!

PR #2204, opened 2023-07-21
Add a recursive parameter to the dotnet-grpc add-files CLI command

TheKroni:
Add a recursive parameter to the dotnet-grpc add-files CLI command. 

The new parameter will search the files in the requested folder recursively which enables adding proto references to projects inside multiple sub-folders. 

A unit test which verifies the functionality has been added and previous non recursive unit tests were changed.

### DIFF
--- a/src/dotnet-grpc/Commands/CommandBase.cs
+++ b/src/dotnet-grpc/Commands/CommandBase.cs
@@ -1,4 +1,4 @@
-﻿#region Copyright notice and license
+#region Copyright notice and license
 
 // Copyright 2019 The gRPC Authors
 //
@@ -174,7 +174,7 @@ internal class CommandBase
         }
 
         var normalizedFile = NormalizePath(file);
-        
+
         var normalizedAdditionalImportDirs = string.Empty;
 
         if (!string.IsNullOrWhiteSpace(additionalImportDirs))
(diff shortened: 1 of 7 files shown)


src/dotnet-grpc/Commands/CommandBase.cs
```diff
@@ -315,7 +315,7 @@ internal string[] GlobReferences(string[] references)
                 var directoryToSearch = Path.GetPathRoot(reference)!;
                 var searchPattern = reference.Substring(directoryToSearch.Length);
 
-                var resolvedFiles = Directory.GetFiles(directoryToSearch, searchPattern);
+                var resolvedFiles = Directory.GetFiles(directoryToSearch, searchPattern, searchOption);
```
Jecha1917, line 318: It's not a good idea to import all proto files, it's better to go through the proto imports recursively and import the specified files if they exist. It will look like this: we add only the `securities.proto` file, the definition of the service, the rest used will be pulled up recursively.
example:
```
syntax = "proto3";
import "proto/fftapi/v1/security.proto";
package grpc.fftapi.v1;

message GetSecuritiesRequest {
}

message GetSecuritiesResult {
  repeated proto.fftapi.v1.Security securities = 1;
}

service Securities {
   rpc GetSecurities(grpc.fftapi.v1.GetSecuritiesRequest) returns (grpc.fftapi.v1.GetSecuritiesResult);
}
```
This example is from a real service. The approach that implements recursive directory search not only fails to complete successfully but also generates unnecessary services and types.

xky0007:
When will this PR be merged? Thanks.

Issue #183, opened 2019-03-22
Enable referencing proto file in NuGet packages


alexvanboxel:
A common pattern in other languages is that your proto files are not part of your project, especially if you are a client but ofter also when you are a server.

Certainly in enterprises that have their stack build on gRPC. What you see is that the proto's are managed in a common mono repo for all contracts (were talking **mono-repo for the proto** files only here). What you then often see is that contract repo has a language specific distribution step. This do not create client libraries, they just _publish the proto files in a package format idiomatic to the language_ (an example is Java: they package the proto files in a jar package and publish them to a maven repo, in general a private one of the company).

What I like to see is that the .NET implementation would be able to reference **proto in NuGet** that it can fetch from a NuGet repository. That will make it easier for enterprises to adapt gRPC in .NET. They only need to add the NuGet distribution step in one place and a project just need to references NuGet as an **external** dependency.

Let me also stress that this is distributing the proto files and **not client libraries**. You want to _avoid_ distribution of clients that are build with an older version of proto as the one you are using in your project.

And example how this works in Java (using gradle):

`dependencies {
    compile("com.google.protobuf:protobuf-java:3.6.1")
    compile("io.grpc:grpc-protobuf:1.18.0")
    protobuf('io.anemos:protobeam-options-proto:0.0.3-SNAPSHOT')
}
`
The 3th dependency points to an artifact just containing proto, gradle will make sure that client/server stubs are generate with the same version of gRPC and protobuf as the project is.

JamesNK:
Your Java example, is it like a csproj PackageReference + Protobuf reference of the contents rolled into one?

JamesNK:
Today I think you could do this:

* Package a nupkg with the proto files as content
* Reference the NuGet package in your client/server, which would then include the content files in your project
* Add a `<Protobuf>` line to your csproj that references the proto files

JunTaoLuo:
Considerations:
- How to pass the options to the compiler, for example the option to use more efficient APIs for serialization.

antonioortizpola:
@JamesNK Could you provide and example? I am trying using my protos from other project in my solution (the idea is latter use nuget), but I can not get it to work.

The idea I think is simple, I have a protos project with the proto files in a folder called "Protos". The files are marked as Content.

![image](https://user-images.githubusercontent.com/11655127/55786407-c9852680-5a79-11e9-93a6-6ef5a6f0bd75.png)

Then I add that project as reference to mi server project, i can see the references of the files

[...]

PR #2205, opened 2023-07-21
First pass at Grpc.StatusProto API

tonydnewell:
Grpc.StatusProto API and NuGet package for implementing the gRPC richer error model.

It is a work in progress. This draft PR is for reviewing the design and to enable discussion. No examples yet included.

@jtattermusch @JamesNK @jskeet


tonydnewell:
> Is the plan still for a lot of this to be in Google.Api.CommonProtos? If so, it's probably worth indicating which bits would be there in this PR.

The parts of this that could be moved into `Google.Api.CommonProtos` are:
- the extension methods in `RpcStatusExtensions.cs` - all except `ToRpcException` as you don't want a dependency on `Grpc.Core.Api`
- the types registry in `DetailsTypesRegistry`

but as I don't have ownership of `Google.Api.CommonProtos` it depends on what makes sense to you.

jskeet:
That split sounds reasonable to me. Will review when I get a chance.

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
+﻿# Grpc C# API for error handling with Status.proto
+
+This is a protoype NuGet package providing client and server side support for the
+[gRPC richer error model](https://grpc.io/docs/guides/error/#richer-error-model).
+
+It had dependencies NuGet packages on:
+* `Google.Api.CommonProtos` - to provide the proto implementations used by the richer error model
```
jtattermusch, line 7: as discussed offline, if some of the C# gets added to Google.Api.CommonProtos as extension methods, let's pay close attention to in which namespace these extension methods live, since we don't want to require adding too many "using x.y.z" statements to be able to use the Grpc.StatusProto functionality (in order for the extension methods to become visible).

src/Grpc.StatusProto/MetadataExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// Get the Google.Rpc.Status from the metadata.
+    /// </summary>
+    /// <param name="metadata"></param>
+    /// <returns>
+    /// The found Google.Rpc.Status or null if it was not present or could
+    /// the data could not be parsed.
+    /// </returns>
+    public static Google.Rpc.Status? GetRpcStatus(this Metadata metadata)
+    {
+        Metadata.Entry? entry = metadata.FirstOrDefault(t => t.Key == StatusDetailsTrailerName);
```
jtattermusch, line 40: the choice of "FirstOrDefault" vs some other method needs to work well with behavior of SetRpcStatus in the sense of my comment.

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+* `Grpc.Core.Api` - for API classes such as `RpcExcetion`
```
jtattermusch, line 8: typo `RpcExcetion`

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+
+Reviewers notes:
+
+* The client side borrows from ideas used in [googleapi/gax-dotnet](https://github.com/googleapis/gax-dotnet), specifically [RpcExceptionExtensions.cs](https://github.com/googleapis/gax-dotnet/blob/main/Google.Api.Gax.Grpc/RpcExceptionExtensions.cs)
```
jtattermusch, line 12: so the plan is that RpcExceptionExtensions from this PR will be added to Google.Api.CommonProtos and after those changes are released, Google.Api.Gax.Grpc will migrate to RpcExceptionExtensions from Google.Api.CommonProtos?

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+doing this. In this error model the error is encapsulated by an instance of `Google.Rpc.Status` and
+returned in the trailing response metadata.  Setting and reading this metadata is handled for you. Details below.
+
+## Server Side
+
+The server side uses C#'s Object and Collection initializer syntax. The avoids the needs to a *builder* API to be developed.
+
+There are two ways that the server can return an error status:
+- by throwing an `RpcException` that contains the details of the status.
+- by explicitly set the `Status` and `ResponseTrailers` in the `ServerCallContext`
```
jtattermusch, line 36: nit: "explicitly set the `Status` and `ResponseTrailers`" doesn't read well.

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+The `Google.Rpc.Status` extension method `ToRpcException` creates the appropriate `RpcException` from the status.
+
+The `Grpc.Core.Metadata` extension method `SetRpcStatus` adds a binary representation of the status to the metadata.
+
+
+Example - throwing a RcpException:
+```C#
+throw new Google.Rpc.Status
+{
+    Code = (int)StatusCode.NotFound,
```
jtattermusch, line 54: needing to assign  Code = with a type cast feels odd. 

~~Idea: Since code and message can be left empty, how about having an overload `.ToRpcException(statusCode, message)` that sets the status code and message for the RpcException in a strongly typed fashion? I'm open to other ideas.~~

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+        })
+    }
+});
+
+return Task.FromResult( /* ... */ );
+```
+
+## Client Side
+
+There is an extension method to retrieve a `Google.Rpc.Status` from the metadata in an `RpcException`.  There are also extension methods to retieve the details from the `Google.Rpc.Status`.
```
jtattermusch, line 93: typo: retieve

jtattermusch, line 93: "There are also extension methods to retieve the details from the `Google.Rpc.Status`" reads as "retrieve Google.Rpc.Status from Google.Rpc.Status" which makes no sense?

src/Grpc.StatusProto/README.md
jtattermusch: on that note, what are the requirements for  Core and Message fields being set/unset and do they have to always match the Rpc's status code and message? What happens when they are different?

I think things can get confusing since we essentially have 2 sources of truth for the rpc's status code and messages so we need some clarity here.

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+        {
+            RequestId = "EchoRequest",
+            ServingData = "Param: " + request.Action.ToString()
+        }),
+        }
+}.ToRpcException();
+```
+Example - setting the status and response trailers instead of throwing an exception:
+```C#
+context.Status = new Grpc.Core.Status(StatusCode.Internal, "Some detail");
```
jtattermusch, line 73: in this example you are setting the Grpc.Core.Status of an RPC to StatusCode.Internal, but the Google.Rpc.Status to NotFound (along with a different message as well) which feels odd and it seems that this should be discouraged? Some clarity is needed on this topic (also see comments above).

Is Google.Rpc.Status.Code from a completely different domain than Grpc.Core.Status.StatusCode or are they supposed to be the same thing?

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+});
+
+return Task.FromResult( /* ... */ );
+```
+
+## Client Side
+
+There is an extension method to retrieve a `Google.Rpc.Status` from the metadata in an `RpcException`.  There are also extension methods to retieve the details from the `Google.Rpc.Status`.
+
+If the client knows what details to expect for a specific error code then it can use the extension methods the explicitly extract the know type from the status details.
```
jtattermusch, line 95: typo: "extract the know type"

src/Grpc.StatusProto/README.md
jtattermusch: https://cloud.google.com/apis/design/errors#error_codes brings some more clarity to this topic, but this README.md doesn't mention the rules/recommendations for Google.Rpc.Status.Code vs Grpc.Core.Status.StatusCode at all.

My understanding is this:
- Google.Rpc.Status.Code is fully independent from Grpc.Core.Status.StatusCode and can be set to a different code and it's fine to do so.
- the recommendation is to use the values of the enum Grpc.Core.Status.StatusCode as values for Google.Rpc.Status.Code as a convention. This is a must for Google APIs and strongly recommended for third party services. But users can use a different domain of values if they want and and as long as their services are mutually compatible, things will work fine.

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+                    break;
+
+                // Other cases handled here ...
+            }
+        }
+    }
+
+```
+
+## Returning errors within gRPC streams
```
jtattermusch, line 172: I didn't know about the streaming use case with per-message error details. Where did you find that?

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+        // or error details
+        google.rpc.Status status = 2;
+   }   
+}
+```
+
+Note: the  [status.proto](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto)
+and [error_details.proto](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto)
+files are not currently provided in any NuGet packages. The text for those proto files can be copied from
+the given links and included in your project.
```
jtattermusch, line 203: nit: reword "The text for those proto files can be copied from
the given links and included in your project."

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+// sending the requests
+foreach (var request in requests)
+{
+    await call.RequestStream.WriteAsync(request);
+}
+```
+
+
+## See also
+* [Richer error model](https://grpc.io/docs/guides/error/#richer-error-model)
```
jtattermusch, line 255: perhaps you can link directly to https://cloud.google.com/apis/design/errors#error_model somewhere in the README since that documentation has quite a lot of useful info.

src/Grpc.StatusProto/README.md
jskeet: Yes for the first part. For the second part, we'd probably migrate the implementation immediately (making the existing extension methods call the new ones) but we can't remove them from the code itself until the next major version of GAX.

src/Grpc.StatusProto/DetailsTypesRegistry.cs
```diff
@@ -0,0 +1,63 @@
...
+            Help.Descriptor,
+            LocalizedMessage.Descriptor
+        }
+        );
+
+    /// <summary>
+    /// Get the registry
+    /// </summary>
+    /// <returns>the registry</returns>
+    public static TypeRegistry GetRegistry() { return registry; }
```
jskeet, line 49: Nit: I'd use an expression-bodied member for this and the Unpack method.

Any reason for this being a method rather than a property?

src/Grpc.StatusProto/DetailsTypesRegistry.cs
```diff
@@ -0,0 +1,63 @@
...
+using Google.Rpc;
+
+namespace Grpc.StatusProto;
+
+/// <summary>
+/// Registry of all the expected types that can be in the "details" of a Status.
+/// See https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto
+/// for a list of expected messages.
+/// </summary>
+public static class DetailsTypesRegistry
```
jskeet, line 27: Does this need to be public at all?

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+    {
+        Any.Pack(new ErrorInfo
+        {
+            Domain = "error example",
+            Reason = "some reason"
+        }),
+        Any.Pack(new RequestInfo
+        {
+            RequestId = "EchoRequest",
+            ServingData = "Param: " + request.Action.ToString()
```
jskeet, line 66: Nit: `ServingData = $"Param: {request.Action}"`

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+Example - setting the status and response trailers instead of throwing an exception:
+```C#
+context.Status = new Grpc.Core.Status(StatusCode.Internal, "Some detail");
+context.ResponseTrailers.SetRpcStatus(new Google.Rpc.Status
+{
+    Code = (int)StatusCode.NotFound,
+    Message = "Simple error message",
+    Details =
+    {
+        Any.Pack(new ErrorInfo
```
jskeet, line 80: FYI: There are Google-internal proposals around the use of ErrorInfo; I'm checking on the status of those, but they could affect the preferred layout for the richer error model.

src/Grpc.StatusProto/DetailsTypesRegistry.cs
jskeet: Or rather, as an alternative, might it be worth adding a public static property to Google.Rpc.Status? Or maybe even just a method of UnpackDetailMessage, to hide how the type registry stuff works?

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+```C#
+void PrintStatusDetails(RpcException ex)
+{
+    // Get the status from the RpcException
+    Google.Rpc.Status? rpcStatus = ex.GetRpcStatus(); // Extension method
+
+    if (rpcStatus != null)
+    {
+        // Decode each "Any" item from the details in turn
+        foreach (Any any in rpcStatus.Details)
```
jskeet, line 136: Another thought on the API surface... how about exposing an `rpcStatus.UnpackedDetails` which is an `IEnumerable<IMessage>`?

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,256 @@
...
+        {
+            // decode the message if it is one of those expected to be in the
+            // status details
+            IMessage msg = DetailsTypesRegistry.Unpack(any); 
+
+            switch (msg)
+            {
+                case null:
```
jskeet, line 144: This would go away in UnpackedDetails - we'd just ignore any unknown error details types. That's probably more convenient for users (who can always do it the long way to get all the info).

src/Grpc.StatusProto/README.md
jskeet: This is interesting... do you think it would be beneficial to ship them in Google.Api.CommonProtos? I wouldn't be averse to that, but I'd need to consult the rest of the team.

test/Grpc.StatusProto.Tests/RpcStatusExtensionsTest.cs
```diff
@@ -0,0 +1,267 @@
...
+
+namespace Grpc.StatusProto.Tests;
+
+/// <summary>
+/// Tests for RpcStatusExtensions
+/// </summary>
+public class RpcStatusExtensionsTest
+{
+    [Test]
+    public void ToRpcExcetionTest()
```
jskeet, line 35: Typo in method name.

src/Grpc.StatusProto/RpcStatusExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+        var expectedName = new T().Descriptor.FullName;
+        var any = status.Details.FirstOrDefault(a => Any.GetTypeName(a.TypeUrl) == expectedName);
+        if (any is null)
+        {
+            return null;
+        }
+        return any.Unpack<T>();
+    }
+
+    /// <summary>
```
jskeet, line 47: I know methods like this are present in GAX at the moment - but I'm not sure whether they're really worthwhile. It's easy enough to just call `GetStatusDetail<BadRequest>`.

I'd suggest sticking to just the generic method to start with (including some examples) and then adding more extension methods later if we need them.

jskeet (commented):
I've focused on the API surface rather than the implementation and tests. I suspect we'll have more discussions.

tonydnewell:
@jtattermusch @jskeet @JamesNK 
I've updated the API and documentation from the review comments.

There is now less code! In summary:
- removed a lot of the methods that were just wrappers for `GetStatusDetail<T>`
- added `UnpackedDetails` to enumerate over the messages in the details
- renamed `DetailsTypesRegistry` to `StandardErrorTypeRegistry` - and it is now not used directly by applications
- updated tests
- add more doc comments
- reworked the README.md documentation

test/Grpc.StatusProto.Tests/Grpc.StatusProto.Tests.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.1;net5.0;net472;net6.0;net7.0</TargetFrameworks>
```
jtattermusch, line 3: you might need to update this before merging (we've upgraded to net8 in the meantime).

src/Grpc.StatusProto/RpcStatusExtensions.cs
jtattermusch: Looks like this was resolved?

src/Grpc.StatusProto/README.md
jtattermusch: At this point, I'd say this wouldn't help much. I'd probably help if we had a general mechanism for shipping .proto file in nuget packages and Grpc.Tools could autogenerate source code for them (see https://github.com/grpc/grpc-dotnet/issues/183), but we currently don't have that.

test/Grpc.StatusProto.Tests/MetadataExtensionsTest.cs
```diff
@@ -0,0 +1,127 @@
+#region Copyright notice and license
+// Copyright 2015 gRPC authors.
```
jtattermusch, line 2: Old copyright notice, please update (here and elsewhere).

jtattermusch (commented):
Added a bunch of comments, but they are mostly minor.
Overall this is looking good so we are getting closer to being able to merge this soon (ish).

- I'd like @JamesNK  to review
- Please add a note like [this one](https://github.com/grpc/grpc-dotnet/blob/b421751176ece4ea8fb969bbfc4e8a64be874a49/src/Grpc.Core.Api/Status.cs#L50) to all of the newly added public APIs, since the first version of Grpc.StatusProto should be an experimental prototype. We can remove the note in the future once we feel more confident that the proposed API is usable and stable.

jtattermusch:
@JamesNK can you please review?

tonydnewell:
Updated from review comments.

I've also added extension methods `ToRcpDebugInfo` for `System.Exception` to populate a `DebugInfo` message from an exception (added message and stack trace).  Not sure about the method name but didn't want to just be `ToDebugInfo` as 
that seems a bit too generic.

I still need to update the project files to add net8.0 but as I branched before the net8.0 change I didn't want to break anything.

The README.md is getting quite long but there isn't really anywhere else to add documentation at the moment. I'm trying to keep a balance between brevity and enough detail to be useful without giving a full tutorial.

jskeet:
I'm not quite sure what the expectation is for me personally at this point. I suggest that I wait for @JamesNK and @jtattermusch to be happy, then I'll re-review - paying particular attention to the code that would be in Google.Api.CommonProtos.

jtattermusch:
> I'm not quite sure what the expectation is for me personally at this point. I suggest that I wait for @JamesNK and @jtattermusch to be happy, then I'll re-review - paying particular attention to the code that would be in Google.Api.CommonProtos.

That seems fair. I think I'm mostly happy (I'll give this one more final pass once we believe the PR is in its final form). We're waiting for @JamesNK's review currently, since he haven't provided any comments yet.

Also, can you remind me what exactly is the plan for "code that should be in Google.Api.CommonProtos"? Is the code going be merged to as part of this PR first (and eventually migrated to  Google.Api.CommonProtos) or are we planning to merge this PR once that code is in Google.Api.CommonProtos?

jtattermusch:
@JamesNK  friendly ping on the review.

test/Grpc.StatusProto.Tests/Grpc.StatusProto.Tests.csproj
JamesNK:
```suggestion
    <TargetFrameworks>net462;net6.0;net7.0;net8.0</TargetFrameworks>
```

src/Grpc.StatusProto/ExceptionExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+/// </code>
+/// </example>
+/// </remarks>
+/// <param name="exception"></param>
+/// <param name="innerDepth">Maximum number of inner exceptions to include in the StackTrace. Defaults
+/// to not including any inner exceptions</param>
+/// <returns>
+/// A new <see cref="Google.Rpc.DebugInfo"/> populated from the exception.
+/// </returns>
+public static DebugInfo ToRpcDebugInfo(this Exception exception, int innerDepth = 0)
```
JamesNK, line 55: indent

src/Grpc.StatusProto/MetadataExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+
+using Google.Protobuf;
+using Grpc.Core;
+
+namespace Grpc.StatusProto;
+
+/// <summary>
+/// Extension methods for the Grpc.Core.Metadata
+/// </summary>
+public static class MetadataExtensions
```
JamesNK, line 23:
```suggestion
public static class MetadataExtensions
```

src/Grpc.StatusProto/MetadataExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// Note: experimental API that can change or be removed without any prior notice.
+    /// </summary>
+    /// <param name="metadata"></param>
+    /// <param name="throwOnParseError">if true then <see cref="Google.Protobuf.InvalidProtocolBufferException"/>
+    /// is thrown if the metadata cannot be parsed. Otherwise null is returned on a parsing error.</param>
+    /// <returns>
+    /// The found <see cref="Google.Rpc.Status"/> or null if it was
+    /// not present or could the data could not be parsed.
+    /// </returns>
+    public static Google.Rpc.Status? GetRpcStatus(this Metadata metadata, bool throwOnParseError=false)
```
JamesNK, line 44:
```suggestion
    public static Google.Rpc.Status? GetRpcStatus(this Metadata metadata, bool throwOnParseError = false)
```

JamesNK: Why suppress the parse error by default? That seems like it could lead to errors disappearing.

src/Grpc.StatusProto/MetadataExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+        try
+        {
+            return Google.Rpc.Status.Parser.ParseFrom(entry.ValueBytes);
+        }
+        catch
+        {
+            if (throwOnParseError)
+            {
+                throw;
+            }
```
JamesNK, line 60:
```suggestion
        catch when (!throwOnParseError)
        {
```

src/Grpc.StatusProto/MetadataExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// <param name="metadata"></param>
+    /// <param name="throwOnParseError">if true then <see cref="Google.Protobuf.InvalidProtocolBufferException"/>
+    /// is thrown if the metadata cannot be parsed. Otherwise null is returned on a parsing error.</param>
+    /// <returns>
+    /// The found <see cref="Google.Rpc.Status"/> or null if it was
+    /// not present or could the data could not be parsed.
+    /// </returns>
+    public static Google.Rpc.Status? GetRpcStatus(this Metadata metadata, bool throwOnParseError=false)
+    {
+        var entry = metadata.LastOrDefault(t => t.Key == StatusDetailsTrailerName);
```
JamesNK, line 46: Add a null check for metadata.

src/Grpc.StatusProto/MetadataExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// <summary>
+    /// Add <see cref="Google.Rpc.Status"/> to the metadata.
+    /// Any existing status in the metadata will be overwritten.
+    /// Note: experimental API that can change or be removed without any prior notice.
+    /// </summary>
+    /// <param name="metadata"></param>
+    /// <param name="status">Status to add</param>
+    public static void SetRpcStatus(this Metadata metadata, Google.Rpc.Status status)
+    {
+        var entry = metadata.Get(StatusDetailsTrailerName);
```
JamesNK, line 76: Add null checks for metadata and status.

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,319 @@
+# Grpc C# API for error handling with status.proto
```
JamesNK, line 1:
```suggestion
# gRPC C# API for error handling with status.proto
```

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,319 @@
...
+To add messages to the `Details` repeated field in `Google.Rpc.Status`, wrap each one in `Any.Pack()` - see example below.
+
+The `Google.Rpc.Status` extension method `ToRpcException` creates the appropriate `RpcException` from the status.
+
+The `Grpc.Core.Metadata` extension method `SetRpcStatus` adds a binary representation of the status to the metadata.
+
+
+__Example__ - throwing a `RpcException`:
+```C#
+throw new Google.Rpc.Status
```
JamesNK, line 67: Put the sample in a gRPC `SayHello` method. Makes it a bit more real world.

src/Grpc.StatusProto/RpcStatusExtensions.cs
```diff
@@ -0,0 +1,210 @@
...
+    ///   }
+    /// </code>
+    /// </example>
+    /// </remarks>
+    /// <typeparam name="T">The message type to decode from within the error details.</typeparam>
+    /// <param name="status">The RPC status to retrieve details from. Must not be null.</param>
+    /// <returns>The first error details of type <typeparamref name="T"/> found, or null if not present</returns>
+    public static T? GetStatusDetail<T>(this Google.Rpc.Status status) where T : class, IMessage<T>, new()
+    {
+        var expectedName = new T().Descriptor.FullName;
```
JamesNK, line 50: This is a bit of a hack. I've gotten a type's descriptors without creating an instance before. I'll see if I can find it.

test/Grpc.StatusProto.Tests/ExceptionExtensionsTest.cs
```diff
@@ -0,0 +1,101 @@
...
+using NUnit.Framework;
+
+namespace Grpc.StatusProto.Tests;
+
+/// <summary>
+/// Tests for ExceptionExtensions
+/// </summary>
+public class ExceptionExtensionsTest
+{
+
```
JamesNK, line 28:
```suggestion
```

src/Grpc.StatusProto/README.md
JamesNK: I don't think we should encourage this pattern. You _can_ set `context.Status` but you still have to return a value from the method (assuming it is a unary or client streaming method).

It's simpler to consistently tell people that if you want to end the gRPC call with a non-OK status you should throw an exception.

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,319 @@
...
+            RequestId = "EchoRequest",
+            ServingData = "Param: " + request.Action.ToString()
+        }),
+        }
+}.ToRpcException();
+```
+__Example__ - setting the status and response trailers instead of throwing an exception:
+```C#
+context.Status = new Grpc.Core.Status(StatusCode.Internal, "Some detail");
+context.ResponseTrailers.SetRpcStatus(new Google.Rpc.Status
```
JamesNK, line 89: Is there a scenario where a method successfully completes and you would want to set a status in the response trailers?

src/Grpc.StatusProto/RpcStatusExtensions.cs
```diff
@@ -0,0 +1,210 @@
...
+    /// }.ToRpcException();
+    /// </code>
+    /// </example>
+    /// </para>
+    /// </remarks>
+    /// <param name="status">The RPC status. Must not be null</param>
+    /// <returns>A <see cref="RpcException"/> populated with the details from the status.</returns>
+    public static RpcException ToRpcException(this Google.Rpc.Status status)
+    {
+        return status.ToRpcException((StatusCode)status.Code, status.Message);
```
JamesNK, line 88: Is this cast safe? If it is, add a comment saying why.

src/Grpc.StatusProto/StandardErrorTypeRegistry.cs
```diff
@@ -0,0 +1,52 @@
...
+            RetryInfo.Descriptor,
+            DebugInfo.Descriptor,
+            QuotaFailure.Descriptor,
+            PreconditionFailure.Descriptor,
+            RequestInfo.Descriptor,
+            ResourceInfo.Descriptor,
+            Help.Descriptor,
+            LocalizedMessage.Descriptor
+        }
+        );
```
JamesNK, line 42: nit
```suggestion
        });
```

src/Grpc.StatusProto/StandardErrorTypeRegistry.cs
```diff
@@ -0,0 +1,52 @@
...
+
+/// <summary>
+/// Registry of the <see href="https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto">
+/// standard set of error types</see> defined in the richer error model developed and used by Google.
+/// These can be sepcified in the <see cref="Google.Rpc.Status.Details"/>.
+/// Note: experimental API that can change or be removed without any prior notice.
+/// </summary>
+public static class StandardErrorTypeRegistry
+{
+    private static readonly TypeRegistry registry = TypeRegistry.FromMessages(
```
JamesNK, line 28: _registry

(the older code from grpc/grpc doesn't use underscores for private fields, but new grpc-dotnet code does)

src/Grpc.StatusProto/StandardErrorTypeRegistry.cs
```diff
@@ -0,0 +1,52 @@
...
+        );
+
+    /// <summary>
+    /// Get the registry
+    /// Note: experimental API that can change or be removed without any prior notice.
+    /// </summary>
+    public static TypeRegistry Registry
+    {
+        get => registry;
+    }
```
JamesNK, line 51:
```suggestion
    public static TypeRegistry Registry => _registry;
```

src/Grpc.StatusProto/RpcStatusExtensions.cs
```diff
@@ -0,0 +1,210 @@
...
+    ///   }
+    /// }
+    /// </code>
+    /// </example>
+    /// </para>
+    /// </remarks>
+    /// <param name="status"></param>
+    /// <param name="registry"></param>
+    /// <returns></returns>
+    public static IEnumerable<IMessage> UnpackDetailMessage(this Google.Rpc.Status status, TypeRegistry registry)
```
JamesNK, line 199: UnpackDetailMessage -> UnpackDetailMessages

src/Grpc.StatusProto/RpcStatusExtensions.cs
```diff
@@ -0,0 +1,210 @@
...
+    ///     // Other cases ...
+    ///   }
+    /// }
+    /// </code>
+    /// </example>
+    /// </para>
+    /// </remarks>
+    /// <param name="status"></param>
+    /// <returns></returns>
+    public static IEnumerable<IMessage> UnpackDetailMessage(this Google.Rpc.Status status)
```
JamesNK, line 158: UnpackDetailMessage -> UnpackDetailMessages

src/Grpc.StatusProto/MetadataExtensions.cs
JamesNK, line 46: `metadata.Get(...)` is faster than LINQ.

src/Grpc.StatusProto/ExceptionExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Rpc;
+
+namespace Grpc.StatusProto;
```
JamesNK, line 17: I think the namespace of the extension types should change. See https://github.com/grpc/grpc-dotnet/pull/2273#issuecomment-1707865018

src/Grpc.StatusProto/RpcExceptionExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// Retrieves the <see cref="Google.Rpc.Status"/> message containing extended error information
+    /// from the trailers in an <see cref="RpcException"/>, if present.
+    /// Note: experimental API that can change or be removed without any prior notice.
+    /// </summary>
+    /// <param name="ex">The RPC exception to retrieve details from. Must not be null.</param>
+    /// <returns>The <see cref="Google.Rpc.Status"/> message specified in the exception, or null
+    /// if there is no such information.</returns>
+    public static Google.Rpc.Status? GetRpcStatus(this RpcException ex)
+    {
+        return ex.Trailers?.GetRpcStatus();
```
JamesNK, line 34: Add a null check for RpcException

JamesNK, line 34: Trailers isn't nullable
```suggestion
        return ex.Trailers.GetRpcStatus();
```

src/Grpc.StatusProto/RpcStatusExtensions.cs
JamesNK: It was over service descriptors and I used reflection. I think I prefer this.

What about adding caching with a static `ConcurrentDictionary<Type, string>`? That reduces the allocation to the first time this method is used with each type.

src/Grpc.StatusProto/README.md
tonydnewell: I'll remove the documentation for the non-standard pattern.

src/Grpc.StatusProto/MetadataExtensions.cs
tonydnewell: Copying the behaviour from Google.Api.Gax.Grpc
https://github.com/googleapis/gax-dotnet/blob/897b4bdf84d5b314c424113a8bb633bb703ff1a3/Google.Api.Gax.Grpc/RpcExceptionExtensions.cs#L73-L81

tonydnewell:
@JamesNK @jtattermusch 
Updated from James' review comments.

Still to do:
- change namespace - does anyone else have an opinion on using Grpc.Core instead of Grpc.StatusProto namespace?
- adding caching in RpcStatusExtensions.cs - I've added a TODO and will look at that

@jskeet The methods that could be moved to Google.Api.CommonProtos are just three methods in RcpStatusExtensions.cs and I've marked those with TODO.

tonydnewell:
Added cache to RpcStatusExtensions

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,305 @@
...
+    } else
+    {
+        response.WidgetDetails = "the details";
+    }
+}
+```
+
+Example client code fragment:
+```C#
+
```
JamesNK, line 276:
```suggestion
```C#
```

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,305 @@
...
+// reading the responses
+var responseReaderTask = Task.Run(async () =>
+{
+    while (await call.ResponseStream.MoveNext())
```
JamesNK, line 280: Change this to use an await foreach + ReadAllAsync. It's a bit cleaner.

Example:

https://github.com/grpc/grpc-dotnet/blob/246c43d4574a5e2ba3fb4f397df304dfb5688b2d/examples/Vigor/Client/Program.cs#L35-L38

test/Grpc.StatusProto.Tests/ExceptionExtensionsTest.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Text;
+using Google.Rpc;
+using NUnit.Framework;
+
+namespace Grpc.StatusProto.Tests;
+
+/// <summary>
+/// Tests for ExceptionExtensions
+/// </summary>
+public class ExceptionExtensionsTest
```
JamesNK, line 26: Add `[TestFixture]` to test classes.

test/Grpc.StatusProto.Tests/ExceptionExtensionsTest.cs
```diff
@@ -0,0 +1,100 @@
...
+{
+    [Test]
+    public void ToRpcDebugInfoTest()
+    {
+        try
+        {
+            ThrowException("extra details");
```
JamesNK, line 33: The other new tests in grpc-dotnet are divided into Arrange/Act/Assert. For example:

https://github.com/grpc/grpc-dotnet/blob/246c43d4574a5e2ba3fb4f397df304dfb5688b2d/test/Grpc.Net.Client.Tests/GrpcChannelTests.cs#L126-L148

Could you add comments to the tests to divided them into sections? It's ok if a test isn't perfect and asserts are intermingled with other test code. Just do your best.

JamesNK, line 33: You can use `var ex = Assert.Throws<Exception>(() => ThrowException("extra details"));` instead of try/catch. It's a bit more idiomatic way of writing tests that deal with exceptions.

For example:

https://github.com/grpc/grpc-dotnet/blob/246c43d4574a5e2ba3fb4f397df304dfb5688b2d/test/Grpc.AspNetCore.Server.Tests/ServerCallContextExtensionsTests.cs#L44

test/Grpc.StatusProto.Tests/MetadataExtensionsTest.cs
```diff
@@ -0,0 +1,139 @@
...
+
+namespace Grpc.StatusProto.Tests;
+
+/// <summary>
+/// Tests for MetadataExtensions
+/// </summary>
+public class MetadataExtensionsTest
+{
+    // creates a status to use in the tests
+    readonly Google.Rpc.Status status = new()
```
JamesNK, line 31:
```suggestion
    private readonly Google.Rpc.Status status = new()
```

test/Grpc.StatusProto.Tests/RpcExceptionExtensionsTest.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Grpc.StatusProto.Tests;
+
+/// <summary>
+/// Tests for RpcExceptionExtensions
+/// </summary>
+public class RpcExceptionExtensionsTest
+{
+    // creates a status to use in the tests
+    readonly Google.Rpc.Status status = new()
```
JamesNK, line 30:
```suggestion
    private readonly Google.Rpc.Status status = new()
```

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+Note: the  [status.proto](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto)
+and [error_details.proto](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto)
+files are not currently provided in any NuGet packages. If you wish to use these in
+your own message definitions then you will need to copy these into your own project.
+
+Example server code fragment:
+```C#
+ while (await requestStream.MoveNext())
```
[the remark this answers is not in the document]
JamesNK: Actually, ignore that. Use `ReadAllAsync` to read the stream. Example is in another comment.

src/Grpc.StatusProto/README.md
jskeet: As it happens, we had a feature request for that last week: https://github.com/googleapis/gax-dotnet/issues/715
@jtattermusch If you believe it wouldn't actually help much, please could you comment on that issue?

src/Grpc.StatusProto/MetadataExtensions.cs
jskeet: Yes, I know this code. It feels like if we're already in an error situation (which is where this code would usually be used), failing in a different way due to parse failures in metadata is unhelpful. It's a bit like ending up with an exception caused by disposing of a socket, when the original exception is trying to *write* to the socket, if you see what I mean.

Happy to discuss alternatives though.

src/Grpc.StatusProto/RpcStatusExtensions.cs
```diff
@@ -0,0 +1,263 @@
...
+    ///   }
+    /// </code>
+    /// </example>
+    /// </remarks>
+    /// <typeparam name="T">The message type to decode from within the error details.</typeparam>
+    /// <param name="status">The RPC status to retrieve details from. Must not be null.</param>
+    /// <returns>The first error details of type <typeparamref name="T"/> found, or null if not present</returns>
+    public static T? GetStatusDetail<T>(this Google.Rpc.Status status) where T : class, IMessage<T>, new()
+    {
+        //TODO(tonydnewell) maybe move this method to Google.Api.CommonProtos
```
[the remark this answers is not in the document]
jskeet: If we move it, the namespace would be Google.Api.

The conflict is definitely a potential cause of annoyance - but it'll be *reasonably* temporary as I'll remove it from GAX in our next major version. If we can think of an equally-good name, I'd be fine to have the two different names... but if `GetStatusDetail` is really the definitely-best name, I'd rather take the short-term ambiguity hit (and probably write a paragraph about it in our client library help) than lose the best name.

src/Grpc.StatusProto/RpcStatusExtensions.cs
jskeet: Yup. Don't think I can resolve it though.

src/Grpc.StatusProto/RpcStatusExtensions.cs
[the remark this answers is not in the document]
jskeet: Ah - if it's on Google.Rpc.Status, rather than the current RpcException, then just `GetDetail` might be better - it's more concise and it avoids the conflict. Thoughts?

src/Grpc.StatusProto/README.md
tonydnewell: It is on my list to document a standard way/best practices for shipping proto files in NuGet packages so that Grpc.Tools can easily pick them up. Basically what @JamesNK does in the [Microsoft.AspNetCore.Grpc.JsonTranscoding](https://www.nuget.org/packages/Microsoft.AspNetCore.Grpc.JsonTranscoding) package.

src/Grpc.StatusProto/MetadataExtensions.cs
tonydnewell: I think that not losing the original error (because reporting the error itself has problems) is my preference. But no strong views.

src/Grpc.StatusProto/RpcStatusExtensions.cs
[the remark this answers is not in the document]
tonydnewell: I'll change to GetDetail as it is already on the Status.  I've not changed the namespace unless you think we should start using the Google.Rpc namespace now before any methods are moved?  Also I'll not be too upset if the methods are never moved.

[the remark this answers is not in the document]
jskeet: I guess if you change the namespace to reflect where the methods will be, then all the calling code will already be "right". And I do want to put the extension methods in Google.Api.CommonProtos. Happy to create a PR for that myself, or you can.

tonydnewell:
@jskeet @JamesNK @jtattermusch 
I think I've addressed all the code review comments with the exception of the namespaces to use.  I've left the namespace as `Grpc.StatusProto` as I don't know what is the right thing to do:
- maybe use `Grpc.Core` to avoid adding more `using` statements? Or having an explicit different namespace shows intention?
- for the methods that might be moved to Google.Api.CommonProtos use `Google.Rpc`?

Also `StandardErrorTypeRegistry.cs` should/could also be moved to Google.Api.CommonProtos, or it could completely disappear as a separate class as it only defines a constant registry that is only used in `RpcStatusExtensions.UnpackDetailMessages()`

jskeet:
> For the methods that might be moved to Google.Api.CommonProtos use Google.Rpc?

Yes - and one thing I hadn't even thought to mention before is that we won't need to use extension methods there. It can just be a partial class.

tonydnewell:
> > For the methods that might be moved to Google.Api.CommonProtos use Google.Rpc?
> 
> Yes - and one thing I hadn't even thought to mention before is that we won't need to use extension methods there. It can just be a partial class.

Should I/can I do that here too to make the moving easier?

tonydnewell:
@jtattermusch @JamesNK Should I create a "proposal" for this feature? I've noticed there is one for python ([L44-python-rich-status.md](https://github.com/grpc/proposal/blob/master/L44-python-rich-status.md))

tonydnewell:
> > > For the methods that might be moved to Google.Api.CommonProtos use Google.Rpc?
> > 
> > 
> > Yes - and one thing I hadn't even thought to mention before is that we won't need to use extension methods there. It can just be a partial class.
> 
> Should I/can I do that here too to make the moving easier?

To answer my own question, looks like partial classes can only be used in the same project/assembly. So no here.

jtattermusch:
> @jskeet @JamesNK @jtattermusch I think I've addressed all the code review comments with the exception of the namespaces to use. I've left the namespace as `Grpc.StatusProto` as I don't know what is the right thing to do:
> 
> * maybe use `Grpc.Core` to avoid adding more `using` statements? Or having an explicit different namespace shows intention?
> * for the methods that might be moved to Google.Api.CommonProtos use `Google.Rpc`?
> 
> Also `StandardErrorTypeRegistry.cs` should/could also be moved to Google.Api.CommonProtos, or it could completely disappear as a separate class as it only defines a constant registry that is only used in `RpcStatusExtensions.UnpackDetailMessages()`

I'll leave this to @apolcyn, @JamesNK and @jskeet  to decide.

Note that since all the APIs are marked as experimental, we could start with the namespace being `Grpc.StatusProto` and move things to `Grpc.Core` namespace (or elsewhere) in the next few months after 1-2 releases, once we collect some initial feedback from real life users (while not risking to break stuff in the Grpc.Core namespace in the very first release of the new nuget).

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,303 @@
...
+The server side uses C#'s Object and Collection initializer syntax.
+
+The server returns the additional error information by throwing an `RpcException` that is
+created from a `Google.Rpc.Status` which contains the details of the error.
+
+To add messages to the `Details` repeated field in `Google.Rpc.Status`, wrap each one in `Any.Pack()` - see example below.
+
+The `Google.Rpc.Status` extension method `ToRpcException` creates the appropriate `RpcException` from the status.
+
+__Example__ - creating and throwing a `RpcException`:
```
codymullins, line 56: Will there be a way to use this without requiring the implementation to throw an exception?

tonydnewell: You can instead explicitly set the `Status` and add to the `ResponseTrailers` in the `ServerCallContext`
before returning from the call instead (this was documented in previous commit but was removed to keep it simple).  However the client will still receive an Exception since any non-OK response status generates an exception on the client.

src/Grpc.StatusProto/README.md
```diff
@@ -0,0 +1,303 @@
...
+```C#
+void PrintStatusDetails(RpcException ex)
+{
+    // Get the status from the RpcException
+    Google.Rpc.Status? rpcStatus = ex.GetRpcStatus(); // Extension method
+
+    if (rpcStatus != null)
+    {
+        // Decode each message item in the details in turn
+        foreach (var msg in status.UnpackDetailMessage())
```
codymullins, line 189: I think this example is incorrect, trying this locally doesn't build.

```suggestion
        foreach (var msg in rpcStatus.UnpackDetailMessages())
```

tonydnewell:
@apolcyn @JamesNK I think we're at a point where we need to decide what happens to the PR. Is there more work to do? Should it be merged? Or abandoned?

JamesNK:
> @apolcyn @JamesNK I think we're at a point where we need to decide what happens to the PR. Is there more work to do? Should it be merged? Or abandoned?

There are things I'd still like changed:

* Change extension method namespace to Grpc.Core: https://github.com/grpc/grpc-dotnet/pull/2205/files#r1316912203
* Move extension methods that don't have Grpc types to Google.Api.CommonProtos (they have TODO in source code). I suggest just creating a PR in that repo with the methods. Also, I suggest making them regular methods instead of extensions. That means the namespace doesn't matter.
* Throw error from GetRpcStatus by default: https://github.com/grpc/grpc-dotnet/pull/2205#discussion_r1316743256

tonydnewell:
FYI Changes to Google.Api.CommonProtos have been merged (moving the TODOs from this code to there). Waiting for Google.Api.CommonProtos 2.13.0 to be released before submitting further changes here.

jskeet:
> FYI Changes to Google.Api.CommonProtos have been merged (moving the TODOs from this code to there). Waiting for Google.Api.CommonProtos 2.13.0 to be released before submitting further changes here.

Just to set expectations, we're currently in a release freeze. I'd expect to release this around November 30th. I don't think there are any blockers for it other than the freeze - I may update to a newer version of protobuf first though (both in terms of generation and runtime).

JamesNK:
@tonydnewell What's the plan on finishing this? I can take over if you're unavailable.

tonydnewell:
> @tonydnewell What's the plan on finishing this? I can take over if you're unavailable.

It is on my list for this week. I'll let you know if I don't get to it.

tonydnewell:
@JamesNK I think I've covered all the review comments. I've also merged master into the branch as there were conflicting changes (package versioning changes). I just need to do some more sanity checks myself with some examples projects, but feel free to review the changes.

JamesNK (approved):

JamesNK:
Looks good. There are small things to clean up in XML comments but it doesn't need to be done now and this PR is long enough.

alrz:
Should this change https://github.com/dotnet/aspnetcore/pull/51394 use these APIs as well?

JamesNK:
Maybe. I created an issue: https://github.com/dotnet/aspnetcore/issues/55410

Issue #2206, opened 2023-07-21
Deadlock when cancelling active call while disposing channel


andrewhickman:
### What version of gRPC and what language are you using?

C#, using Grpc.Net.Client version 2.55.0

### What operating system (Linux, Windows,...) and version?

Windows 10.0.19042

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)

.NET 6.0.400

### What did you do?

We have a streaming gRPC call between two windows services which runs for the duration of the client service. When the client service is stopped, the call is canceled and the grpc client is disposed.

We have recently started seeing occasional hangs when the client is shutting down. The symptoms appear very similar to #2119, however I've confirmed that we are using version 2.55.0 of the client which should be fixed.

Debugging into the hung service, the cancellation callback for the service stop cancellation token seems to be stuck waiting for a lock in Grpc.Net.Client. I think the line numbers are referring to decompiled code; [this line](https://github.com/grpc/grpc-dotnet/blob/8ca08ebf05b0c81c15cb68baf5196ffe5ed7b5a3/src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs#L474C1-L474C12) is what visual studio points at for me.

<details>
 <summary>Stack trace for thread 219848</summary>

```
[Waiting on lock owned by Thread 213652, double-click or press enter to switch to thread]	
ntdll.dll!NtWaitForMultipleObjects()	Unknown
KERNELBASE.dll!00007ffb63f41b40()	Unknown
[Managed to Native Transition]	
Grpc.Net.Client.dll!Grpc.Net.Client.Internal.Retry.RetryCallBase<Google.Protobuf.WellKnownTypes.Empty, Spectrum.Site.SyncService.Grpc.SubscriptionList>.ClearRetryBuffer() Line 422	C#
Grpc.Net.Client.dll!Grpc.Net.Client.Internal.Retry.RetryCallBase<System.__Canon, System.__Canon>.Cleanup() Line 397	C#
Grpc.Net.Client.dll!Grpc.Net.Client.Internal.Retry.RetryCall<Google.Protobuf.WellKnownTypes.Empty, Spectrum.Site.SyncService.Grpc.SubscriptionList>.StartRetry(System.Action<Grpc.Net.Client.Internal.GrpcCall<Google.Protobuf.WellKnownTypes.Empty, Spectrum.Site.SyncService.Grpc.SubscriptionList>> startCallFunc) Line 185	C#
[...]

PR #2209, opened 2023-07-26
Fix gRPC client retry deadlock

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2206

I believe the problem is caused by a [call's TCS being created without `RunContinuationsAsynchronously`](https://github.com/grpc/grpc-dotnet/blob/8ca08ebf05b0c81c15cb68baf5196ffe5ed7b5a3/src/Grpc.Net.Client/Internal/GrpcCall.cs#L74). The simple fix would be to add that setting. Unfortunately, `RunContinuationsAsynchronously` is not used here on purpose because it breaks `Activity`.

When the call is disposed under a lock, the TCS is completed, and continuations are run on the same thread, still holding the lock. That includes logic after the [await here in RetryCall](https://github.com/grpc/grpc-dotnet/blob/8ca08ebf05b0c81c15cb68baf5196ffe5ed7b5a3/src/Grpc.Net.Client/Internal/Retry/RetryCall.cs#L255).

I moved disposing the active call outside of the lock. I think this fixes the deadlock, but I don't have repro to confirm.

Another idea I had, which isn't in this PR, is to place a `Task.Yield()` after the await in RetryCall. It forces the call's TCS to stop executing synchronously, `_callTcs.TrySetResult(...)` returns, the call's dispose finishes, and exits the lock.

cc @andrewhickman

### DIFF
--- a/src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
+++ b/src/Grpc.Net.Client/Internal/Retry/HedgingCall.cs
@@ -365,10 +365,10 @@ internal sealed partial class HedgingCall<TRequest, TResponse> : RetryCallBase<T
 
     protected override void Dispose(bool disposing)
     {
+        base.Dispose(disposing);
+
         lock (Lock)
         {
-            base.Dispose(disposing);
-
             CleanUpUnsynchronized();
         }
     }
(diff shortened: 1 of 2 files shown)


mgravell (approved):
Logic checks out

Issue #2210, opened 2023-07-26
Grpc.Core.RpcException: Status(StatusCode=\"Internal\", Detail=\"Error starting gRPC call. HttpRequestException: An HTTP/2 connection could not be established because the server did not complete the HTTP/2 handshake


jhyuzhijian:
### What version of gRPC and what language are you using?
Grpc.Net.Client / 2.56.0-dev202307260702

### What operating system (Linux, Windows,...) and version?
Linux containers in debian 

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.NET SDK 6.0.600

### What did you do?
If possible, provide a recipe for reproducing the error. Try being specific and include code snippets if helpful.
```c#
var services = new ServiceCollection();
        services.AddScoped<ClientLoggingInterceptor>();
        AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
        AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2Support", true);
        services.AddSingleton<ResolverFactory>(sp => new DnsResolverFactory(TimeSpan.FromSeconds(10)));
        services
             .AddGrpcClient<TClient>(factoryOptions =>
             {
                 factoryOptions.Address = new Uri("dns://");
             })
            .ConfigureChannel(channelOptions =>
            {
                channelOptions.Credentials = ChannelCredentials.Insecure;
                channelOptions.ServiceConfig = new ServiceConfig
                {
                    LoadBalancingConfigs =
            {
                new RoundRobinConfig()
            }
                };
                channelOptions.HttpHandler = new SocketsHttpHandler()
                {
                    EnableMultipleHttp2Connections = true,
                    SslOptions = new SslClientAuthenticationOptions
                    {
                        RemoteCertificateValidationCallback = delegate { return true; }
                    }
                };
            }).AddInterceptor<ClientLoggingInterceptor>(); 
public class ClientLoggingInterceptor : Interceptor
{
    private readonly ILogger _logger;

    public ClientLoggingInterceptor(ILoggerFactory loggerFactory)
    {
        _logger = loggerFactory.CreateLogger<ClientLoggingInterceptor>();
    }

    public override AsyncUnaryCall<TResponse> AsyncUnaryCall<TRequest, TResponse>(TRequest request, ClientInterceptorContext<TRequest, TResponse> context, AsyncUnaryCallContinuation<TRequest, TResponse> continuation)
    {
[...]

PR #2213, opened 2023-07-27
Support idle connection timeout with pending sockets

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2210

I think the issue is caused by a race when there is a long time interval between a channel connecting and a request being made:

1. When a request is made, the socket is double-checked to be healthy before it's given to SocketsHttpHandler to be turned into a connection.
2. After the health check, the server/proxy closes what it thinks is an idle socket.
3. The handler then uses the closed socket to send a request and fails.

The prospective fix is to close and recreate the socket if it exceeds the idle connection timeout. Default of 1 minute.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
+++ b/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
@@ -56,6 +56,7 @@ internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDi
     private readonly ILogger _logger;
     private readonly Subchannel _subchannel;
     private readonly TimeSpan _socketPingInterval;
+    private readonly TimeSpan _connectionIdleTimeout;
     private readonly Func<Socket, DnsEndPoint, CancellationToken, ValueTask> _socketConnect;
     private readonly List<ActiveStream> _activeStreams;
     private readonly Timer _socketConnectedTimer;
@@ -64,6 +65,7 @@ internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDi
     internal Socket? _initialSocket;
     private BalancerAddress? _initialSocketAddress;
     private List<ReadOnlyMemory<byte>>? _initialSocketData;
+    private DateTime? _initialSocketCreatedTime;
     private bool _disposed;
     private BalancerAddress? _currentAddress;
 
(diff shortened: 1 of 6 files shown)


JamesNK:
cc @jhyuzhijian @kalduzov

jhyuzhijian:
> cc @jhyuzhijian @kalduzov

#2210  the issue behavior
.net client request java server no problem
.net client request .net server has problem
so i think  close and recreate the socket if it exceeds the idle connection timeout. Default of 1 minute can not resove this question

JamesNK:
> Default of 1 minute can not resove this question

Why not? Kestrel has a default keep-alive timeout of 130 seconds.

https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserverlimits.keepalivetimeout?view=aspnetcore-7.0#microsoft-aspnetcore-server-kestrel-core-kestrelserverlimits-keepalivetimeout

jhyuzhijian:
> 

all right,i will test it

kalduzov:
We currently have 3 services in testing. On two of them, such an error occurs and they do not actively load the channel. Now about 1 error per hour. The third one is actively working with Grpc and there have been no such errors in the last 24 hours.

This partly confirms what you're thinking @JamesNK .

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -64,20 +65,23 @@ internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDi
...
     private List<ReadOnlyMemory<byte>>? _initialSocketData;
+    private DateTime? _initialSocketCreatedTime;
     private bool _disposed;
     private BalancerAddress? _currentAddress;
 
     public SocketConnectivitySubchannelTransport(
         Subchannel subchannel,
         TimeSpan socketPingInterval,
         TimeSpan? connectTimeout,
+        TimeSpan connectionIdleTimeout,
```
mgravell, line 76: nit naming... "idle timeout" suggests "if idle for X, kill it" - but this actually only applies during connect; maybe `connectIdleTimeout`?

mgravell (approved):
nit on naming (this is a "consider", not a "please change"), but: LGTM

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
JamesNK: "connect" in this context is the action of connecting the socket.

I renamed it to `socketIdleTimeout`. It's more accurate to what is idling.

PR #2215, opened 2023-07-29
Correctly check socket on stream creation

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2210

I recreated the error from https://github.com/grpc/grpc-dotnet/issues/2210. The previous fix (socket timeout) will address this issue for the vast majority of people. However, some servers may have configured a timeout of less than 60 seconds. They could still run into problems.

Previously if a server had written data to the socket (e.g. HTTP/2 GOAWAY frame) and closed it, then it would incorrectly be seen as healthy. This PR updates the stream creation method to correctly check that a socket is still healthy before using it. This involves reusing the check logic from the ping that reads available data. When available data is read, the client will see the socket is closed and create a new one for the connection.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
+++ b/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
@@ -253,53 +253,7 @@ internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDi
                 {
                     CompatibilityHelpers.Assert(socketAddress != null);
 
-                    try
-                    {
-                        SocketConnectivitySubchannelTransportLog.CheckingSocket(_logger, _subchannel.Id, socketAddress);
-
-                        // Poll socket to check if it can be read from. Unfortunatly this requires reading pending data.
-                        // The server might send data, e.g. HTTP/2 SETTINGS frame, so we need to read and cache it.
-                        //
-                        // Available data needs to be read now because the only way to determine whether the connection is closed is to
-                        // get the results of polling after available data is received.
-                        bool hasReadData;
-                        do
-                        {
-                            closeSocket = IsSocketInBadState(socket, socketAddress);
-                            var available = socket.Available;
-                            if (available > 0)
-                            {
-                                hasReadData = true;
-                                var serverDataAvailable = CalculateInitialSocketDataLength(_initialSocketData) + available;
-                                if (serverDataAvailable > MaximumInitialSocketDataSize)
-                                {
-                                    // Data sent to the client before a connection is started shouldn't be large.
-                                    // Put a maximum limit on the buffer size to prevent an unexpected scenario from consuming too much memory.
-                                    throw new InvalidOperationException($"The server sent {serverDataAvailable} bytes to the client before a connection was established. Maximum allowed data exceeded.");
-                                }
-
-                                SocketConnectivitySubchannelTransportLog.SocketReceivingAvailable(_logger, _subchannel.Id, socketAddress, available);
-
-                                // Data is already available so this won't block.
-                                var buffer = new byte[available];
-                                var readCount = socket.Receive(buffer);
-
-                                _initialSocketData ??= new List<ReadOnlyMemory<byte>>();
-                                _initialSocketData.Add(buffer.AsMemory(0, readCount));
-                            }
-                            else
-                            {
-                                hasReadData = false;
-                            }
-                        }
-                        while (hasReadData);
-                    }
-                    catch (Exception ex)
-                    {
-                        closeSocket = true;
-                        checkException = ex;
-                        SocketConnectivitySubchannelTransportLog.ErrorCheckingSocket(_logger, _subchannel.Id, socketAddress, ex);
-                    }
+                    closeSocket = CheckSocket(socket, socketAddress, ref _initialSocketData, out checkException);
                 }
             }
 
@@ -334,6 +288,68 @@ internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDi
                 // Schedule next ping.
                 _socketConnectedTimer.Change(_socketPingInterval, Timeout.InfiniteTimeSpan);
             }
+        }
+    }
+
+    /// <summary>
+    /// Checks whether the socket is healthy. May read available data into the passed in buffer.
+    /// Returns true if the socket should be closed.
+    /// </summary>
+    private bool CheckSocket(Socket socket, BalancerAddress socketAddress, ref List<ReadOnlyMemory<byte>>? socketData, out Exception? checkException)
+    {
+        checkException = null;
+
+        try
+        {
+            SocketConnectivitySubchannelTransportLog.CheckingSocket(_logger, _subchannel.Id, socketAddress);
+
+            // Poll socket to check if it can be read from. Unfortunatly this requires reading pending data.
+            // The server might send data, e.g. HTTP/2 SETTINGS frame, so we need to read and cache it.
+            //
+            // Available data needs to be read now because the only way to determine whether the connection is closed is to
+            // get the results of polling after available data is received.
+            bool hasReadData;
+            do
+            {
+                if (IsSocketInBadState(socket, socketAddress))
+                {
+                    return true;
+                }
+
+                var available = socket.Available;
+                if (available > 0)
+                {
+                    hasReadData = true;
+                    var serverDataAvailable = CalculateInitialSocketDataLength(socketData) + available;
+                    if (serverDataAvailable > MaximumInitialSocketDataSize)
+                    {
+                        // Data sent to the client before a connection is started shouldn't be large.
+                        // Put a maximum limit on the buffer size to prevent an unexpected scenario from consuming too much memory.
+                        throw new InvalidOperationException($"The server sent {serverDataAvailable} bytes to the client before a connection was established. Maximum allowed data exceeded.");
+                    }
+
+                    SocketConnectivitySubchannelTransportLog.SocketReceivingAvailable(_logger, _subchannel.Id, socketAddress, available);
+
+                    // Data is already available so this won't block.
+                    var buffer = new byte[available];
+                    var readCount = socket.Receive(buffer);
+
+                    socketData ??= new List<ReadOnlyMemory<byte>>();
+                    socketData.Add(buffer.AsMemory(0, readCount));
+                }
+                else
+                {
+                    hasReadData = false;
+                }
+            }
+            while (hasReadData);
+            return false;
+        }
+        catch (Exception ex)
+        {
+            checkException = ex;
+            SocketConnectivitySubchannelTransportLog.ErrorCheckingSocket(_logger, _subchannel.Id, socketAddress, ex);
+            return true;
         }
     }
 
@@ -383,7 +399,7 @@ internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDi
                 SocketConnectivitySubchannelTransportLog.ClosingSocketFromIdleTimeoutOnCreateStream(_logger, _subchannel.Id, address, _socketIdleTimeout);
                 closeSocket = true;
             }
-            else if (IsSocketInBadState(socket, address))
+            else if (CheckSocket(socket, address, ref socketData, out _))
             {
                 SocketConnectivitySubchannelTransportLog.ClosingUnusableSocketOnCreateStream(_logger, _subchannel.Id, address);
                 closeSocket = true;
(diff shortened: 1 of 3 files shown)


src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -337,6 +291,68 @@ private void OnCheckSocketConnection(object? state)
...
+                    socketData ??= new List<ReadOnlyMemory<byte>>();
+                    socketData.Add(buffer.AsMemory(0, readCount));
+                }
+                else
+                {
+                    hasReadData = false;
+                }
+            }
+            while (hasReadData);
+            return false;
```
Wolffy101, line 346: I recommend using **return IsSocketInBadState(socket, socketAddress)**
Because aspnetcore server timeout send FIN and close connection, Poll forever return true and available = 0 in read bytes after

JamesNK: I don't understand. `IsSocketInBadState` is called [at line 314](https://github.com/grpc/grpc-dotnet/pull/2215/files#diff-99c512af792b0c055d726a6532a19e8b4df851a45c4d10a5804ceacf9e934088R314-R317). If it is true then true is returned even if Available is 0.

I don't see the point of calling `IsSocketInBadState` here because we've already checked that we can read from the socket and there is no available data.

Wolffy101: sorry , I read 341 line wrong

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -337,6 +291,68 @@ private void OnCheckSocketConnection(object? state)
...
+    private bool CheckSocket(Socket socket, BalancerAddress socketAddress, ref List<ReadOnlyMemory<byte>>? socketData, out Exception? checkException)
+    {
+        checkException = null;
+
+        try
+        {
+            SocketConnectivitySubchannelTransportLog.CheckingSocket(_logger, _subchannel.Id, socketAddress);
+
+            // Poll socket to check if it can be read from. Unfortunatly this requires reading pending data.
+            // The server might send data, e.g. HTTP/2 SETTINGS frame, so we need to read and cache it.
```
JamesNK, line 307:
```suggestion
            // The server might send data, such as an HTTP/2 SETTINGS or GOAWAY frame, so we need to read and cache it.
```

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -337,6 +291,68 @@ private void OnCheckSocketConnection(object? state)
         }
     }
 
+    /// <summary>
+    /// Checks whether the socket is healthy. May read available data into the passed in buffer.
+    /// Returns true if the socket should be closed.
+    /// </summary>
+    private bool CheckSocket(Socket socket, BalancerAddress socketAddress, ref List<ReadOnlyMemory<byte>>? socketData, out Exception? checkException)
```
amcasey, line 298: I think this name could be clearer.  It returns `true` if the check failed and the socket should be closed?  What about something like "ShouldCloseSocket"?

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -253,53 +253,7 @@ private void OnCheckSocketConnection(object? state)
...
-                        }
-                        while (hasReadData);
-                    }
-                    catch (Exception ex)
-                    {
-                        closeSocket = true;
-                        checkException = ex;
-                        SocketConnectivitySubchannelTransportLog.ErrorCheckingSocket(_logger, _subchannel.Id, socketAddress, ex);
-                    }
+                    closeSocket = CheckSocket(socket, socketAddress, ref _initialSocketData, out checkException);
```
amcasey, line 256: It's not a change, but I'm a little surprised this call site has less logging than the other one.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -337,6 +291,68 @@ private void OnCheckSocketConnection(object? state)
...
+                var available = socket.Available;
+                if (available > 0)
+                {
+                    hasReadData = true;
+                    var serverDataAvailable = CalculateInitialSocketDataLength(socketData) + available;
+                    if (serverDataAvailable > MaximumInitialSocketDataSize)
+                    {
+                        // Data sent to the client before a connection is started shouldn't be large.
+                        // Put a maximum limit on the buffer size to prevent an unexpected scenario from consuming too much memory.
+                        throw new InvalidOperationException($"The server sent {serverDataAvailable} bytes to the client before a connection was established. Maximum allowed data exceeded.");
```
amcasey, line 328: Not new but it's a little ugly to throw and then immediately catch this.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -337,6 +291,68 @@ private void OnCheckSocketConnection(object? state)
...
+                    // Data is already available so this won't block.
+                    var buffer = new byte[available];
+                    var readCount = socket.Receive(buffer);
+
+                    socketData ??= new List<ReadOnlyMemory<byte>>();
+                    socketData.Add(buffer.AsMemory(0, readCount));
+                }
+                else
+                {
+                    hasReadData = false;
```
amcasey, line 342: Seems like we could just break out of an infinite loop, rather than updating a flag variable.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -337,6 +291,68 @@ private void OnCheckSocketConnection(object? state)
...
+    /// </summary>
+    private bool CheckSocket(Socket socket, BalancerAddress socketAddress, ref List<ReadOnlyMemory<byte>>? socketData, out Exception? checkException)
+    {
+        checkException = null;
+
+        try
+        {
+            SocketConnectivitySubchannelTransportLog.CheckingSocket(_logger, _subchannel.Id, socketAddress);
+
+            // Poll socket to check if it can be read from. Unfortunatly this requires reading pending data.
```
amcasey, line 306: Typo: Unfortunatly

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -337,6 +291,68 @@ private void OnCheckSocketConnection(object? state)
...
+            // The server might send data, such as an HTTP/2 SETTINGS or GOAWAY frame, so we need to read and cache it.
```
amcasey, line 307: Available data needs to be read now because the only way to determine whether the connection is closed is to get the results of polling after available data is received.  For example, the server may have sent an HTTP/2 SETTINGS or GOAWAY frame.  We need to cache whatever we read so it isn't dropped.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -337,6 +291,68 @@ private void OnCheckSocketConnection(object? state)
...
+                }
+
+                var available = socket.Available;
+                if (available > 0)
+                {
+                    hasReadData = true;
+                    var serverDataAvailable = CalculateInitialSocketDataLength(socketData) + available;
+                    if (serverDataAvailable > MaximumInitialSocketDataSize)
+                    {
+                        // Data sent to the client before a connection is started shouldn't be large.
```
amcasey, line 326: Does this mean that we'll treat a connection with more data than we expect as in a bad state?

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -383,7 +399,7 @@ public async ValueTask<Stream> GetStreamAsync(BalancerAddress address, Cancellat
                 SocketConnectivitySubchannelTransportLog.ClosingSocketFromIdleTimeoutOnCreateStream(_logger, _subchannel.Id, address, _socketIdleTimeout);
                 closeSocket = true;
             }
-            else if (IsSocketInBadState(socket, address))
+            else if (CheckSocket(socket, address, ref socketData, out _))
```
amcasey, line 402: Is this the only caller that needs to be updated?  When would you want to make a raw `IsSocketInBadState` call?

test/FunctionalTests/Balancer/ConnectionTests.cs
```diff
@@ -180,6 +180,55 @@ Task<HelloReply> UnaryMethod(HelloRequest request, ServerCallContext context)
...
+        {
+            return Task.FromResult(new HelloReply { Message = request.Name });
+        }
+
+        // In this test the client connects to the server, and the server then closes it after keep-alive is triggered.
+        // The client then starts a gRPC call to the server. The client should discard the closed socket and create a new one.
+
+        // Arrange
+        using var endpoint = BalancerHelpers.CreateGrpcEndpoint<HelloRequest, HelloReply>(
+            50051,
```
amcasey, line 202: Is this a port number and, if so, do we need to worry about collisions?

test/FunctionalTests/Balancer/ConnectionTests.cs
```diff
@@ -180,6 +180,55 @@ Task<HelloReply> UnaryMethod(HelloRequest request, ServerCallContext context)
...
+        {
+            return true;
+        });
+
+        Task<HelloReply> UnaryMethod(HelloRequest request, ServerCallContext context)
+        {
+            return Task.FromResult(new HelloReply { Message = request.Name });
+        }
+
+        // In this test the client connects to the server, and the server then closes it after keep-alive is triggered.
```
amcasey, line 197: Would the behavior be different if the server aborted the connection, rather than timing out?  AFAIK, the kestrel code paths are the same, so I'm guessing not?

amcasey (approved):
Seems sensible to me.  I don't consider any of my suggestions blocking.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
JamesNK: Yes. A socket shouldn't send much data before it is used.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
JamesNK: The exception gets logged and pass out in an out parameter.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
JamesNK: You wouldn't. It now is only used by the method that also consumes data. I'll leave it as its own method to prevent the other one from getting too big. And add a comment.

test/FunctionalTests/Balancer/ConnectionTests.cs
JamesNK: It's a port number. It's ok as long as the endpoint is disposed.

test/FunctionalTests/Balancer/ConnectionTests.cs
JamesNK: Depends on the kind of abort. A HTTP/2 abort would also send a GOAWAY. A transport-level aport would just close the socket. There are tests that already cover a transport-level abort.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
JamesNK: I added a lot of logging in to the other call site while trying to figure out this bug 😄

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
amcasey: You could just set the out parameter here, but you'd have to duplicate the logging statement, so it's probably not worth it.

PR #2216, opened 2023-07-31
Upgrade Newtonsoft.Json package version

sebastienros:
ASP.NET benchmarks don't compile anymore with the current version due to public vulnerability warnings.


sebastienros:
@JamesNK

JamesNK:
There are new nullable warnings to fix.

Issue #2225, opened 2023-08-03
New attributes in BalancerAddress are ignored unless Endpoint has changed


kalduzov:
@JamesNK , hi.

We encountered a very incomprehensible behavior for us in the balancing implementation.

In our implementation, which is inherited from `SubchannelsLoadBalancer`, we add attributes for `BalancerAddress `to the `Attributes `property for balancing by mesh versions.

Our pod address can change its mesh version while the grpc client is running. Then a new list of addresses arrives, in which the `Endpoint `has not actually changed, but the composition of the attributes for it has changed.

However, the implementation in `SubchannelsLoadBalancer `does not allow changing the current subchannel address, because if the subchannel matches the one that came in, then the "new" address with new attributes is ignored and the old subchannel is used.

We think that at least the presence of new attributes on the `BalancerAddress `should be reflected in the existing subchannel, because from the point of view of logic, `BalancerAddress `is not only an address, but also its additional properties, which are now ignored.

It seems that in general it will be enough to do in the method
`UpdateChannelState(ChannelState state) `call like
such `newOrCurrentSubchannel.Subchannel.UpdateAddresses(new[] { address });`

and correct the conditions in `UpdateAddresses `to update/add at least attributes to an existing address.

Or add a new mechanism to forward these attributes to an existing subchannel.

### What version of gRPC and what language are you using?

All versions with client side balancing

JamesNK:
Why are changing the attributes useful for you? I'd like to understand what you're using them for.

What do you think should happen to a connection if its `BalancerAddress` has different attributes? Should it be recreated (socket closed, and reopened) or should it continue to be connected but with a different `BalancerAddress`?

JamesNK:
A version with this change in it is available now - https://www.nuget.org/packages/Grpc.Net.Client/2.57.0-pre1

Please try it out. If you see anything wrong, it can be fixed in a the next week. Otherwise, you'll have to wait a few months for the next release.

kalduzov:
We'll check it out soon.

kalduzov:
We'll check it out soon.

kalduzov:
Unfortunately, we were unable to test the solution this week. :(

JamesNK:
Ok, no problem. If there are issues, then you'll need to wait for the next release.

2.57.0 is already out with the change. You can try that.

PR #2228, opened 2023-08-04
Change subchannel BalancerAddress when attributes change

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2225

This PR:
* Updates how `BalancerAttributes` are stored to always use `Dictionary<string, object>`. This is to improve comparison performance.
* Adds an equals comparison on subchannels that includes `BalancerAttributes`.
* Changes subchannel transport to just store the `DnsEndPoint`. The endpoint is then used to look up the balancer attributes to use in the subchannel.
* Changes subchannel transport to use `DnsEndPoint` instead of `BalancerAttributes` in all logs.
* A `BalancerAddress` with different attributes replaces the address on the subchannel. If the subchannel has an open connection, it stays open.

cc @kalduzov

### DIFF
--- a/src/Grpc.Net.Client/Balancer/BalancerAddress.cs
+++ b/src/Grpc.Net.Client/Balancer/BalancerAddress.cs
@@ -1,4 +1,4 @@
-﻿#region Copyright notice and license
+#region Copyright notice and license
 
 // Copyright 2019 The gRPC Authors
 //
@@ -30,7 +30,9 @@ namespace Grpc.Net.Client.Balancer;
 /// </summary>
 public sealed class BalancerAddress
 {
-    private BalancerAttributes? _attributes;
+    // Internal so address attributes can be compared without using the Attributes property.
+    // The property allocates an empty collection if one isn't already present.
+    internal BalancerAttributes? _attributes;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="BalancerAddress"/> class with the specified <see cref="DnsEndPoint"/>.
@@ -48,7 +50,7 @@ public sealed class BalancerAddress
     /// <param name="host">The host.</param>
     /// <param name="port">The port.</param>
     [DebuggerStepThrough]
-    public BalancerAddress(string host, int port) : this(new DnsEndPoint(host, port))
+    public BalancerAddress(string host, int port) : this(new BalancerEndPoint(host, port))
     {
     }
 
@@ -68,6 +70,22 @@ public sealed class BalancerAddress
     public override string ToString()
     {
         return $"{EndPoint.Host}:{EndPoint.Port}";
+    }
+
+    private sealed class BalancerEndPoint : DnsEndPoint
+    {
+        private string? _cachedToString;
+
+        public BalancerEndPoint(string host, int port) : base(host, port)
+        {
+        }
+
+        public override string ToString()
+        {
+            // Improve ToString performance when logging by caching ToString.
+            // Don't include DnsEndPoint address family.
+            return _cachedToString ??= $"{Host}:{Port}";
+        }
     }
 }
 #endif
--- a/src/Grpc.Net.Client/Balancer/BalancerAttributes.cs
+++ b/src/Grpc.Net.Client/Balancer/BalancerAttributes.cs
@@ -38,20 +38,22 @@ public sealed class BalancerAttributes : IDictionary<string, object?>, IReadOnly
     /// <summary>
     /// Gets a read-only collection of metadata attributes.
     /// </summary>
-    public static readonly BalancerAttributes Empty = new BalancerAttributes(new ReadOnlyDictionary<string, object?>(new Dictionary<string, object?>()));
+    public static readonly BalancerAttributes Empty = new BalancerAttributes(new Dictionary<string, object?>(), readOnly: true);
 
-    private readonly IDictionary<string, object?> _attributes;
+    private readonly Dictionary<string, object?> _attributes;
+    private readonly bool _readOnly;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="BalancerAttributes"/> class.
     /// </summary>
-    public BalancerAttributes() : this(new Dictionary<string, object?>())
+    public BalancerAttributes() : this(new Dictionary<string, object?>(), readOnly: false)
     {
     }
 
-    private BalancerAttributes(IDictionary<string, object?> attributes)
+    private BalancerAttributes(Dictionary<string, object?> attributes, bool readOnly)
     {
         _attributes = attributes;
+        _readOnly = readOnly;
     }
 
     object? IDictionary<string, object?>.this[string key]
@@ -62,6 +64,7 @@ public sealed class BalancerAttributes : IDictionary<string, object?>, IReadOnly
         }
         set
         {
+            ValidateReadOnly();
             _attributes[key] = value;
         }
     }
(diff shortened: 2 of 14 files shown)


kalduzov:
@JamesNK  hi.

And why is the work on this PR not moving?

src/Grpc.Net.Client/Balancer/BalancerAddress.cs
```diff
@@ -30,7 +30,7 @@ namespace Grpc.Net.Client.Balancer;
 /// </summary>
 public sealed class BalancerAddress
 {
-    private BalancerAttributes? _attributes;
+    internal BalancerAttributes? _attributes;
```
mgravell, line 33: I get nervous about exposed fields; ok, in reality it isn't going to cause harm, but it *feels icky*; maybe I'm overthinking and adding artificial internal properties etc is just the same, with extra steps

src/Grpc.Net.Client/Balancer/BalancerAttributes.cs
```diff
@@ -38,20 +38,22 @@ public sealed class BalancerAttributes : IDictionary<string, object?>, IReadOnly
     /// <summary>
     /// Gets a read-only collection of metadata attributes.
     /// </summary>
-    public static readonly BalancerAttributes Empty = new BalancerAttributes(new ReadOnlyDictionary<string, object?>(new Dictionary<string, object?>()));
```
mgravell, line 41: you cite performance as the reason for this change; I would be genuinely surprized if `Dictionary<string,object?` out-performed `ReadOnlyDictionary<string,object?>` except on the custom struct iterator - but if perf is key, I also wonder whether we should be looking to `FrozenDictionary<string, object?>` on TFMs where it is available, via `#if`; is there data behind this change? If the enumerator is the big difference, `FrozenDictrionary<string,object?>` has a custom iterator too, and `foreach` isn't a common scenario on dictionary, usually

src/Grpc.Net.Client/Balancer/Internal/BalancerAddressEqualityComparer.cs
```diff
@@ -44,6 +44,46 @@ public bool Equals(BalancerAddress? x, BalancerAddress? y)
...
+        var yAttributes = y._attributes?._attributes;
+        if (!AttributesEqual(xAttributes, yAttributes))
+        {
+            return false;
+        }
+
+        return true;
+    }
+
+    private bool AttributesEqual(Dictionary<string, object?>? x, Dictionary<string, object?>? y)
```
mgravell, line 57: nit: maybe a `ReferenceEquals` check first?

nit: as a general guidance, I've been trying to push people to always prefer `is null` / `is not null` when performing `null` tests - it isn't a big thing here, but it can make a big difference when custom equality operators are in play (for example `string`); entirely up to you, but I see only upside to `is null` vs `== null` etc

src/Grpc.Net.Client/Balancer/Subchannel.cs
```diff
@@ -69,7 +69,7 @@ public sealed class Subchannel : IDisposable
     /// <summary>
     /// Gets the current connected address.
     /// </summary>
-    public BalancerAddress? CurrentAddress => _transport.CurrentAddress;
+    public BalancerAddress? CurrentAddress => GetAddress(_transport.CurrentEndPoint);
```
mgravell, line 72: how often is this invoked? i.e. is there a CPU or allocation problem here? (I notice that `GetAddress` is non-trivial); is this worth memoizing (remembering to wipe the memoized value if a relevant config change occurs)

mgravell:
@kalduzov please blame me for that; review comments now added

kalduzov:
> @kalduzov please blame me for that; review comments now added

All OK. It’s just that the previous PRs were quickly reviewed, and this is just silence. We are very much looking forward to this change. Without it, all our written balancing code does not work :(

mgravell:
because I like data, I did this: https://gist.github.com/mgravell/61c3a892020860a87a3b6cb3ad1687a5

short version: for key lookup (`FindKeys`), it really doesn't make much difference what you choose - for enumerator, there is a cost delta, and `FrozenDictionary<,>` is *bad* for that

src/Grpc.Net.Client/Balancer/BalancerAttributes.cs
JamesNK: The performance gain isn't `Dictionary` vs `ReadOnlyDictionary`. The performance gain is from consistently referencing the values as Dictionary instead of IDictionary.

Using `Dictionary` avoids interface overhead and allows `foreach` over the dictionary without allocating.

src/Grpc.Net.Client/Balancer/Subchannel.cs
JamesNK: The reason I changed this to be lazily calculated is layering. It's hard to cache and get notifications when it changes. Fortunately, the property is rarely used. Load balancing is built around caching at a higher level.

The property is public API, so it can't change. If it could, then a method would be better, but the ship has sailed.

src/Grpc.Net.Client/Balancer/Internal/BalancerAddressEqualityComparer.cs
JamesNK: > nit: maybe a ReferenceEquals check first (replaces both-null)?

Done.

> is null

The types here are all known and don't have custom equals. I agree in principle, but not enough to change.

src/Grpc.Net.Client/Balancer/BalancerAddress.cs
amcasey: I also don't love exposing a backing field, in particular.  Personally, I'd probably add 
```csharp
internal BalancerAttributes? AttributesOrNull => _attributes;
```
I'd expect that accessor to be inlined anyway.

src/Grpc.Net.Client/Balancer/BalancerAddress.cs
```diff
@@ -69,5 +71,21 @@ public override string ToString()
...
+
+        public BalancerEndPoint(string host, int port) : base(host, port)
+        {
+        }
+
+        public override string ToString()
+        {
+            // Improve ToString performance when logging by caching ToString.
+            // Don't include DnsEndPoint address family.
+            return _cachedToString ??= $"{Host}:{Port}";
```
amcasey, line 87: I assume, from context, that this is logged _a lot_?

src/Grpc.Net.Client/Balancer/BalancerAttributes.cs
```diff
@@ -62,28 +64,49 @@ private BalancerAttributes(IDictionary<string, object?> attributes)
...
+            ValidateReadOnly();
             _attributes[key] = value;
         }
     }
 
     ICollection<string> IDictionary<string, object?>.Keys => _attributes.Keys;
     ICollection<object?> IDictionary<string, object?>.Values => _attributes.Values;
     int ICollection<KeyValuePair<string, object?>>.Count => _attributes.Count;
-    bool ICollection<KeyValuePair<string, object?>>.IsReadOnly => _attributes.IsReadOnly;
+    bool ICollection<KeyValuePair<string, object?>>.IsReadOnly => _readOnly ? true : ((ICollection<KeyValuePair<string, object?>>)_attributes).IsReadOnly;
```
amcasey, line 75: I might have written `_readonly ||`

src/Grpc.Net.Client/Balancer/BalancerAttributes.cs
```diff
@@ -135,10 +159,55 @@ public void Set<TValue>(BalancerAttributesKey<TValue> key, TValue value)
...
+        }
+
+        if (xValues.Count != yValues.Count)
+        {
+            return false;
+        }
+
+        foreach (var kvp in xValues)
+        {
+            if (!yValues.TryGetValue(kvp.Key, out var value))
```
amcasey, line 196: Is it interesting to consider the case where the dictionaries have different key comparers?

src/Grpc.Net.Client/Balancer/Subchannel.cs
```diff
@@ -402,6 +406,38 @@ internal void RaiseStateChanged(ConnectivityState state, Status status)
...
+                }
+            }
+        }
+
+        return null;
+    }
+
+    private static bool HasMatchingEndpoint(List<BalancerAddress> addresses, BalancerAddress currentAddress)
+    {
+        foreach (var a in addresses)
```
amcasey, line 430: This is suspiciously similar to the loop in the method above.

amcasey (commented):
Initial notes.  I'll take a higher level view tomorrow.

src/Grpc.Net.Client/Balancer/BalancerAddress.cs
JamesNK: Not that much. But it's simple enough to cache.

src/Grpc.Net.Client/Balancer/BalancerAttributes.cs
JamesNK: The dictionaries will never have different key comparers.

mgravell (approved):

Issue #2226, opened 2023-08-03
Validate Windows version when using WinHttpHandler


JamesNK:
https://github.com/grpc/grpc-dotnet/pull/2220 adds a net462 target and automatically wires up WinHttpHandler.

WinHttpHandler has some Windows version requirements. Constructing a channel should throw an error if WinHttpHandler is used on unsupported versions of Windows.

Note: on Windows Server 2022 there is partial support for WinHttpHandler. Unary and server streaming calls succeed, client and bidi streaming calls fail. Channel shouldn't throw an error on Windows Server 2022 to not block people who only need unary on Win2022.

Related: https://github.com/grpc/grpc-dotnet/issues/2197

jskeet:
For the "note": I understand the motivation here, and not throwing is probably the right call - but it would be really useful to have some way of detecting this without each user having to write the Windows Server 2022 detection code. (Perhaps something in GrpcChannelOptions to indicate which features are required, and throw if they're not all met? I'm not sure how feasible that is.)

PR #2229, opened 2023-08-04
Validate Windows version when using WinHttpHandler

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2226

cc @jtattermusch @jskeet

### DIFF
--- a/src/Grpc.Net.Client/GrpcChannel.cs
+++ b/src/Grpc.Net.Client/GrpcChannel.cs
@@ -184,6 +184,18 @@ public sealed class GrpcChannel : ChannelBase, IDisposable
         {
             Log.AddressPathUnused(Logger, Address.OriginalString);
         }
+
+        // Validate the Windows version can support WinHttpHandler.
+        const int WinServer2022BuildVersion = 20348;
+        if (HttpHandlerType == HttpHandlerType.WinHttpHandler &&
+            OperatingSystem.IsWindows &&
+            OperatingSystem.OSVersion.Build < WinServer2022BuildVersion)
+        {
+            throw new InvalidOperationException("The channel configuration isn't valid on this operating system. " +
+                "The channel is configured to use WinHttpHandler and the current version of Windows " +
+                "doesn't support HTTP/2 features required by gRPC. Windows Server 2022 or Windows 11 or later is required. " +
+                "For more information, see https://aka.ms/aspnet/grpc/netframework.");
+        }
     }
 
     private void ResolveCredentials(GrpcChannelOptions channelOptions, out bool isSecure, out List<CallCredentials>? callCredentials)
--- a/src/Grpc.Net.Client/Internal/OperatingSystem.cs
+++ b/src/Grpc.Net.Client/Internal/OperatingSystem.cs
@@ -1,4 +1,4 @@
-﻿#region Copyright notice and license
+#region Copyright notice and license
 
 // Copyright 2019 The gRPC Authors
 //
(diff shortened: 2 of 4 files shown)


mgravell (approved):

jskeet:
Would still love to see a way of interrogating whether I can expect full gRPC feature support, but this is definitely a good start.

JamesNK:
> Would still love to see a way of interrogating whether I can expect full gRPC feature support, but this is definitely a good start.

There isn't one. WinHTTP doesn't have APIs to say whether the required features are available. The only test is the Windows version.

If OS build version is greater or equal to WinServer 2022 then there is partial support.
If OS build version is greater or equal to Win 11 then there is full support.

It's not perfect, but testing the build version isn't difficult.

jskeet:
> There isn't one. WinHTTP doesn't have APIs to say whether the required features are available. The only test is the Windows version.

Sorry, I should have been clearer. That makes sense - but it makes less sense to me for each developer to have to find the details of precise OS/build versions, when that logic can be centralized - and when we're already fetching the build version for this feature.

jskeet:
(If it's *only* Google that needs to do this, then I'm happy to put it in our code. I'd just be sad to see basically the same code in 100 different codebases around the world.)

src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -184,6 +184,18 @@ internal GrpcChannel(Uri address, GrpcChannelOptions channelOptions) : base(addr
         {
             Log.AddressPathUnused(Logger, Address.OriginalString);
         }
+
+        // Validate the Windows version can support WinHttpHandler.
+        const int WinServer2022BuildVersion = 20348;
```
jtattermusch, line 189: I think the comment needs to elaborate on why this particular build version was chosen.
Basically, as you said
- If OS build version is greater or equal to WinServer 2022 then there is partial support.
- If OS build version is greater or equal to Win 11 then there is full support.

The also need to be a note that this will allow creating GrpcChannel on WinServer 2022 with "partial support" (and explain what that is).

src/Grpc.Net.Client/Internal/OperatingSystem.cs
```diff
@@ -41,5 +45,7 @@ private OperatingSystem()
 #else
         IsAndroid = false;
 #endif
+        IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+        OSVersion = Environment.OSVersion.Version;
```
jtattermusch, line 49: excerpt from offline discussion: 

> I experimented with Environment.OSVersion and one issue I found is that when called from .NET Framework, it seems to return garbage on newer windows versions (e.g. I tried on Microsoft Windows Server 2019 Datacenter, 10.0.17763 Build 17763 and it returned "Microsoft Windows NT 6.2.9200.0"). Looks like the value returned is the latest windows version the .NET Framework knows about (Which is way too old now and not useful to us).
So some better way of detecting OS Version will be needed.

Can you confirm that this has been taken into account and that we really rely on Environment.OSVersion.Version providing the right info when on .NET Framework (which is actually where we care about it the most)?
This would also deserve a test that actually targets net462 (not sure if we currently have one).

jtattermusch: Btw, if Environment.OSVersion.Version gives outdated info about build number, in practice this would mean that a high-enough build number would never be detected and creating the GrpcChannel would always throw on .NET Framework?

Issue #2180, opened 2023-06-27
Remove .NET Core 3 and .NET 5 targets


Varorbc:
Why are unsupported target frameworks still supported? For example, netcoreapp3.0 and net5.0

JamesNK:
Hasn't been a reason to yet. Might remove them when .NET 8 target is added.

JamesNK:
Move to ActivitySource when old targets are dropped - https://github.com/open-telemetry/opentelemetry-dotnet/pull/4695#discussion_r1276720269

Add force yield in RetryCall and HedgeCall of GrpcCall.CallTask in .NET 8.

JamesNK:
.NET standard 2.0 is important. Maybe for the others.

PR #2021, opened 2023-01-17
Update to use .NET 8

JamesNK:
~Work in progress. Don't review.~

* Add .NET 8 target
* In .NET 8 target, use `IHttpRequestTimeoutFeature` to disable timeout middleware for streaming methods
* Remove .NET Core 3 and .NET 5 targets (out of support)
* Add .NET 8 named pipes option to benchmarks
* Use argument helpers
* Update tests to consistently target NET462 for .NET Framework

PR #2244, opened 2023-08-15
Update Grpc.Net.Client to use ActivitySource

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/2180

* Use ActivitySource
* Remove unused telemetry handler (no longer needed since the .NET 5 target is removed) and some using statements


JamesNK:
re: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4695#discussion_r1276720269

cc @tarekgh @noahfalk @eerhardt

Let me know if this fixes the issue or if more changes are required.

test/Grpc.Net.Client.Tests/DiagnosticsTests.cs
```diff
@@ -164,8 +164,14 @@ public void DiagnosticListener_MakeCall_ActivityHasNameAndDuration()
         };
 
         // Act
+        using (var activityListener = new ActivityListener())
         using (GrpcDiagnostics.DiagnosticListener.Subscribe(new ActionObserver<KeyValuePair<string, object?>>(onDiagnosticMessage)))
         {
+            activityListener.ShouldListenTo = activitySource => activitySource == GrpcDiagnostics.ActivitySource;
+            activityListener.Sample = (ref ActivityCreationOptions<ActivityContext> _) => ActivitySamplingResult.AllData;
+
+            ActivitySource.AddActivityListener(activityListener);
```
eerhardt, line 173: Do you want to assert something got written to the activityListener?

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -813,9 +813,12 @@ private void SetFailedResult(Status status)
         // 3. There is an existing activity (to enable activity propagation)
         if (diagnosticSourceEnabled || Logger.IsEnabled(LogLevel.Critical) || Activity.Current != null)
         {
-            activity = new Activity(GrpcDiagnostics.ActivityName);
-            activity.AddTag(GrpcDiagnostics.GrpcMethodTagName, Method.FullName);
-            activity.Start();
+            activity = GrpcDiagnostics.ActivitySource.CreateActivity(GrpcDiagnostics.ActivityName, ActivityKind.Internal);
```
eerhardt, line 816: Why isn't this

```suggestion
            activity = GrpcDiagnostics.ActivitySource.CreateActivity(GrpcDiagnostics.ActivityName, ActivityKind.Client);
```

?

That's what OpenTelemetry is setting it to today:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/031ed48714e16ba4a5b099b6e14647994a0b9c1b/src/OpenTelemetry.Instrumentation.GrpcNetClient/Implementation/GrpcClientDiagnosticListener.cs#L127

tarekgh: CC @reyang @CodeBlanch

tarekgh:
>         var diagnosticSourceEnabled = GrpcDiagnostics.DiagnosticListener.IsEnabled() &&

Is this the old way to check for DiagnosticSource? if so, I think this may need to change too to use [https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitysource.haslisteners?view=net-7.0](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitysource.haslisteners?view=net-7.0)?

---
Refers to: src/Grpc.Net.Client/Internal/GrpcCall.cs:806 in 83d9037. [](commit_id = 83d9037bc3381b5410e90953d54232862173caa6, deletion_comment = False)

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -813,9 +813,12 @@ private void SetFailedResult(Status status)
...
+            if (activity != null)
```
tarekgh, line 817: do we care about the compatibility scenario here? previously if `diagnosticSourceEnabled` is enabled, the activity creation is forced. Now it is not guaranteed having the Activity gets created. Old listener clients will start miss activities logging.

src/Grpc.Net.Client/Internal/GrpcCall.cs
tarekgh, line 816: Instead of creating the Activity then setting the tags and starting it, you can call StartActivity and passing the tags list using TagList structure.

CodeBlanch: When it is gRPC over HTTP we [have an option which allows users to ignore the HTTP span](https://github.com/open-telemetry/opentelemetry-dotnet/blob/031ed48714e16ba4a5b099b6e14647994a0b9c1b/src/OpenTelemetry.Instrumentation.GrpcNetClient/GrpcClientInstrumentationOptions.cs#L48) so they don't see 2 spans for every call. For that reason maybe we should go with `ActivityKind.Client` for these?

/cc @alanwest

reyang: > When it is gRPC over HTTP we [have an option which allows users to ignore the HTTP span](https://github.com/open-telemetry/opentelemetry-dotnet/blob/031ed48714e16ba4a5b099b6e14647994a0b9c1b/src/OpenTelemetry.Instrumentation.GrpcNetClient/GrpcClientInstrumentationOptions.cs#L48) so they don't see 2 spans for every call. For that reason maybe we should go with `ActivityKind.Client` for these?

What does this mean? Depending on whether the option is enabled or not, the SpanKind will be different for gRPC client???

vishweshbankwar: If I am interpreting the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind:~:text=Note%3A%20there%20are,the%20relevant%20libraries.) correctly, CLIENT span parent of another CLIENT span is a possibility. 

RPC spec does call out the relationship between RPC and Http span here https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/rpc-spans.md#distinction-from-http-spans but does not specifically talks about `SPANKIND`

@lmolkova may have more context around this.

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -813,9 +813,12 @@ private void SetFailedResult(Status status)
         // 3. There is an existing activity (to enable activity propagation)
         if (diagnosticSourceEnabled || Logger.IsEnabled(LogLevel.Critical) || Activity.Current != null)
```
noahfalk, line 814: The story is unfortunately a bit messy, but I'd suggest using the same logic that HttpClient uses to create the Activity:
https://source.dot.net/#System.Net.Http/System/Net/Http/DiagnosticsHandler.cs,48
https://source.dot.net/#System.Net.Http/System/Net/Http/DiagnosticsHandler.cs,56
I think the important difference is the fallback to use Activity constructor. ActivitySource.CreateActivity() will return a null Activity if no ActivityListener requests it, which would break cases 2,3, and 4 if there is no fallback.

There are probably four different scenarios that are interesting:
1. Folks using the new ActivityListener APIs want to be able to turn on the Activity just by subscribing via ActivityListener. As-is this would be insufficient because ActivityListener doesn't necessarily cause any of the three conditions the code checks for to be true.
2. Folks using the DiagnosticListener APIs to get notified. For newer tools hopefully the only use for this is if they want to populate custom tags on the Activity by using the ActivityStartData and ActivityStopData that is passed in the WriteDiagnosticEvent calls. Eventually I hope we can deprecate the check for DiagnosticListener.IsEnabled(GrpcDiagnostics.ActivityName), but we'd want to give plenty of time for tools to adopt the ActivityListener mechanism first.
3. ILogger can log Activity trace information (the Logger.IsEnabled(LogLevel.Critical) check). I hope in the future we can stop the practice of implicitly driving distributed tracing configuration via logging configuration, but we probably need something like https://github.com/dotnet/runtime/pull/90201 for distributed tracing first and we aren't there yet.
4. Propagating trace information from inbound to outbound networking requests even when none of 1-3 are activated (The Activity.Current != null check). Similar to the logging case, I hope this is one we can avoid the need for this in the future when we have more explicit configuration APIs.

eerhardt: > I think the important difference is the fallback to use Activity constructor.

Would this still allow us to remove the Reflection usage in OpenTelemetry? I think we should eliminate this Reflection usage in OpenTelemetry when running on the latest bits.

https://github.com/open-telemetry/opentelemetry-dotnet/pull/4695#discussion_r1276692304

My understanding is the reason they are using Reflection today is because libraries are constructing the Activity through a constructor, and the Activity.Source and Activity.Kind properties are not being set (and they can't be set publicly - thus Reflection).

src/Grpc.Net.Client/Internal/GrpcCall.cs
alanwest: Agree that if Grpc.Net.Client is starting an activity then it should use `ActivityKind.Client`.

While it does seem odd for the parent of a client span to also be a client span, as @vishweshbankwar points out there's nothing in either the HTTP or gRPC conventions that preclude this today. Currently, both conventions state the spans MUST be client spans.

In probably our first discussion on this topic, some felt it would be optimal if only one span was emitted. Is it an option to _not_ start a span from Grpc.Net.Client and instead just add gRPC attributes to the underlying HttpClient span? I suspect the answer is no because if I recall the HTTP client span is not necessarily representative of the duration of the gRPC call https://github.com/open-telemetry/opentelemetry-dotnet/pull/687#discussion_r428856938.

src/Grpc.Net.Client/Internal/GrpcCall.cs
JamesNK: Yes. Noah went into detail: https://github.com/grpc/grpc-dotnet/pull/2244#discussion_r1295028120

src/Grpc.Net.Client/Internal/GrpcCall.cs
JamesNK: Grpc needs its own span, at least in .NET. The span that HttpClient creates ends when the HTTP response headers are returned.

That means:
* The span won't report time spent streaming messages after that point
* The span can't be used to report the overall status of the gRPC call in a tag. That is specified in a response trailer which arrives after the span is stopped.

JamesNK: It seems strictly worse than what we're currently doing. I'll leave it as is.

JamesNK:
I think all PR feedback is applied.

src/Grpc.Net.Client/Internal/GrpcCall.cs
lmolkova: gRPC semconv [requires](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/rpc-spans.md?plain=1#L58) RPC span to have client kind and it's ok to have client children of client spans. 

Client children of other client spans create some complications for Azure Monitor and maybe other backends, but it's not a new problem or not something that can be solved in gRPC instrumentation.

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -811,9 +811,15 @@ private void SetFailedResult(Status status)
...
         // 3. There is an existing activity (to enable activity propagation)
-        if (diagnosticSourceEnabled || Logger.IsEnabled(LogLevel.Critical) || Activity.Current != null)
+        // 4. ActivitySource has listeners
+        if (diagnosticSourceEnabled || Logger.IsEnabled(LogLevel.Critical) || Activity.Current != null || GrpcDiagnostics.ActivitySource.HasListeners())
         {
-            activity = new Activity(GrpcDiagnostics.ActivityName);
+            activity = GrpcDiagnostics.ActivitySource.CreateActivity(GrpcDiagnostics.ActivityName, ActivityKind.Client);
+
+            // ActivitySource only returns an activity if someone is listening.
+            // We always want there to be an activity so fallback to creating the activity manually.
```
JamesNK, line 820:
```suggestion
            // If we're at this point then we always want there to be an activity. Create the activity manually.
```

amcasey (approved):
Seems sensible (though I was a little surprised ActivitySource and deprecating .net 5 were in the same PR)

JamesNK:
> Seems sensible (though I was a little surprised ActivitySource and deprecating .net 5 were in the same PR)

.NET 5 was deprecated here https://github.com/grpc/grpc-dotnet/pull/2021. This PR cleans up some left over files.

test/Grpc.Net.Client.Tests/DiagnosticsTests.cs
```diff
@@ -176,6 +176,48 @@ public void DiagnosticListener_MakeCall_ActivityHasNameAndDuration()
         Assert.AreNotEqual(TimeSpan.Zero, activityDurationOnStop);
     }
 
+    [Test]
+    public void ActivitySource_MakeCall_ActivityHasNameAndDuration()
```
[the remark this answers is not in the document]
JamesNK: Yes

src/Grpc.Net.Client/Internal/GrpcCall.cs
```diff
@@ -811,9 +811,15 @@ private void SetFailedResult(Status status)
...
-        if (diagnosticSourceEnabled || Logger.IsEnabled(LogLevel.Critical) || Activity.Current != null)
+        // 4. ActivitySource has listeners
+        if (diagnosticSourceEnabled || Logger.IsEnabled(LogLevel.Critical) || Activity.Current != null || GrpcDiagnostics.ActivitySource.HasListeners())
         {
-            activity = new Activity(GrpcDiagnostics.ActivityName);
+            activity = GrpcDiagnostics.ActivitySource.CreateActivity(GrpcDiagnostics.ActivityName, ActivityKind.Client);
+
+            // ActivitySource only returns an activity if someone is listening.
+            // If we're at this point then we always want there to be an activity. Create the activity manually.
+            activity ??= new Activity(GrpcDiagnostics.ActivityName);
```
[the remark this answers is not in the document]
JamesNK: The default is internal, but it can't be set when using the `Activity` ctor.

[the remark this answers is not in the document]
tarekgh: We can consider in the future release to add a constructor that can take a `ActivityKind`. I think this can help in such scenarios but still will not help older Grpc client versions that have already been released.

PR #2245, opened 2023-08-15
Force yielding after awaiting CallTask to avoid holding onto locks

JamesNK:
Extra fix for https://github.com/grpc/grpc-dotnet/issues/2206

### DIFF
--- a/src/Shared/CompatibilityHelpers.cs
+++ b/src/Shared/CompatibilityHelpers.cs
@@ -66,4 +66,26 @@ internal static class CompatibilityHelpers
         return cancellationToken.Register((state) => callback(state, cancellationToken), state);
 #endif
     }
+
+    public static Task<T> AwaitWithYieldAsync<T>(Task<T> callTask)
+    {
+        // A completed task doesn't need to yield because code after it isn't run in a continuation.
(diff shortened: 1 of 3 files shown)


src/Shared/CompatibilityHelpers.cs
```diff
@@ -66,4 +66,26 @@ public static CancellationTokenRegistration RegisterWithCancellationTokenCallbac
         return cancellationToken.Register((state) => callback(state, cancellationToken), state);
 #endif
     }
+
+    public static Task<T> AwaitWithYieldAsync<T>(Task<T> callTask)
+    {
+        // A completed task doesn't need to yield because code after it isn't run in a continuation.
+        if (callTask.IsCompletedSuccessfully())
```
ReubenBond, line 73: Why not `IsCompleted`?

ReubenBond (approved):

src/Shared/CompatibilityHelpers.cs
JamesNK: Good point. The particular task we want to force this behavior on `GrpcCall.CallTask` will never throw, but this code ended up being split out into a helper method that could technically be used with any task.

I'll change it.

PR #2251, opened 2023-08-22
Update version to 2.56.0 (on v2.56.x branch)

jtattermusch:
I'll publish the release on/after Aug 23 if everything looks good.

### DIFF
--- a/build/version.props
+++ b/build/version.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
 
     <!-- package version of grpc-dotnet -->
-    <GrpcDotnetVersion>2.56.0-pre2</GrpcDotnetVersion>
+    <GrpcDotnetVersion>2.56.0</GrpcDotnetVersion>
     
     <!-- assembly version of grpc-dotnet -->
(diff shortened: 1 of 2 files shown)


JamesNK (approved):

jtattermusch:
Nugets pushed, release published: https://github.com/grpc/grpc-dotnet/releases/tag/v2.56.0

tomkerkhove:
It might be beneficial to update the release with an overview of changes since last stable version and not preview as it currently is a bit vague:
image

jtattermusch:
> It might be beneficial to update the release with an overview of changes since last stable version and not preview as it currently is a bit vague: image

Fair point, but we'd have to update all the past releases as well for consistency, since this is the way we'd been structuring the release notes for a quite a long time now (basically, the full changelog of the stable release is the changelog since the last pre-release AND all the release notes of all the previous pre-releases).

tomkerkhove:
That's OK - But what about upcoming releases? Can we please start using a more end-user friendly way? Otherwise you have to go through every single preview that was done to understand what changed which adds a lot of cognitive load

PR #2255, opened 2023-08-25
Fix unobserved exceptions with retries

JamesNK:
This PR stops unobserved exceptions in the client when using the retry feature. It's similar to https://github.com/grpc/grpc-dotnet/pull/1660, except this PR is for calls with retry enabled.

* Suppress `ResponseAsync` exception if `ResponseHeadersAsync` is called. If someone accesses headers with `ResponseHeadersAsync` then they're unlike to call `ResponseAsync` to see the same error.
* Always suppress exceptions when a call ends with a `Cancel` or `Deadline` status. This is consistent with the automatic suppression of `OperationCanceledException`.
* Update most tests in Grpc.Net.Client to always pass a `Method<TRequest, TResponse>` with the right type. This is important because retry/hedging calls look at the method type.

Feel free to ignore the test changes.


src/Grpc.Net.Client/Internal/HttpClientCallInvoker.cs
```diff
@@ -127,6 +135,17 @@ public HttpClientCallInvoker(GrpcChannel channel)
         return callWrapper;
     }
 
+    private static void AssertMethodType(IMethod method, MethodType methodType)
```
amcasey, line 138: Is `ConditionalAttribute` applicable here?

src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
```diff
@@ -111,13 +113,34 @@ protected RetryCallBase(GrpcChannel channel, Method<TRequest, TResponse> method,
...
+            _responseHeadersTask = GetResponseHeadersCoreAsync();
+
+            // ResponseHeadersAsync could be called inside a client interceptor when a call is wrapped.
+            // Most people won't use the headers result. Observed exception to avoid unobserved exception event.
+            _responseHeadersTask.ObserveException();
+
+            // If there was an error fetching response headers then it's likely the same error is reported
+            // by response TCS. The user is unlikely to observe both errors.
+            // Observed exception to avoid unobserved exception event.
+            _responseTask?.ObserveException();
```
amcasey, line 137: The comment makes it seem like the desirability of this behavior depends on the outcome of the header task, but that's not obviously reflected in the code.  Will this observe/suppress the response exception if there's no header exception?

amcasey (approved):
Seems reasonable.

PR #2256, opened 2023-08-25
Reduce logger allocations by not using generic CreateLogger

JamesNK:
Non-generic `CreateLogger` caches logger instances on the log name. Meanwhile, `CreateLogger<T>` always creates a new instance: https://github.com/dotnet/runtime/blob/ea2022818b117bead26fc19c8ebb47b455907cab/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/LoggerFactoryExtensions.cs#L20-L25

These loggers last the lifetime of a channel, so the cost of the generic create is low. But this saves a few allocations per channel for apps with multiple channels.


amcasey (approved):
I wonder if it would be worthwhile to add a comment so that no one comes along and "fixes" it by switching back to the generic version.

PR #2258, opened 2023-08-25
Log socket lifetime when closing unusable sockets

JamesNK:
Log the lifetime to make it easier to debug if sockets are unexpectedly being closed early.

Small breaking change to rename `ClosingUnusableSocketOnCreateStream` to `ClosingUnusableSocket`. I did this because it is now used in two places where a socket is closed. However, this log is only a few months old. I think it's unlikely anyone would explicitly care about the name.


amcasey (approved):
I don't know how likely the breaking change is to affect consumers, but I would also be fine with introducing a substantially similar message for the second call-site (CloseUnusableSocketOnCheck or something).

PR #2261, opened 2023-08-25
Add transport status to subchannel picked log

JamesNK:
Help debug call issues by reporting information about the subchannel transport status when a subchannel is picked.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
+++ b/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
@@ -96,7 +96,7 @@ internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDi
                 {
                     return TransportStatus.ActiveStream;
                 }
-                return TransportStatus.Unknown;
+                return TransportStatus.NotConnected;
             }
         }
     }
--- a/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
+++ b/src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
@@ -82,6 +82,24 @@ internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDi
     private object Lock => _subchannel.Lock;
     public BalancerAddress? CurrentAddress => _currentAddress;
     public TimeSpan? ConnectTimeout { get; }
+    public TransportStatus TransportStatus
+    {
+        get
+        {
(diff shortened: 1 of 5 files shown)


src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -82,6 +82,24 @@ internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDi
     private object Lock => _subchannel.Lock;
     public BalancerAddress? CurrentAddress => _currentAddress;
     public TimeSpan? ConnectTimeout { get; }
+    public TransportStatus TransportStatus
+    {
+        get
+        {
+            lock (Lock)
+            {
+                if (_initialSocket != null)
```
amcasey, line 91: Without knowing anything about this type, I would have guessed that having active streams was more specific than having an initial(ized?) socket and, therefore, would have assumed that that status would take precedence.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
```diff
@@ -82,6 +82,24 @@ internal class SocketConnectivitySubchannelTransport : ISubchannelTransport, IDi
...
+                {
+                    return TransportStatus.InitialSocket;
+                }
+                if (_activeStreams.Count > 0)
+                {
+                    return TransportStatus.ActiveStream;
+                }
+                return TransportStatus.Unknown;
```
amcasey, line 99: Is it actually unknown or does this mean uninitialized?

amcasey (approved):
The meanings of the enum values aren't very clear to an outsider.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
JamesNK: Both will never be true. Order doesn't matter.

src/Grpc.Net.Client/Balancer/Internal/SocketConnectivitySubchannelTransport.cs
JamesNK: I renamed it to `NotConnected`. There are a bunch of reasons to be in that state. `NotConnected` name fits them all.

amcasey: Much clearer.  Thanks!

PR #2265, opened 2023-08-30
Fix connection bugs from BalancerAddress changes

JamesNK:
Fixes connection bugs from `BalancerAddress` changes in https://github.com/grpc/grpc-dotnet/pull/2228:

* Round-robin load balancing was unnecessarily reconnecting after updating subchannel addresses. Null and not null check was mixed up.
* Pick first load balancing wasn't reconnecting after updating addresses. Need to get `CurrentAddress` before updating addresses.
* Fixes flaky `HasSubchannels_ResolverRefresh_MatchingSubchannelUnchanged` test.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/SubchannelsLoadBalancer.cs
+++ b/src/Grpc.Net.Client/Balancer/SubchannelsLoadBalancer.cs
@@ -17,6 +17,7 @@
 #endregion
 
 #if SUPPORT_LOAD_BALANCING
+using System.Diagnostics;
 using Grpc.Core;
 using Grpc.Net.Client.Balancer.Internal;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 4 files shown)


src/Grpc.Net.Client/Balancer/SubchannelsLoadBalancer.cs
```diff
@@ -144,8 +145,11 @@ public override void UpdateChannelState(ChannelState state)
                 // The new subchannel address has the same endpoint so the connection isn't impacted.
                 if (!BalancerAddressEqualityComparer.Instance.Equals(address, newOrCurrentSubchannel.Address))
                 {
+                    newOrCurrentSubchannel = new AddressSubchannel(
```
amcasey, line 148: Another way to express this might be to expose a `WithAddress` method on `AddressSubchannel` that returns a new object differing only in the address.

amcasey (approved):

PR #2271, opened 2023-09-05
Add request and channel to call debugging

JamesNK:
This adds the channel and request (for unary and server streaming calls) to gRPC debugging.

This change is tricky because types are split across projects and don't have access to each other's internal views. The hack workaround is to use `IEnumerable` to return values. Nothing bad is exposed to the user and if another channel implementation doesn't implement enumerable and return values with the expected keys then the debugging information is just missing.

image

### DIFF
--- a/src/Grpc.Core.Api/Internal/CallDebuggerHelpers.cs
+++ b/src/Grpc.Core.Api/Internal/CallDebuggerHelpers.cs
@@ -17,18 +17,27 @@
 #endregion
 
 using System;
+using System.Collections;
 using System.Diagnostics;
 
 namespace Grpc.Core.Internal;
 
 internal static class CallDebuggerHelpers
 {
+    public const string MethodKey = "Method";
+    public const string ChannelKey = "Channel";
+    public const string RequestKey = "Request";
+
     public static string DebuggerToString(AsyncCallState callState)
     {
         string debugText = string.Empty;
-        if (callState.State is IMethod method)
+        if (GetDebugValue<ChannelBase>(callState, ChannelKey) is { } channel)
         {
-            debugText = $"Method = {method.FullName}, ";
+            debugText = $"Channel = {channel.Target}, ";
+        }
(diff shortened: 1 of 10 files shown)


src/Grpc.Core.Api/Internal/CallDebuggerHelpers.cs
```diff
@@ -40,6 +50,25 @@ public static string DebuggerToString(AsyncCallState callState)
         return debugText;
     }
 
+    public static T? GetDebugValue<T>(AsyncCallState callState, string key) where T : class
```
amcasey, line 53: Why does `T` have to be a class?

src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
```diff
@@ -83,6 +84,20 @@ internal static string GetMessageAcceptEncoding(Dictionary<string, ICompressionP
 #endif
     }
 
+    public static IEnumerator<KeyValuePair<string, object>> GetDebugEnumerator(ChannelBase channel, IMethod method, object? request)
```
amcasey, line 87: I feel like we've had this conversation before, but I've forgotten the output.  Why a list of pairs rather than a dictionary?

amcasey (approved):

src/Grpc.Net.Client/Internal/GrpcProtocolConstants.cs
JamesNK: It's easier to implement `IEnumerable` (one method) on the state than a dictionary (lots of methods). I'll add a comment explaining the choice.

src/Grpc.Core.Api/Internal/CallDebuggerHelpers.cs
JamesNK: All the values so far as classes. It is easier to work with nullable values when constrained to class.

PR #2273, opened 2023-09-06
Add ProtoStatus example

JamesNK:
Example is under `examples/Error`


JamesNK:
It feels ok. I still need to check out what is happening with exception stacks.

My main feedback is there are a lot of namespaces when everything is used together. For example, on the server:

```cs
using Google.Protobuf.WellKnownTypes;
using Google.Rpc;
using Greet;
using Grpc.Core;
using Grpc.StatusProto;
```

Move the extension methods from `Grpc.StatusProto` to `Grpc.Core` or `Google.Rpc`?

tonydnewell (approved):

PR #2274, opened 2023-09-11
Support disabling load balancing with SocketsHttpHandler.Properties

JamesNK:
Allows someone to configure `SocketsHttpHandler` without using gRPC load balancing.

The setting uses the object bag `SocketsHttpHandler.Properties`. The UI for configuring this value is ugly but it isn't intended for general use.

### DIFF
--- a/src/Grpc.Net.Client/GrpcChannel.cs
+++ b/src/Grpc.Net.Client/GrpcChannel.cs
@@ -266,16 +266,8 @@ public sealed class GrpcChannel : ChannelBase, IDisposable
         }
         if (HttpRequestHelpers.HasHttpHandlerType(channelOptions.HttpHandler, "System.Net.Http.SocketsHttpHandler"))
         {
-            HttpHandlerType type;
-            TimeSpan? connectTimeout;
-            TimeSpan? connectionIdleTimeout;
-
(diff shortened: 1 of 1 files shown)


davidfowl (approved):

src/Grpc.Net.Client/GrpcChannel.cs
```diff
@@ -292,6 +292,15 @@ private static HttpHandlerContext CalculateHandlerContext(ILogger logger, Uri ad
                 }
             }
 
+            // Load balancing has been disabled on the SocketsHttpHandler.
+            if (socketsHttpHandler.Properties.TryGetValue("__GrpcLoadBalancingDisabled", out var value)
+                && value is bool loadBalancingDisabled && loadBalancingDisabled)
+            {
+                type = HttpHandlerType.Custom;
+                connectTimeout = null;
+                connectionIdleTimeout = null;
```
davidfowl, line 301: What does this mean? That they won't be overridden on the handler?

JamesNK: They have no meaning if the handler type isn't `SocketsHttpHandler`.

I've simplified the method to avoid doing this.

Issue #2278, opened 2023-09-19
Update targetframeworks for all Grpc.* packages


jskeet:
As agreed in internal discussions, suggested new TFMs are:

- netstandard2.0
- net462
- net6.0 if it provides benefits

jtattermusch:
My only concern is the Grpc.Core.Api package (which is supposed to continue working with the legacy Grpc.Core releases).
What would we gain by updating the targetframeworks for Grpc.Core.Api (as opposed to just updating Grpc.Auth which is the original reason why this is requested) and doesn't the risk outweigh the gains? I'm happy to be convinced that this change is safe, but I really don't want to break Grpc.Core users with this optional change.

CC @apolcyn @tonydnewell

https://github.com/grpc/grpc-dotnet/blob/40a26c9877ed3a96e853fc889ea7dcb30d4f99eb/src/Grpc.Core.Api/Grpc.Core.Api.csproj#L8

jskeet:
It would probably be worth evaluating what situations would be broken by this. By my reckoning, which could be wrong, someone would have to:

- Have dependencies on both Grpc.Core and Grpc.Core.Api with different versions (if they've only got Grpc.Core, then that would presumably use the Grpc.Core.Api version that Grpc.Core depends on)
- Be targeting a long-unsupported platform (e.g. net45 or .NET Core 1.x)

I think it's reasonable to not support that scenario - but if there's a scenario which we think we *should* still support which would be broken by this, then sure, just changing Grpc.Auth would still unblock the Google client libraries team.

jtattermusch:
> It would probably be worth evaluating what situations would be broken by this. By my reckoning, which could be wrong, someone would have to:
> 
> * Have dependencies on both Grpc.Core and Grpc.Core.Api with different versions (if they've only got Grpc.Core, then that would presumably use the Grpc.Core.Api version that Grpc.Core depends on)

I think a reasonable scenario is to have the latest Grpc.Core.Api (e.g. because there was a backward-compatible fix in the API or some of the helpers contained in Grpc.Core.Api and you want to benefit from having a patched version of Grpc.Core.Api) and the latest version of Grpc.Core (which always starts with 2.46.x these days as there are no newer versions of it). So you can have Grpc.Core.Api 2.57.0 (Grpc.Core.Api still gets new releases) together with Grpc.Core 2.46.6 and they should still work together well.

> * Be targeting a long-unsupported platform (e.g. net45 or .NET Core 1.x)
> 
[...]

PR #2279, opened 2023-09-20
Remove netstandard1.5 from some projects

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2278

cc @jskeet

### DIFF
--- a/src/Grpc.Core.Api/Grpc.Core.Api.csproj
+++ b/src/Grpc.Core.Api/Grpc.Core.Api.csproj
@@ -5,7 +5,7 @@
 
     <IsGrpcPublishedPackage>true</IsGrpcPublishedPackage>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <TargetFrameworks>net462;netstandard1.5;netstandard2.0;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>net462;netstandard2.0;netstandard2.1</TargetFrameworks>
     <PackageReadmeFile>README.md</PackageReadmeFile>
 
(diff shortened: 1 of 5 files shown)


jskeet (approved):

jtattermusch:
LGTM once my concern from https://github.com/grpc/grpc-dotnet/issues/2278#issuecomment-1727133647 is addressed.

src/Grpc.Core.Api/Grpc.Core.Api.csproj
```diff
@@ -5,7 +5,7 @@
 
     <IsGrpcPublishedPackage>true</IsGrpcPublishedPackage>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <TargetFrameworks>net462;netstandard1.5;netstandard2.0;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>net462;netstandard2.0;netstandard2.1</TargetFrameworks>
```
jtattermusch, line 8: FYI, `netstandard2.1` isn't listed in the recommended list of frameworks in https://github.com/grpc/grpc-dotnet/issues/2278 (which makes this PR only a partial fix), so I'm not sure if not updating it was intentional or not.

jskeet: That may just be my fault when reporting the issue. I'm personally okay with it supporting netstandard2.1 as well, as that doesn't add any burden on dependencies (which need to support netstandard2.0 anyway).

JamesNK:
@jtattermusch Approve please

jtattermusch (approved):

PR #2294, opened 2023-10-13
Fix HttpContext race condition by copying values to reader and writer

JamesNK:
Potential fix to `HttpContext.RequestAborted` race condition. Fix is to copy fields need from the context when reader and writer are created.

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/HttpContextStreamReader.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/HttpContextStreamReader.cs
@@ -17,6 +17,7 @@
 #endregion
 
 using System.Diagnostics;
+using System.IO.Pipelines;
 using Grpc.Core;
 using Grpc.Shared;
 
@@ -28,6 +29,8 @@ internal class HttpContextStreamReader<TRequest> : IAsyncStreamReader<TRequest> 
 {
     private readonly HttpContextServerCallContext _serverCallContext;
     private readonly Func<DeserializationContext, TRequest> _deserializer;
+    private readonly PipeReader _bodyReader;
+    private readonly CancellationToken _cancellationToken;
     private bool _completed;
     private long _readCount;
     private bool _endOfStream;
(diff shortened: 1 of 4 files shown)


src/Grpc.AspNetCore.Server/Internal/HttpContextStreamReader.cs
```diff
@@ -36,6 +40,12 @@ public HttpContextStreamReader(HttpContextServerCallContext serverCallContext, F
     {
         _serverCallContext = serverCallContext;
         _deserializer = deserializer;
+
+        // Copy HttpContext values.
+        // This is done to avoid a race condition when reading them from HttpContext later when running in a separate thread.
+        _bodyReader = _serverCallContext.HttpContext.Request.BodyReader;
+        // Copy lifetime feature because HttpContext.RequestAborted on .NET 6 doesn't return the real cancellation token.
```
BrennanConroy, line 47: Could you explain both the 6.0 comment as well as this whole PR?

From my understanding; If you're using the cancellation token on a background thread it might be disposed instead of canceled when the HttpContext is reused. And if you access the token property after it's been reset you might be using a token from another request.

JamesNK: Root problem: https://github.com/dotnet/aspnetcore/issues/42040
The fix it to access the cancellation token on the feature directly. `HttpContext.RequestAborted` isn't thread-safe but the feature is.

re: .NET 6. I observed that the cancellation token returned by `HttpContext.RequestAborted` behaved strangely. It wouldn't immediately be canceled when the request was aborted. The previous `HttpContext.RequestAborted` value appeared to be different from the value after abort happened (but while the request was still in progress). I didn't investigate why since there was a workaround and .NET 7 and .NET 8 behaved how I expected.

BrennanConroy: > Root problem: [dotnet/aspnetcore#42040](https://github.com/dotnet/aspnetcore/issues/42040)
> The fix it to access the cancellation token on the feature directly. `HttpContext.RequestAborted` isn't thread-safe but the feature is.

Oh, so gRPC isn't using the token after the middleware returns? That's what my comment was focused on.

JamesNK: No. This is in the context of one request.

BrennanConroy (approved):

PR #2321, opened 2023-11-16
Update to GA .NET 8 SDK

JamesNK:
Update various things to use the latest bits from the .NET 8 release.

### DIFF
--- a/build/get-dotnet.sh
+++ b/build/get-dotnet.sh
@@ -18,7 +18,7 @@ ensure_dir() {
 # main
 
 # resolve SDK version
-sdk_version="8.0.100-rc.1.23407.2"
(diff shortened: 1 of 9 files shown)


build/get-dotnet.sh
```diff
@@ -18,7 +18,7 @@ ensure_dir() {
 # main
 
 # resolve SDK version
-sdk_version="8.0.100-rc.1.23407.2"
```
amcasey, line 21: I don't know what this does.

amcasey (approved):

build/get-dotnet.sh
JamesNK: It reads the SDK version from global.json

PR #2326, opened 2023-11-21
Trim regex from Grpc.AspNetCore

JamesNK:
`AddGrpc` calls `AddRouting` to configure a route constraint. This is a problem when publishing with trimming/AOT because `AddRouting` registers the regex route constraint.

Configure `RouteOptions` directly and rely on the host to add routing. If the host is created with slim builder, then regex is never referenced and can be trimmed away.

I confirmed regex is trimmed after this change with sizoscope:

image


JamesNK:
Ping for review now that holidays are over.

The changes are in [GrpcServiceExtensions.cs](https://github.com/grpc/grpc-dotnet/pull/2326/files#diff-2f18d13e93b07e61486fe7987c22dd2a99547a9889dd92d7e8fc1cda54635438). Everything else is for testing.

BrennanConroy (approved):
Technically a breaking change. But not sure if it's worth calling out, or if grpc has a way to call out breaking changes if it is worth it.

JamesNK:
I don't think it is worth calling out.

Regex route would never be used with gRPC (it defines routes for you) and if you have regex in routes in another framework, e.g. MVC, Minimal API, then that framework or hosting will already be calling `AddRouting` to add the regex route.

Issue #2304, opened 2023-10-20
OperationCanceledException thrown when canceling a CancelationToken does not reference the token canceled when using Retries


Watno:
### What version of gRPC and what language are you using?
Grpc.Net.Client 2.58.0, C#

### What operating system (Linux, Windows,...) and version?
Windows 11

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
SDK 7.0.402

### What did you do?
The situation is similar to #2014: OperationCanceledException thrown when canceling a CancelationToken does not reference the token canceled
When having Retries enabled in the GrpcClient, (i.e. when specifying a `RetryPolicy` with `MaxRetryAttemps > 0` in the ClientConfiguration), the client is called and then the CancelationToken passed is cancelled before the gRPC call completes.

### What did you expect to see?
An `OperationCanceledException` with the CancelationToken passed.

### What did you see instead?
An `OperationCanceledException` with a different CancelationToken than the one passed.
Stacktrace of the exception:
```
 ---> System.OperationCanceledException: The operation was canceled.
   at System.Threading.CancellationToken.ThrowOperationCanceledException()
   at Grpc.Net.Client.Internal.Retry.RetryCall`2.StartRetry(Action`1 startCallFunc)
   --- End of inner exception stack trace ---
   at Grpc.Net.Client.Internal.Retry.RetryCallBase`2.GetResponseCoreAsync()```

PR #2345, opened 2024-01-02
Fix cancellation token reported when using retries

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2304

### DIFF
--- a/src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
+++ b/src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
@@ -549,6 +549,7 @@ internal abstract partial class RetryCallBase<TRequest, TResponse> : IGrpcCall<T
                 }
                 else
                 {
+                    // Replace the OCE from CancellationTokenSource with an OCE that has the passed in cancellation token if it is canceled.
                     if (Options.CancellationToken.IsCancellationRequested && Options.CancellationToken != operationCanceledException.CancellationToken)
                     {
                         ex = new OperationCanceledException(Options.CancellationToken);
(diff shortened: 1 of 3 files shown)


src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
```diff
@@ -542,7 +542,20 @@ protected void HandleUnexpectedError(Exception ex)
...
                 commitReason = CommitReason.Canceled;
-                resolvedCall = CreateStatusCall(Disposed ? GrpcProtocolConstants.CreateDisposeCanceledStatus(ex) : GrpcProtocolConstants.CreateClientCanceledStatus(ex));
+                Status status;
+                if (Disposed)
+                {
+                    status = GrpcProtocolConstants.CreateDisposeCanceledStatus(exception: null);
+                }
+                else
+                {
+                    if (Options.CancellationToken.IsCancellationRequested && Options.CancellationToken != operationCanceledException.CancellationToken)
```
mgravell, line 552: I just want to check the intent here... "if something cancelled (`ex`), and `Options` is *also* cancelled, but isn't the thing being reported by `ex`: replace `ex` with the cancellation of `Options`" - is that right? are we happy to simply lose the original `ex` (which is a *different* calcellation)? should it be an inner-exception?

mgravell (approved):
discussion point added, but looks good

src/Grpc.Net.Client/Internal/Retry/RetryCallBase.cs
JamesNK: This is replacing the OCE from CancellationTokenSource with an OCE using Options.CancellationToken if it is canceled. I added a comment.

The original exception is from an internal CT so it isn't valuable to keep. An OCE nested inside an OCE would be confusing.

PR #2365, opened 2024-01-30
Add .NET 9 NuGet feed

JamesNK:
Benchmark environment runs with .NET 9. Add .NET 9 feed so it can resolve 9.0 packages.


amcasey:
How long will 7.0 be required?

amcasey (approved):

JamesNK:
> How long will 7.0 be required?

I’ll remove it when net6 and net7 is out of support.

PR #2366, opened 2024-02-05
Add missing package mapping for net9 feed

JamesNK:
Missed in https://github.com/grpc/grpc-dotnet/pull/2365


amcasey (approved):
What's the default?

JamesNK:
I think NuGet resolves feeds in order from top to bottom.

Issue #2375, opened 2024-02-25
Can't create channel in UWP app in Release mode


michalleptuch:
### What version of gRPC and what language are you using?
Grpc.Net.Client 2.61.0
Grpc.Net.Client.Web 2.61.0

### What operating system (Linux, Windows,...) and version?
Windows 11 23H2 22631.3155

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.NET Native (UWP)

### What did you do?
I used this tutorial to create gRPC channel in UWP app: https://learn.microsoft.com/en-us/aspnet/core/grpc/netstandard. Everything works fine in Debug mode. The problem occurs when my app is installed using MSIX package (and works in Release mode).
``` cs
var channel = GrpcChannel.ForAddress($"http://localhost:{port}", new GrpcChannelOptions
{
  HttpHandler = new GrpcWebHandler(new HttpClientHandler()),
});

return channel.CreateGrpcService<T>();
```

### What did you expect to see?
Working channel

### What did you see instead?
Exception message:
```
Unresolved P/Invoke method 'ntdll.dll!RtlGetVersion' from this method. Please look for this method in build warnings for more details. 
```
and stack trace:
```
at __Interop.ntdll_dll.RtlGetVersion(OSVERSIONINFOEX& versionInfo)
at Grpc.Net.Client.Internal.NtDll.DetectWindowsVersion(Version& version, Boolean& isWindowsServer) in /_/src/Grpc.Net.Client/Internal/NtDll.cs:line 40
at Grpc.Net.Client.Internal.OperatingSystem..ctor() in /_/src/Grpc.Net.Client/Internal/OperatingSystem.cs:line 77
at Grpc.Net.Client.Internal.OperatingSystem..cctor() in /_/src/Grpc.Net.Client/Internal/OperatingSystem.cs:line 34
at System.Runtime.CompilerServices.ClassConstructorRunner.EnsureClassConstructorRun(StaticClassConstructionContext* pContext) in f:\dd\ndp\fxcore\CoreRT\src\System.Private.CoreLib\src\System\Runtime\CompilerServices\ClassConstructorRunner.cs:line 92
```

### Anything else we should know about your project / environment?

The problem occurs in `NtDll.cs` file invoked here: 
https://github.com/grpc/grpc-dotnet/blob/da2038226c0a39edf03cedd6339dba5d5eb03315/src/Grpc.Net.Client/Internal/OperatingSystem.cs#L77
[...]

PR #2376, opened 2024-02-26
Fix OS detection error in UWP

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2375

RtlGetVersion isn't available in UWP. First check if the app is UWP before calling RtlGetVersion.

### DIFF
--- a/src/Grpc.Net.Client/Internal/Native.cs
+++ b/src/Grpc.Net.Client/Internal/Native.cs
@@ -51,6 +51,12 @@ internal static class Native
 
     internal static bool IsUwp(Version version)
     {
+        if (RuntimeInformation.FrameworkDescription.StartsWith(".NET Native", StringComparison.OrdinalIgnoreCase))
+        {
+            return true;
+        }
+
+        // From https://github.com/dotnet/runtime/blob/d752f9a19f2d4bc4559e0e303e9374e4042a916e/src/libraries/Common/tests/TestUtilities/System/PlatformDetection.Windows.cs#L190
         const int Windows8Build = 9200;
         if (version.Build < Windows8Build)
         {
--- a/src/Grpc.Net.Client/Internal/Native.cs
+++ b/src/Grpc.Net.Client/Internal/Native.cs
@@ -49,9 +49,9 @@ internal static class Native
         isWindowsServer = osVersionInfo.ProductType == VER_NT_SERVER;
     }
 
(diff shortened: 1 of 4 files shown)


src/Grpc.Net.Client/Internal/Native.cs
```diff
@@ -46,6 +49,40 @@ internal static void DetectWindowsVersion(out Version version, out bool isWindow
...
+                    return false;
+                case 0:     // ERROR_SUCCESS
+                case 122:   // ERROR_INSUFFICIENT_BUFFER
+                            // Success is actually insufficient buffer as we're really only looking for
+                            // not NO_APPLICATION and we're not actually giving a buffer here. The
+                            // API will always return NO_APPLICATION if we're not running under a
+                            // WinRT process, no matter what size the buffer is.
+                    return true;
+                default:
+                    throw new InvalidOperationException($"Failed to get AppId, result was {result}.");
```
amcasey, line 81:
```suggestion
                    throw new InvalidOperationException($"Failed to get AppModelId, result was {result}.");
```

amcasey (approved):

src/Grpc.Net.Client/Internal/OperatingSystem.cs
```diff
@@ -72,9 +72,9 @@ private OperatingSystem()
         //
         // Get correct Windows version directly from Windows by calling RtlGetVersion.
         // https://www.pinvoke.net/default.aspx/ntdll/RtlGetVersion.html
```
amcasey, line 74: Maybe add a note that RtlGetVersion isn't available in UWP - that check isn't very intuitive otherwise.

src/Grpc.Net.Client/Internal/Native.cs
```diff
@@ -46,6 +49,40 @@ internal static void DetectWindowsVersion(out Version version, out bool isWindow
         isWindowsServer = osVersionInfo.ProductType == VER_NT_SERVER;
     }
 
+    internal static bool IsUwp(Version version)
```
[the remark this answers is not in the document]
JamesNK: I don't like embedding static in a help method. For consistency I'll pass in also `RuntimeInformation.FrameworkDescription` (I got lazy 😄 ).

amcasey (approved):
Still LGTM

JamesNK:
FYI @tonydnewell an update to OS detection

Issue #2380, opened 2024-03-07
Deadlock in balancing implementation


kalduzov:
Hi @JamesNK

We have already set tasks related to balancing. You corrected them and now everything works as it should. We successfully wrote our own balancers and even transferred about **100** services to them. Everything worked perfectly for more than 2 months.

But yesterday we encountered a deadlock in the balancer code. We can't reproduce it right now, but its presence really worries us.

The next 2 lines of code, taking into account blocking, in some situation block each other.

https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.Net.Client/Balancer/Internal/ConnectionManager.cs#L291

https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.Net.Client/Balancer/Internal/ConnectionManager.cs#L368

We use standard classes for our balancers and pickers.

Our balancer works according to the push model, i.e. it receives new data on endpoints, and then we send it to the standard Listener().

At this time, our Picker implementations are also working - they are constantly being edited. Every time new endpoints appear, we recreate the pickers with new endpoints (just like in the examples in the documentation).

In the example, it looks like the PickAsync method simply gets stuck in the WaitAsync method immediately after new endpoints arrive. The service continues to work, but each new request in balancing begins to eat up the thread from the ThreadPool. As a result, the picture looks like this:
1. Calls to a quiet client no longer work (but with us they are rejected by CancellationToken or Deadline)
2. The lock continues to hang indefinitely and block the pool thread.
3. The pool begins to swell because they still try to make requests to clients.
4. And because our services in pods with memory limitations - after a while OOM comes (each thread eats up memory a little) and restarts the pod.
5. After this balancing it works stably.

This is a really rare situation. The discovery happened completely by chance out of about 500 pods that occurred in just 3 months over the past month.

We would really like to somehow solve this problem - it greatly hinders us, because... We don’t understand how it works at all and whether it blocks the extreme API for us at some point.

Attached is a screenshot from the captured trace from the pod
![image](https://github.com/grpc/grpc-dotnet/assets/1679308/6f80e246-7827-4db0-bf5b-52b0ac644958)

Library version: 2.60.0

krasin-ga:
[...]

PR #2385, opened 2024-03-11
Prevent block inside ResolveAsync from blocking PollingResolver.Refresh

JamesNK:
Addresses https://github.com/grpc/grpc-dotnet/issues/2380#issuecomment-1987257515

User code is called inside a lock in the `Refresh` method. If the user code blocks without returning a task then the lock blocks future calls to `Refresh`.

The fix is to ensure user code is async by calling it from `Task.Run`.

### DIFF
--- a/test/Grpc.Net.Client.Tests/Balancer/ResolverTests.cs
+++ b/test/Grpc.Net.Client.Tests/Balancer/ResolverTests.cs
@@ -43,6 +43,71 @@ namespace Grpc.Net.Client.Tests.Balancer;
 public class ResolverTests
 {
     [Test]
+    public async Task Refresh_BlockInsideResolveAsync_ResolverNotBlocked()
+    {
+        var waitHandle = new ManualResetEvent(false);
+
+        var services = new ServiceCollection();
+        var testSink = new TestSink();
+        services.AddLogging(b =>
+        {
+            b.AddProvider(new TestLoggerProvider(testSink));
+        });
+        services.AddNUnitLogger();
+        var loggerFactory = services.BuildServiceProvider().GetRequiredService<ILoggerFactory>();
+
+        var logger = loggerFactory.CreateLogger<ResolverTests>();
+        logger.LogInformation("Starting.");
+
+        var ss = new LockingPollingResolver(loggerFactory, waitHandle);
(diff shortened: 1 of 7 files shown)


krasin-ga:
Looks good. I was also thinking about a scenario like this, but they both have the same root cause.

**First thread**
```
LockingPollingResolver.SomeCallback  🟢 lock (LockingPollingResolver._lock){base.Listener(endpoints)} 
  ConnectionManager.OnResolverResult  🔴 lock(ConnectionManager._lock){LoadBalancer.UpdateChannelState}
   ...
```

**Second thread**
```
SocketConnectivitySubchannelTransport.TryConnectAsync 
 Subchannel.UpdateConnectivityState 
   ConnectionManager.OnSubchannelStateChange 🟢 lock(ConnectionManager._lock){ subchannel.RaiseStateChanged(state, status);} 
      ChildHandlerController.RefreshResolver() 
       PollingResolver.Refresh() 
        PollingResolver.ResolveNowAsync() 
          LockingPollingResolver.ResolveAsync 🔴 lock(CustomPollingResolver._lock)base.Listener()} 
            ConnectionManager.OnResolverResult 
              lock(ConnectionManager._lock){LoadBalancer.UpdateChannelState} 
```

test/Grpc.Net.Client.Tests/Balancer/ResolverTests.cs
```diff
@@ -42,6 +42,71 @@ namespace Grpc.Net.Client.Tests.Balancer;
...
+        {
+            b.AddProvider(new TestLoggerProvider(testSink));
+        });
+        services.AddNUnitLogger();
+        var loggerFactory = services.BuildServiceProvider().GetRequiredService<ILoggerFactory>();
+
+        var logger = loggerFactory.CreateLogger<ResolverTests>();
+        logger.LogInformation("Starting.");
+
+        var ss = new LockingPollingResolver(loggerFactory, waitHandle);
```
krasin-ga, line 62:
```suggestion
        var lockingResolver = new LockingPollingResolver(loggerFactory, waitHandle);
```

mgravell (approved):

PR #2409, opened 2024-04-12
Status code can be obtained from servers that do not support Trailers-Only response

castaneai:
We found that RetryPolicy did not work for some servers. Upon further investigation, we found that `Grpc.Net.Client.Internal.GrpcCall.ValidateHeaders` used for grpc-dotnet retries rely on a response format called "Trailers-Only".
However, the server did not support the "Trailers-Only" response format, and grpc-dotnet retries did not work.

This patch solves the problem of retries not working for servers that do not return "Trailers-Only" responses (e.g., [Connect-Go](https://github.com/connectrpc/connect-go)).

ref: https://github.com/connectrpc/connect-go/issues/724

### DIFF
--- a/src/Grpc.Net.Client/Internal/GrpcCall.NonGeneric.cs
+++ b/src/Grpc.Net.Client/Internal/GrpcCall.NonGeneric.cs
@@ -147,14 +147,17 @@ internal abstract class GrpcCall
 
     internal static Status? ValidateHeaders(HttpResponseMessage httpResponse, out Metadata? trailers)
     {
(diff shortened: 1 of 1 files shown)


JamesNK:
It's only valid to retry Trailers-Only responses according to the spec. See https://github.com/grpc/proposal/blob/master/A6-client-retries.md#when-retries-are-valid

Also, the changes you've made won't work. The response body stream must be read to the end (even if it is empty) before trailers are available. This is why Trailers-Only are required.

castaneai:
@JamesNK 
Thank you! I have checked the Client-side retry specifications.

PR #2415, opened 2024-04-17
Fix ObjectDisposedException message

drewnoakes:
The call to `ObjectDisposedThrowHelper.ThrowIf` expects an instance or a type. This code was passing a name, which resulted in an error message resembling:

> System.ObjectDisposedException : Cannot access a disposed object.
> Object name: 'System.String'.

It should mention the `GrpcChannel` type, not the string type.


drewnoakes:
@JamesNK do you know who is the CLA Manager for Microsoft. I can't sign the CLA without providing a contact.

JamesNK:
I don't know.

drewnoakes:
I sorted out the CLA. Looks like the failure is in package restore which seems unrelated to the change here.

JamesNK:
Fixed build. Rebase to fix this PR

JamesNK (approved):

PR #2416, opened 2024-04-18
Enable multiple connections with WinHttpHandler by default

JamesNK:
Feedback from a customer about streams pausing when a limit is reached. Can help this scenario by defaulting the WinHttpHandler to create more streams if required.

### DIFF
--- a/src/Shared/HttpHandlerFactory.cs
+++ b/src/Shared/HttpHandlerFactory.cs
@@ -38,7 +38,7 @@ internal static class HttpHandlerFactory
 #endif
 
 #if NET462
-        // Create SocketsHttpHandler with EnableMultipleHttp2Connections set to true. That will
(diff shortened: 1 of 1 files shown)


amcasey:
Does this increase the risk of DoS?

BrennanConroy (approved):
Prior art with doing this for SocketsHttpHandler already
https://github.com/grpc/grpc-dotnet/blob/2a36215697ced47e5b2beb73d2f5887dd4f51384/src/Shared/HttpHandlerFactory.cs#L35

amcasey (approved):

JamesNK:
> Does this increase the risk of DoS?

This is a client setting. It controls its own destiny (how many HTTP/2 requests do I make).

There is nothing stopping a client from having multiple connections via multiple `WinHttpHandler` instances. This setting makes it more user friendly by allowing one.

Issue #2419, opened 2024-04-26
Memory Leak when using ContextPropagationInterceptor


aputlock:
### What version of gRPC and what language are you using?
C#
gRPC codegen compiled with Grpc.Tools 2.40.0
Affected process importing Grpc.AspNetCore 2.50.0

### What operating system (Linux, Windows,...) and version?
Windows

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.net7

### What did you do?
My service implements a server streaming call, which as part of its implementation polls an rpc from another service. We also have ContextPropagationInterceptor added to the pipeline.

Simplified example:

```csharp
public override async Task Stream(StreamRequest request, IServerStreamWriter<StreamResponse> responseStream, ServerCallContext context)
{
    using CancellationTokenSource cts = new CancellationTokenSource();
    while (!context.CancellationToken.IsCancellationRequested)
    {
        // _otherService is a codegen'd grpc client
        var response = await _otherService.GetStateAsync(new GetStateRequest(), new Metadata(), cancellationToken: cts.Token);
        await responseStream.WriteAsync(response, context.CancellationToken);
    }
}
```

### What did you expect to see?
All resources cleaned up

### What did you see instead?
Process memory increases over time -- memory dump + ETW trace indicates there's a leak of Linked2CancellationTokenSource objects (+ associated CallbackRegistration, CallbackNode).

ContextPropagationInterceptor allocates a LinkedCancellationTokenSource and sets it up to be disposed when the containing GrpcCall is disposed. However, it doesn't seem like the call is ever implicitly disposed.

Using unary calls as an example:
https://github.com/grpc/grpc-dotnet/blob/c9c902c8daa54d2bd47a638df6a9d47fc5cae5d7/src/Grpc.AspNetCore.Server.ClientFactory/ContextPropagationInterceptor.cs#L153
https://github.com/grpc/grpc-dotnet/blob/c9c902c8daa54d2bd47a638df6a9d47fc5cae5d7/src/Grpc.AspNetCore.Server.ClientFactory/ContextPropagationInterceptor.cs#L115-L121
https://github.com/grpc/grpc-dotnet/blob/c9c902c8daa54d2bd47a638df6a9d47fc5cae5d7/src/Grpc.AspNetCore.Server.ClientFactory/ContextPropagationInterceptor.cs#L236
[...]

PR #2421, opened 2024-04-28
Fix memory leak when using call context propagation with cancellation token

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2419

There was an incorrect assumption that the disposable call was always disposed of by the client. It isn't, which could create a memory leak under the right situation (context propagation + long lasting cancellation token + many unary calls)

Fix is to make the interceptor configure the returned call type to automatically dispose its own state once finished with reading the response.

### DIFF
--- a/src/Grpc.AspNetCore.Server.ClientFactory/ContextPropagationInterceptor.cs
+++ b/src/Grpc.AspNetCore.Server.ClientFactory/ContextPropagationInterceptor.cs
@@ -22,7 +22,6 @@ using Grpc.Core;
 using Grpc.Core.Interceptors;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
 
 namespace Grpc.AspNetCore.ClientFactory;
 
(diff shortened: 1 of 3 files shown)


src/Grpc.AspNetCore.Server.ClientFactory/ContextPropagationInterceptor.cs
```diff
@@ -129,6 +132,19 @@ public ContextPropagationInterceptor(GrpcContextPropagationOptions options, IHtt
         return response;
     }
 
+    // Automatically dispose state after awaiting the response.
+    private static async Task<TResponse> OnResponseAsync<TResponse>(Task<TResponse> task, IDisposable state)
+    {
+        try
+        {
+            return await task.ConfigureAwait(false);
```
mgravell, line 140: Might this be a rare occasion when ContinueWith makes more sense?

mgravell (approved):
Seems fine; added one "consider, but up to you"; a shame this couldn't go into the underlying stream to prevent needing a decorator

src/Grpc.AspNetCore.Server.ClientFactory/ContextPropagationInterceptor.cs
JamesNK: Why would it be better here? From what I know, async/await is preferred.

PR #2445, opened 2024-05-30
Assign primary handler before other configuration

JamesNK:
PrimaryHandler is non-nullable so we shouldn't be setting it to null. This will cause problems when combined with changes in https://github.com/dotnet/runtime/pull/101808

This PR assigns primary handler at the start of config and tracks the initial value to see whether it's changed at post-configure if not valid.

### DIFF
--- a/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
+++ b/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
@@ -310,7 +310,7 @@ public static class GrpcClientServiceExtensions
         builder.Services.AddTransient<TClient>(s =>
         {
             var clientFactory = s.GetRequiredService<GrpcClientFactory>();
-            return clientFactory.CreateClient<TClient>(builder.Name);
+            return clientFactory.CreateClient<TClient>(name);
         });
 
         // Insert primary handler before other configuration so there is the opportunity to override it.
@@ -343,7 +343,7 @@ public static class GrpcClientServiceExtensions
         services.Insert(0, configurePrimaryHandler);
 
         // Some platforms don't have a built-in handler that supports gRPC.
-        // Validate that a handler was set after all configuration has run.
+        // Validate that a handler was set by the app to after all configuration has run.
         services.PostConfigure<HttpClientFactoryOptions>(name, options =>
         {
             options.HttpMessageHandlerBuilderActions.Add(builder =>
--- a/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
+++ b/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
@@ -306,21 +306,21 @@ public static class GrpcClientServiceExtensions
     {
         ArgumentNullThrowHelper.ThrowIfNull(services);
 
-        services
-            .AddHttpClient(name)
-            .ConfigurePrimaryHttpMessageHandler(() =>
-            {
-                // Set PrimaryHandler to null so we can track whether the user
-                // set a value or not. If they didn't set their own handler then
-                // one will be created by PostConfigure.
-                return null!;
-            });
+        services.AddHttpClient(name);
+
+        // Get PrimaryHandler o we can track whether the user set a value or not.
+        // This action comes before registered user actions so the primary handler here will be the one created by the factory.
+        HttpMessageHandler? initialPrimaryHandler = null;
+        services.Configure<HttpClientFactoryOptions>(name, options =>
+        {
+            options.HttpMessageHandlerBuilderActions.Add(b => initialPrimaryHandler = b.PrimaryHandler);
+        });
 
         services.PostConfigure<HttpClientFactoryOptions>(name, options =>
         {
             options.HttpMessageHandlerBuilderActions.Add(builder =>
             {
-                if (builder.PrimaryHandler == null)
+                if (builder.PrimaryHandler == initialPrimaryHandler)
                 {
                     // This will throw in .NET Standard 2.0 with a prompt that a user must set a handler.
                     // Because it throws it should only be called in PostConfigure if no handler has been set.
--- a/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
+++ b/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
@@ -16,6 +16,7 @@
 
 #endregion
 
+using System;
 using System.Diagnostics.CodeAnalysis;
 using Grpc.Net.ClientFactory;
 using Grpc.Net.ClientFactory.Internal;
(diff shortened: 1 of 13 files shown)


src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
```diff
@@ -306,21 +306,23 @@ public static class GrpcClientServiceExtensions
...
-            });
+        services.AddHttpClient(name);
+
+        // Get PrimaryHandler and store it to compare later. Used to track whether the user set a handler or not.
+        // This action comes before registered user actions so the primary handler here will be the one created by the factory.
+        HttpMessageHandler? initialPrimaryHandler = null;
+        services.Configure<HttpClientFactoryOptions>(name, options =>
+        {
+            options.HttpMessageHandlerBuilderActions.Add(b => initialPrimaryHandler = b.PrimaryHandler);
+        });
```
CarnaViire, line 317: I'd like to discourage the use of `HttpMessageHandlerBuilderActions` if possible... we've already obsoleted `ConfigureHttpMessageHandlerBuilder`

```suggestion
        // Get PrimaryHandler and store it to compare later. Used to track whether the user set a handler or not.
        // This action comes before registered user actions so the primary handler here will be the one created by the factory.
        HttpMessageHandler? initialPrimaryHandler = null;
        
        services
            .AddHttpClient(name)
            .ConfigurePrimaryHttpMessageHandler((primaryHandler, _) =>
            {
                initialPrimaryHandler = primaryHandler;
            });
```

> This action comes before registered user actions so the primary handler here will be the one created by the factory.

Hmmmmmm btw -- it won't be true anymore if the user calls `ConfigureHttpClientDefaults`, right?

(on the other hand, it does _already_ work "unexpectedly" if e.g. the user calls `services.AddHttpClient(name)` manually for some reason, _before_ `AddGrpcHttpClient` 😅)

So:

Couldn't this logic use `ConfigureHttpClientDefaults` instead? or is there smth that prevents it?

I assume, it could be the fact that the user can choose to use _both_ gRPC and "ordinary" HttpClientFactory?

CarnaViire: Can be related: https://github.com/dotnet/runtime/issues/101719
(e.g. track gRPC client names and apply the defaults only to them 🤔)

Another thought:

What is the reason to _actually_ do this as a post-configure? Is it to avoid creating the handler that won't be needed (if the user ends up creating their own handler)?

if so, there was a draft of an optimization to skip unnecessary primary handler creations if possible: https://github.com/dotnet/runtime/pull/90272/files#diff-282c22a5ee4a798e35d989d1b19b7072d4defc52573b31a876ec440016b4086eR138 -- I can extract it to a separate PR and push it forward. Will that help?

src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
```diff
@@ -306,21 +306,23 @@ public static class GrpcClientServiceExtensions
...
 
         services.PostConfigure<HttpClientFactoryOptions>(name, options =>
         {
             options.HttpMessageHandlerBuilderActions.Add(builder =>
             {
-                if (builder.PrimaryHandler == null)
+                // If the primary handler is unchanged from what the factory created then replace the handler
+                // with one that has settings optimized for a gRPC client.
+                if (builder.PrimaryHandler == initialPrimaryHandler)
```
CarnaViire, line 325: > If the primary handler is unchanged from what the factory created

It is possible to modify the existing instance, so theoretically the user could have made some changes to a "factory-created" handler -- that would be lost here 🤔

While I assume it never happened with the existing code, bc currently there was always null instead of a "factory-created" handler -- but it could happen in the future, and the loss of user configuration will be a bug.

imagine e.g.:

```c#
services.AddGrpcHttpClient<TClient>("name")
    .ConfigurePrimaryHttpMessageHandler((h, _) =>
	{
	    if (h is HttpClientHandler hch)
	    {
	        hch.Credentials = ...
	    }
	    else if (h is SocketsHttpHandler shh)
	    {
	        shh.Credentials = ...
	    }
	});
```

This will not work because the handler will be overwritten by the post-configure 🤔

src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
JamesNK: grpc-dotnet references 6.0 versions of the packages because the oldest supported .NET Core target is .NET 6. `ConfigureHttpClientDefaults` isn't available.

> What is the reason to actually do this as a post-configure? Is it to avoid creating the handler that won't be needed (if the user ends up creating their own handler)?

Yes. But it isn't important as avoiding unexpected behavior. The code runs once per named client, so the overhead is minimal.

src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
JamesNK: Hmm, that's true.

It feels more and more like the best thing to do is use `ConfigureHttpClientDefaults`. If `IHttpClientBuilder.Name` matches a gRPC client name, then set the primary handler at the start of configuration instead of the end.

CarnaViire:
> Would it be safe for a .NET 6 target to reference 8.0 version of Microsoft.Extensions.Http?

Yes, it's built for .NET Standard 2.0.

src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
```diff
@@ -336,37 +332,34 @@ public static class GrpcClientServiceExtensions
...
+                else
+                {
+                    b.PrimaryHandler = UnsupportedHttpHandler.Instance;
                 }
             });
-        });
-
-        var builder = new DefaultHttpClientBuilder(services, name);
+        }));
+        services.Insert(0, configurePrimaryHandler);
```
CarnaViire, line 343: Ahah we're substituting one hack with another hack 😅

src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs
```diff
@@ -306,25 +305,22 @@ public static class GrpcClientServiceExtensions
...
-                return null!;
-            });
+        var builder = services.AddHttpClient(name);
 
-        services.PostConfigure<HttpClientFactoryOptions>(name, options =>
+        builder.Services.AddTransient<TClient>(s =>
         {
-            options.HttpMessageHandlerBuilderActions.Add(builder =>
+            var clientFactory = s.GetRequiredService<GrpcClientFactory>();
+            return clientFactory.CreateClient<TClient>(builder.Name);
```
CarnaViire, line 313: Why not just `name`?

CarnaViire (approved):
LGTM

PR #2448, opened 2024-06-01
Support some HttpClientAction settings and log debug message instead of an error

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/4233

### DIFF
--- a/src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
+++ b/src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
@@ -39,6 +39,7 @@ internal class GrpcCallInvokerFactory
     private readonly IServiceScopeFactory _scopeFactory;
     private readonly ConcurrentDictionary<EntryKey, CallInvoker> _activeChannels;
     private readonly Func<EntryKey, CallInvoker> _invokerFactory;
+    private readonly ILogger<GrpcCallInvokerFactory> _logger;
 
     public GrpcCallInvokerFactory(
         IServiceScopeFactory scopeFactory,
@@ -57,6 +58,7 @@ internal class GrpcCallInvokerFactory
         _scopeFactory = scopeFactory;
         _activeChannels = new ConcurrentDictionary<EntryKey, CallInvoker>();
         _invokerFactory = CreateInvoker;
+        _logger = _loggerFactory.CreateLogger<GrpcCallInvokerFactory>();
     }
 
     public CallInvoker CreateInvoker(string name, Type type)
@@ -75,7 +77,7 @@ internal class GrpcCallInvokerFactory
             var httpClientFactoryOptions = _httpClientFactoryOptionsMonitor.Get(name);
             if (httpClientFactoryOptions.HttpClientActions.Count > 0)
             {
-                throw new InvalidOperationException($"The ConfigureHttpClient method is not supported when creating gRPC clients. Unable to create client with name '{name}'.");
+                Log.UnsupportedHttpClientActions(_logger, name);
             }
 
             var clientFactoryOptions = _grpcClientFactoryOptionsMonitor.Get(name);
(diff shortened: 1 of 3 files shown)


src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
```diff
@@ -190,4 +192,15 @@ public override void SetCompositeCredentials(object state, ChannelCredentials ch
         {
         }
     }
+
+    private static class Log
+    {
+        private static readonly Action<ILogger, string, Exception?> _unsupportedHttpClientActions =
+            LoggerMessage.Define<string>(LogLevel.Debug, new EventId(1, "UnsupportedHttpClientActions"), "The ConfigureHttpClient method is not supported when creating gRPC clients. Unable to apply configuration to client with the name '{ClientName}'.");
```
amcasey, line 199: This appears to indicate a misconfiguration - maybe it should be a warning?

src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
```diff
@@ -75,7 +77,7 @@ private CallInvoker CreateInvoker(EntryKey key)
             var httpClientFactoryOptions = _httpClientFactoryOptionsMonitor.Get(name);
             if (httpClientFactoryOptions.HttpClientActions.Count > 0)
```
amcasey, line 78: To confirm, the HttpClientActions will just be ignored?

amcasey (approved):
Feels like a Warning to me, but I leave it up to you.

src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
JamesNK: I've changed the log message content. I think debug is the right level for the new message.

src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
JamesNK: I've come up with a semi-hacky way to make them used. Actions now run on a dummy HttpClient and values are then partially applied to gRPC client. It's not perfect, but should cover most scenarios, and a log message provides more info.

src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
```diff
@@ -75,12 +77,58 @@ private CallInvoker CreateInvoker(EntryKey key)
...
+                foreach (var applyOptions in httpClientFactoryOptions.HttpClientActions)
+                {
+                    applyOptions(httpClient);
+                }
+
+                // Copy configuration from HttpClient to GrpcChannel/CallOptions.
+                // This configuration should be overriden by gRPC specific config methods.
+                if (clientFactoryOptions.Address == null)
+                {
+                    clientFactoryOptions.Address = httpClient.BaseAddress;
```
amcasey, line 100: Would it be interesting to log when the HttpClient value is present but ignored?

src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
```diff
@@ -75,12 +77,58 @@ private CallInvoker CreateInvoker(EntryKey key)
...
+                    // Don't replace or add new header values if the header name has already been set.
+                    clientFactoryOptions.CallOptionsActions.Add(callOptionsContext =>
+                    {
+                        var metadata = callOptionsContext.CallOptions.Headers ?? new Metadata();
+                        foreach (var entry in defaultHeaders)
+                        {
+                            // grpc requires header names are lower case before being added to collection.
+                            var resolvedKey = entry.Key.ToLower(CultureInfo.InvariantCulture);
+
+                            if (metadata.Get(resolvedKey) == null)
```
amcasey, line 118: Would it be interesting to log when it's not null and values form the HttpClient are ignored?

amcasey (approved):
LGTM

src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
JamesNK: I don't think we need to go to this detail.

src/Grpc.Net.ClientFactory/Internal/GrpcCallInvokerFactory.cs
JamesNK: I don't think we need to go to this detail.

PR #2453, opened 2024-06-05
Update various package versions

JamesNK:


Directory.Build.targets
```diff
@@ -30,6 +30,12 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup Condition="$(MSBuildProjectName.EndsWith('Tests'))">
```
amcasey, line 33: Someone will eventually call a project FooTest and then include these manually. 😆 I don't think we need to do anything about it.

amcasey (approved):

PR #2465, opened 2024-06-20
Update Grpc.Tools 1.65.0-pre1

apolcyn:
As title, prep for branch cut

@JamesNK


JamesNK (changes requested):
This isn't needed. The next version is 2.64.0 and I already updated Grpc.Tools to 1.64.0.

PR #2471, opened 2024-06-28
Bump branch version to 2.64.0-pre1

stanley-cheung:
This is to bump the branch `v2.64.x` version to `2.64.0-pre1`.


apolcyn (approved):
cc @JamesNK FYI (I think the last couple of releases had the minor version string wrong here)

Issue #2483, opened 2024-07-15
feature: Grpc.AspNetCore.Server.Reflection Allow Anonymous, even if authentication is required by default


wooln:
### Is your feature request related to a problem? Please describe.
Reflection service blocks anonymous access when the application set authentication is required by default.

### Describe the solution you'd like
Add [AllowAnonymous] attribute on Grpc.Reflection.ReflectionServiceImpl

### Describe alternatives you've considered
(1) Set ReflectionServiceImpl allow anonymous in my auth policy code;
(2) New class Inherits the base class ReflectionServiceImpl, set [AllowAnonymous].

JamesNK:
I don't think it is a good change to allow anonymous access to this by default. It would be unexpected and potentially create a security vulnerability.

What someone can do to make the reflection service accessible is add the `AllowAnonymous()` call:

```cs
var builder = WebApplication.CreateBuilder(args);
builder.Services.AddGrpcReflection();

var app = builder.Build();
app.MapGrpcReflectionService().AllowAnonymous();
app.Run();
```

Does that fix your scenario?

wooln:
> I don't think it is a good change to allow anonymous access to this by default. It would be unexpected and potentially create a security vulnerability.
> 
> What someone can do to make the reflection service accessible is add the `AllowAnonymous()` call:
> 
> ```cs
> var builder = WebApplication.CreateBuilder(args);
> builder.Services.AddGrpcReflection();
> 
> var app = builder.Build();
> app.MapGrpcReflectionService().AllowAnonymous();
> app.Run();
> ```
> 
> Does that fix your scenario?

Great, it works. knowledge+1
``` csharp
public static class AuthorizationEndpointConventionBuilderExtensions  {
        public static TBuilder AllowAnonymous<TBuilder>(this TBuilder builder) where TBuilder : IEndpointConventionBuilder;
        ...
}
```

PR #2484, opened 2024-07-15
feature: ReflectionServiceImpl allow anonymous

wooln:
In general, the API allows anonymity default, but in our app it defaults to non-anonymity. The purpose is usually only enabled in the development environment and anonymity is allowed.
#2483


wooln:
fix my scenario: ReflectionServiceImpl allow anonymous #2484 ，`app.MapGrpcReflectionService().AllowAnonymous();`

PR #2496, opened 2024-07-30
Fix failure to create GrpcChannel under Wine compatibility layer (including Steam Proton and Apple Game Porting Toolkit)

mayuki:
Wine compatibility layers such as Steam Deck/Steam OS Proton and the Apple Game Porting Toolkit return Windows 8 or later as the OS version, but does not implement the `GetCurrentApplicationUserModelId` API.

As a result, an issue occurs where it is not possible to create a `GrpcChannel` due to an `EntryPointNotFoundException`.

```
System.TypeInitializationException: The type initializer for 'Grpc.Net.Client.Internal.OperatingSystem' threw an exception. ---> System.EntryPointNotFoundException: GetCurrentApplicationUserModelId assembly:<unknown assembly> type:<unknowntype> member:(null)
  at (wrapper managed-to-native) Grpc.Net.Client.Internal.Native.GetCurrentApplicationUserModelId(uint&,byte[])
  at Grpc.Net.Client.Internal.Native.IsUwp (System.String frameworkDescription, System.Version version) [0x00028] in <cfd57654def7423a841054daeb7fc483>:0
  at Grpc.Net.Client.Internal.OperatingSystem..ctor () [0x00049] in <cfd57654def7423a841054daeb7fc483>:0
  at Grpc.Net.Client.Internal.OperatingSystem..cctor () [0x00000] in <cfd57654def7423a841054daeb7fc483>:0
```

For example, if using Grpc.Net.Client in a game developed with Unity, the game will not be playable on Steam Deck.

### DIFF
--- a/src/Grpc.Net.Client/Internal/Native.cs
+++ b/src/Grpc.Net.Client/Internal/Native.cs
@@ -64,21 +64,30 @@ internal static class Native
         }
         else
         {
-            var bufferSize = 0U;
-            var result = GetCurrentApplicationUserModelId(ref bufferSize, Array.Empty<byte>());
-            switch (result)
+            try
             {
-                case 15703: // APPMODEL_ERROR_NO_APPLICATION
-                    return false;
-                case 0:     // ERROR_SUCCESS
-                case 122:   // ERROR_INSUFFICIENT_BUFFER
-                            // Success is actually insufficient buffer as we're really only looking for
-                            // not NO_APPLICATION and we're not actually giving a buffer here. The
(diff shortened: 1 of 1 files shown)


JamesNK:
Do you have a link to a web page that discusses this? It would be good to include that in the comment so people can look their if they want more information.

mayuki:
This originated from a report from one of our clients, and I'm sorry to say there is no similar discussion on the web.

JamesNK:
Thanks for the good description and fix.

PR #2512, opened 2024-08-19
Fix Google auth interop test

JamesNK:


JamesNK:
@apolcyn @jskeet Interop tests pass after downgrading Google.Apis.Auth.

Does that mean the change in Google.Apis.Auth that causes the interop error is intentional or a regression?

Problematic line: https://github.com/grpc/grpc-dotnet/blob/c397bc7804824100c983052f4a6e7e291ddf1c06/testassets/Shared/InteropClient.cs#L248

jskeet:
cc @amanda-tarafa who is on holiday at the moment, but will want to look at this later.

I strongly suspect this is basically just a test issue, due to the change in https://github.com/googleapis/google-api-dotnet-client/commit/bd80f0afb3f22f2e91ca3a29d7d7cbca6f9b4bf2 which introduced the *feature* of Compute Credentials accepting scopes. (The name of `IsCreateScopedRequired` is unfortunate - it's more "permitted" than "required" IMO, but Amanda can give more details.)

If I'm right, then we should be able to narrow down the version range: I'd expect the current test to pass with 1.57.0 and fail with 1.58.0.

I suspect you could invert the assertion, or just remove it.

jskeet:
> I'm guessing the now-failing check was trying to do a sanity check that we did in fact resolve a ComputeCredential

If that's the case, there's a *much* simpler alternative, at least now - fetch `googleCredential.UnderlyingCredential` and check that it's a `ComputeCredential`.

JamesNK:
Test is fixed. @apolcyn feel free to approve and merge.

apolcyn (approved):

amanda-tarafa:
I'm back now, just catching up.

The name `IsCreateScopeRequired` is indeed very unfortunate but legacy and we cannot remove it without a significant breaking change. For reference, we internally call this [SupportsExplicitScopes](https://github.com/googleapis/google-api-dotnet-client/blob/f3dcaf8079604c01763e8f03d5f1e7b3d61e0541/Src/Support/Google.Apis.Auth/OAuth2/IGoogleCredential.cs#L58) which is a way better name.

Although I cannot comment on the original intention of testing for "IsCreateScopeRequired" to be false, I can confirm the rest of  @apolcyn's [comment](https://github.com/grpc/grpc-dotnet/pull/2512#issuecomment-2298062140). `ComputeCredential.IsCreateScopeRequired` will always be true now because some (but we cannot tell which) ComputeCredentials support explicit scoping. The best way to check the credential to be a ComputeCredential is Jon's suggestion that's been implemented, so all good here.

Issue #2404, opened 2024-04-02
Only 100 streaming calls get responses under very specific conditions


jskeet:
Original GitHub issue: https://github.com/googleapis/google-cloud-dotnet/issues/12318
Diagnostic test solution: https://github.com/googleapis/google-cloud-dotnet/tree/main/issues/Issue12318

The most useful directory in there (in terms of looking at the code) is probably RawGrpcStubs.

Currently, I can reproduce this when talking to Firestore. I haven't managed to reproduce it with a local gRPC server - you can see attempts in the repro directory above, but they didn't come to anything.

The repro goes:

- Code: this is *just* the code generated by the vanilla C# protobuf generator and the gRPC plugin; it *doesn't* include the code we normally use for full Google client libraries ("GAPIC" code) or the handwritten Firestore code.
- Prework: Create a collection with 300 documents in (the PopulateCollection project does this if you want to repro yourself against Firestore)
- With a single FirestoreClient
  - Call ListDocuments (a unary RPC) to get all the documents
  - For each document, call Listen (a bidi streaming RPC), send a single request within that stream and observe whether we see any responses
  - Wait for 10 seconds to give everything a chance to stabilize, then report results

I would *expect* every listener to receive a few responses - so we'd report at the end "We have 300 documents, and 300 listeners received a response."

What we actually see is:

- On Linux (.NET 6 and 8; not tested on 7), expected behavior
- On Windows 10 (.NET 6, 7 and 8), expected behavior
- On Windows 11, .NET 6, expected behavior
- On Windows 11, .NET 7, 8 and 9-preview: only 100 of the 300 listeners receive a response

I can give more version details if that would be useful - let me know exactly what's required.

Some experiments:

- Using a different client (with a different channel) for the initial unary RPC: all 300 listeners work. It's as if the initial unary call taints the channel somehow, stopping it from handling all 300 listeners, only handling 100 at a time.
- Running the "listen" part in multiple iterations: we get 100 active clients per iteration
- Explicitly specifying the HttpHandler as a SocketsHttpHandler with various options: doesn't *seem* to make a difference (experimented with EnableMultipleHttp2Connections and MaxConnectionsPerServer)
- Tried using nightly version 2.61.0-dev202404020702 (with Grpc.Auth 2.61.0) - still just 100 listeners with .NET 8

It's not clear to me what the next step is in terms of diagnosing this - any suggestions would be welcome.

[...]

PR #2514, opened 2024-08-21
Add HTTP version configuration to GrpcChannelOptions

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2404

The grpc client today can automatically upgrade to HTTP3. That is unexpected for most users and it can cause changes in behavior.

Also, there are situations where you want more exact control over HTTP version. Having to use a delegating handler is an awkward solution.

This PR:

* Adds `GrpcChannelOptions.HttpVersion` and `GrpcChannelOptions.HttpVersionPolicy`.
* Obsoletes `GrpcWebHandler.HttpVersion` and points users to the channel options
* Changes default version policy behavior to `RequestVersionExact`. Client will use HTTP/2 unless configured otherwise.

### DIFF
--- a/src/Grpc.Net.Client.Web/GrpcWebHandler.cs
+++ b/src/Grpc.Net.Client.Web/GrpcWebHandler.cs
@@ -44,6 +44,7 @@ public sealed class GrpcWebHandler : DelegatingHandler
     /// be overridden.
     /// </para>
     /// </summary>
+    [Obsolete("HttpVersion is obsolete and will be removed in a future release. Use GrpcChannelOptions.HttpVersion and GrpcChannelOptions.HttpVersionPolicy instead.")]
     public Version? HttpVersion { get; set; }
 
     /// <summary>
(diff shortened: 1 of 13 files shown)


JamesNK:
cc @jskeet

jskeet:
Wonderful, thank you so much!

src/Grpc.Net.Client/GrpcChannelOptions.cs
```diff
@@ -287,6 +287,35 @@ public TimeSpan InitialReconnectBackoff
...
+    /// <para>
+    /// When a <see cref="Version"/> is specified the value will be set on <see cref="HttpRequestMessage.Version"/>
+    /// as gRPC calls are made. Changing this property allows the HTTP version of gRPC calls to
+    /// be overridden.
+    /// </para>
+    /// <para>
+    /// A <c>null</c> value doesn't override the HTTP version of gRPC calls. Defaults to <c>2.0</c>.
+    /// </para>
+    /// </summary>
+    public Version? HttpVersion { get; set; }
```
amcasey, line 301: I assume something blows up downstream if someone sets `2.5` or `7.0`, but what happens if they select `1.1`?

src/Grpc.Net.Client.Web/GrpcWebHandler.cs
```diff
@@ -44,6 +44,7 @@ public sealed class GrpcWebHandler : DelegatingHandler
     /// be overridden.
     /// </para>
     /// </summary>
+    [Obsolete("HttpVersion is obsolete and will be removed in a future release. Use GrpcChannelOptions.HttpVersion and GrpcChannelOptions.HttpVersionPolicy instead.")]
```
amcasey, line 47:
```suggestion
#if NET5_0_OR_GREATER
    [Obsolete("HttpVersion is obsolete and will be removed in a future release. Use GrpcChannelOptions.HttpVersion and GrpcChannelOptions.HttpVersionPolicy instead.")]
#else
    [Obsolete("HttpVersion is obsolete and will be removed in a future release. Use GrpcChannelOptions.HttpVersion instead.")]
#endif
```

src/Grpc.Net.Client/GrpcChannelOptions.cs
amcasey: From the tests, it seems like this is handled somewhere (or might even be allowable?).

amcasey (approved):

src/Grpc.Net.Client/GrpcChannelOptions.cs
JamesNK: 1.1 will successfully create a connection if the server supports 1.1 and you're using gRPC-Web.

PR #2531, opened 2024-09-09
Update logging to use generated logs

### LINKED ISSUE #2529: Update logging to use generated logs
Logging makes heavy use of `LoggerMessage.Define`. It provides named, structured logging with good performance.
https://github.com/grpc/grpc-dotnet/blob/3ed9fb9f31474b4d1318051123a37d9b6f68cda4/src/Grpc.AspNetCore.Server/Internal/GrpcServerLog.cs

Logging code can be simplified and reduced by using generated logs:
https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator

I think now that `Microsoft.Extensions.Logging` is 6.0+ we should be able to use this feature.

Note that there are a lot of logs. It will be a fair bit of work to do all of them. I don't think there is anything stopping one type at a time from being updated.

wabalubdub:
This PR uses the generated logs feature to simplify and reduce the log declarations.

closes #2529 

Also my first PR so feedback would be lovely.

### DIFF
--- a/src/Grpc.AspNetCore.Server/Internal/GrpcServerLog.cs
+++ b/src/Grpc.AspNetCore.Server/Internal/GrpcServerLog.cs
@@ -23,11 +23,14 @@ namespace Grpc.AspNetCore.Server.Internal;
 
 internal static partial class GrpcServerLog
 {
-    [LoggerMessage(Level = LogLevel.Trace, EventId = 27, EventName = "DeadlineStopped", Message = "Request deadline stopped.")]
-    internal static partial void DeadlineStopped(ILogger logger);
+    [LoggerMessage(Level = LogLevel.Debug, EventId = 1, EventName = "UnableToDisableMaxRequestBodySizeLimit", Message = "Unable to disable the max request body size limit.")]
+    public static partial void UnableToDisableMaxRequestBodySize(ILogger logger);
 
-    [LoggerMessage(Level = LogLevel.Trace, EventId = 26, EventName = "DeadlineStarted", Message = "Request deadline timeout of {Timeout} started.")]
-    public static partial void DeadlineStarted(ILogger logger, TimeSpan timeout);
+    [LoggerMessage(Level = LogLevel.Information, EventId = 2, EventName = "UnsupportedRequestContentType", Message = "Request content-type of '{ContentType}' is not supported.")]
+    public static partial void UnsupportedRequestContentType(ILogger logger, string? contentType);
+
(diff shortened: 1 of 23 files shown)


src/Grpc.AspNetCore.Server/Internal/GrpcServerLog.cs
```diff
@@ -21,229 +21,89 @@
 
 namespace Grpc.AspNetCore.Server.Internal;
 
-internal static class GrpcServerLog
+internal static partial class GrpcServerLog
 {
-    private static readonly Action<ILogger, Exception?> _unableToDisableMaxRequestBodySize =
-        LoggerMessage.Define(LogLevel.Debug, new EventId(1, "UnableToDisableMaxRequestBodySizeLimit"), "Unable to disable the max request body size limit.");
+    [LoggerMessage(Level = LogLevel.Trace, EventId = 27, EventName = "DeadlineStopped", Message = "Request deadline stopped.")]
```
JamesNK, line 26: Could you preserve the previous order? I'd like to have them ordered by the event id. That way new log methods can be added to the end and it's easy to see the previous event id.

Also, keeping that order also makes code reviewing the changes easier.

JamesNK:
Hi, thanks for the PR.

FYI this is just one file of many with logs. See https://github.com/search?q=repo%3Agrpc%2Fgrpc-dotnet+LoggerMessage.Define&type=code

You're welcome to update more if you'd like or stick with this one. I'll close the issue once all are updated.

wabalubdub:
Hey, of course! I ordered the methods according to Event ID.
I didn't notice there were more, ill work on the rest of them.  would you rather i update the rest of the in this PR? I can probably get to all of them tomorrow.

JamesNK:
One pr is fine

wabalubdub:
Hey I updated the PR to include changes to all the files that have the `LoggerMessage.Define` syntax. 

A couple notes:
 -  I opted to keep the interface as identical as possible, therefore when the call for logging included any formatting beyond a call to the default `ToString()`  I kept the original function signature and made a call to a private generated log after the formatting.  The LoggerMessage Attribute doesn't seem to have a way to access member variables of objects that are passed to it so I couldn't generate the formatting automatically as well. 
 -  I did the same thing when there were conditional statements for optimization like `logger.isTraceEnabled()`
 - Some of the logging classes are nested in larger classes, the generated logs requires the class to be partial so i made the surrounding class partial as well.

JamesNK (approved):

JamesNK:
Thanks! Looks good to me. There are a lot of changes so I haven't excuasively checked every log, but the ones I did look at were correct.

PR #2532, opened 2024-09-10
add target framework variable

Varorbc:
@JamesNK 
Hey, I noticed things got a bit messy when I was defining the target frameworks. Can you help me confirm the final target frameworks? For example, `Grpc.Net.Client` supports `net462`, `netstandard2.0`, `netstandard2.1`, `net6.0`, `net7.0`, and `net8.0`, but `Grpc.Net.ClientFactory `only supports `netstandard2.0`, `netstandard2.1`, `net6.0`, `net7.0`, and `net8.0`—it’s missing `net462`. Also, can you check if this is correct? And by the way, is `netstandard2.1` really necessary?

### DIFF
--- a/src/Grpc.Core.Api/AsyncStreamReaderExtensions.cs
+++ b/src/Grpc.Core.Api/AsyncStreamReaderExtensions.cs
@@ -16,9 +16,9 @@
 
 #endregion
 
-using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Grpc.Shared;
 
 namespace Grpc.Core;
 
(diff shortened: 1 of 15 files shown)


JamesNK:
> Grpc.Net.Client supports net462

net462 was added so that target could automatically include WinHttpHandler package.

netstandard2.0 supports .NET Framework so I don't think it was added anywhere else.

> And by the way, is netstandard2.1 really necessary?

It's not needed anymore.

MoonlitLithium:
I believe this would allow load balancing support for net9.0.  Which is something that is needed at the moment.

JamesNK:
Load balancing is already supported with .net9. What happens when you use it?

PR #2548, opened 2024-10-01
Cleanup gRPC unit testing helpers in tester sample

JamesNK:
* Remove old nullable suppressions
* Support `WriteAsync(message, token)` in tests

### DIFF
--- a/examples/Tester/Tests/Server/UnitTests/Helpers/TestServerStreamWriter.cs
+++ b/examples/Tester/Tests/Server/UnitTests/Helpers/TestServerStreamWriter.cs
@@ -1,4 +1,4 @@
-﻿#region Copyright notice and license
+#region Copyright notice and license
 
 // Copyright 2019 The gRPC Authors
 //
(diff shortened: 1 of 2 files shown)


examples/Tester/Tests/Server/UnitTests/Helpers/TestServerStreamWriter.cs
```diff
@@ -77,6 +79,10 @@ public Task WriteAsync(T message)
 
             return Task.CompletedTask;
         }
+
+        public Task WriteAsync(T message)
```
amcasey, line 83: Why not just provide a default value above?

amcasey (approved):

examples/Tester/Tests/Server/UnitTests/Helpers/TestServerStreamWriter.cs
JamesNK: Can't because these are interface methods.

PR #2553, opened 2024-10-08
Fix UpdateBalancingState not called when address attributes are modified

JamesNK:
gRPC load balancing would update subchannels with modified address attributes when they changed, but the overall channel state was never updated.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,5 @@
 {
   "sdk": {
-    "version": "9.0.100-preview.7.24407.12",
-    "rollForward": "latestFeature"
+    "version": "9.0.100-rc.1.24452.12"
   }
 }
--- a/global.json
+++ b/global.json
@@ -1,5 +1,6 @@
 {
   "sdk": {
-    "version": "9.0.100-rc.1.24452.12"
+    "version": "9.0.100-rc.2.24474.11",
+    "rollForward": "latestFeature"
   }
 }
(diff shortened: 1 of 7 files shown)


JamesNK:
Ready for review. I know grpc-web test is failing. Will fix it once RC2 is public.

global.json
```diff
@@ -1,6 +1,5 @@
 {
   "sdk": {
-    "version": "9.0.100-preview.7.24407.12",
```
amcasey, line 3: What's the significance of removing `rollForward`?

src/Grpc.Net.Client/Balancer/SubchannelsLoadBalancer.cs
```diff
@@ -122,6 +122,7 @@ public override void UpdateChannelState(ChannelState state)
 
         var allUpdatedSubchannels = new List<AddressSubchannel>();
         var newSubchannels = new List<Subchannel>();
+        var modifiedSubchannelsCount = 0;
```
amcasey, line 125: It doesn't seem important either way, but it looks like this could be a simple bool.

amcasey (approved):
The code seems to match the description and the description seems reasonable.

global.json
JamesNK: RC2 preview was causing trouble on my machine.

PR #2559, opened 2024-10-21
Correctness: Make some private & internal classes sealed where possible

Henr1k80:
Should in theory also be faster, as some virtual method table lookups can be avoided.
I haven't tested if it introduces any performance regressions due to runtime black magic inner workings.
My laptop cannot produce repeatable results due to throttling & background services.
I hope you have a benchmark setup to test it.

There are some whitespace changes due to:
`trim_trailing_whitespace = true`
`insert_final_newline = true`
in [`.editorconfig`](https://github.com/grpc/grpc-dotnet/blob/master/.editorconfig#L19)


JamesNK (approved):

JamesNK:
Thanks!

PR #2563, opened 2024-10-24
update ArgumentNullException.ThrowIfNull usage

WeihanLi:
The `ArgumentNullException.ThrowIfNull` is introduced since .NET 6, maybe we could use if for `NET6_0_OR_GREATER`

https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception.throwifnull?view=net-6.0


JamesNK (approved):

JamesNK:
Thanks

Issue #2572, opened 2024-11-15
Running server-streaming calls on HealthCheck API delay server shutdown


freibenjamin:
### What version of gRPC and what language are you using?

Grpc 2.46.6  
Grpc.AspNetCore.HealthChecks 2.66.0  

### What operating system (Linux, Windows,...) and version?

Windows 11  

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)

.NET SDK 8.0.404

### What did you do?

We integrated the `Grpc.AspNetCore.HealthChecks` NuGet package and tested its gRPC API with clients using the `Watch` server-streaming RPC. To monitor the server's status, we implemented a simple health check class that verifies the server's running state. However, during server shutdown, we noticed a delay if clients were connected to the `Watch` RPC at the time.

### What did you expect to see?

A graceful shutdown that completes within a few seconds at most.

### What did you see instead?

When the server stops, the shutdown is significantly delayed if clients are connected to the `Watch` RPC at the time of shutdown. The still opened calls cause the server to reach its shutdown timeout. We retained the host shutdown timeout at 30 seconds.
The screenshot below shows a client connected to the `Watch` RPC. The shutdown was initiated at 12:07:26 (via the terminal) and completed at 12:07:56.
![Screenshot 2024-11-15 120818](https://github.com/user-attachments/assets/d82d75a3-0b78-4ae8-adc6-d9ba9e1f803f)

### Anything else we should know?

This issue is related to: <https://github.com/grpc/grpc-dotnet/issues/2357>

The problem is that running server-streaming calls are not informed about the shutdown. You will find my proposed solution later in the linked PR.

PR #2573, opened 2024-11-15
Fix delayed server shutdown if clients are connected to the Watch server-streaming RPC of the Health API

freibenjamin:
Fixes <https://github.com/grpc/grpc-dotnet/issues/2572>

## Changes

- Copied the `Grpc.HealthCheck.HealthServiceImpl` class to the `Grpc.AspNetCore.HealthChecks` namespace. This was necessary because the `HealthServiceImpl` now requires the `IHostApplicationLifetime`, which is not available in the `Grpc.HealthCheck` package. I assumed you prefer not to modify references of that package. I did not change the original implementation to ensure backwards compatibility.
- Used the `ApplicationStopping` cancellation token in the `Grpc.AspNetCore.HealthChecks.HealthServiceImpl` class to gracefully terminate `Watch` streams during server shutdown. The `ApplicationStopping` token was passed to the `WaitToReadAsync` method to allow running calls to finish when the server is stopping. In the catch block, a `NotServing` response is written to the response stream, though an `RpcException` could be thrown if preferred.
- Minor changes to comments.

## Testing

The following screenshot shows that the server now stops within a second, even if clients are connected to the `Watch` RPC at the time of shutdown. The server was stopped at 12:09:17 (via the terminal).
Screenshot 2024-11-15 120952

### DIFF
--- a/src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksServiceExtensions.cs
+++ b/src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksServiceExtensions.cs
@@ -1,6 +1,6 @@
 #region Copyright notice and license
 
-// Copyright 2019 The gRPC Authors
+// Copyright 2024 The gRPC Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


JamesNK:
Thanks for the PR. This is a good improvement but I don't like duplicating the implementation of the health checks service.

I've made a new PR that adds the same capability but with a different implementaiton: https://github.com/grpc/grpc-dotnet/pull/2582

freibenjamin:
Thank you, @JamesNK, for addressing the issue. I agree that duplicating the health check service wasn't the most elegant approach.

PR #2582, opened 2024-12-09
Complete health checks watch service on server shutting down

JamesNK:
Fixes the same issue as https://github.com/grpc/grpc-dotnet/pull/2573: health checks `Watch` call completes when the server begins shutting down. This avoids the problem of a shutting down server waiting for the health check call to be complete (default is 30 seconds) before forcefully killing the request.

Differences:
* Avoids duplicating the health checks service implementaiton. Cancellation is now passed in via ServerCallContext
* Adds option to enable and disable this behavior. Completing on shutdown is enabled by default with the option to suppress it and go back to old behavior.
* Adds tests

### DIFF
--- a/src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksOptions.cs
+++ b/src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksOptions.cs
@@ -17,6 +17,7 @@
 #endregion
 
 using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Microsoft.Extensions.Hosting;
 
 namespace Grpc.AspNetCore.HealthChecks;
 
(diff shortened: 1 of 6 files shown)


BrennanConroy (approved):

src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksOptions.cs
```diff
@@ -39,4 +40,21 @@ public sealed class GrpcHealthChecksOptions
     /// published by <see cref="IHealthCheckPublisher"/> are returned.
     /// </remarks>
     public bool UseHealthChecksCache { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether to complete <c>Watch</c> health check calls when the application is stopping.
```
JamesNK, line 45:
```suggestion
    /// Gets or sets a value indicating whether to suppress completing <c>Watch</c> health check calls when the application begins shutting down.
```

src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksOptions.cs
```diff
@@ -39,4 +40,21 @@ public sealed class GrpcHealthChecksOptions
...
+    /// The default value is <c>false</c>.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// When <c>false</c>, health checks <c>Watch</c> calls are completed with a status of NotServing when the server application begins shutting down.
+    /// Shutdown is indicated by the <see cref="IHostApplicationLifetime.ApplicationStopping"/> token being raised and causes <c>Watch</c> to complete.
+    /// When <c>true</c>, health checks <c>Watch</c> calls are left running until the server forcefully aborts the request when the server shuts down.
```
JamesNK, line 52:
```suggestion
    /// When <c>true</c>, this behavior is suppressed and health checks <c>Watch</c> calls are left running. Running calls will be eventually be forcefully aborted when the server finishes shutting down.
```

src/Grpc.AspNetCore.HealthChecks/GrpcHealthChecksOptions.cs
```diff
@@ -39,4 +40,21 @@ public sealed class GrpcHealthChecksOptions
...
+    /// </para>
+    /// <para>
+    /// Completing the <c>Watch</c> call allows the server to gracefully exit. If <c>Watch</c> calls aren't shutdown then the server runs until
+    /// <see cref="HostOptions.ShutdownTimeout"/> is exceeded and the server forcefully aborts the request.
```
JamesNK, line 56:
```suggestion
    /// <see cref="HostOptions.ShutdownTimeout"/> is exceeded and the server forcefully aborts remaining active requests.
```

PR #2596, opened 2025-02-14
Don't compress if data.Length is 0

lewing:
There is no need to compress 0 length messages and in .NET 9 GZip stream no longer writes the zlib wrapper which can break consumers. Instead we send a 5 byte empty message

Purely speculative change based off of https://github.com/dotnet/runtime/issues/112563


lewing:
is it valid to not compress the response?

PR #2597, opened 2025-02-16
Avoid using ConcurrentDictionary for channels with few methods

JamesNK:
`GrpcChannel` has a `ConcurrentDictionary` to cache method details. However, if the channel never has more than a handful of methods then `ConcurrentDictionary` uses a large amount of memory to store just a few instances.

This PR adds a `ThreadSafeLookup` type that uses a replacing array for a small number of items and upgrades to a dictionary beyond a threashold.

### DIFF
--- a/src/Grpc.Net.Client/Internal/ThreadSafeLookup.cs
+++ b/src/Grpc.Net.Client/Internal/ThreadSafeLookup.cs
@@ -22,7 +22,7 @@ internal sealed class ThreadSafeLookup<TKey, TValue> where TKey : notnull
 {
     // Avoid allocating ConcurrentDictionary until the threshold is reached.
     // Looking up a key in an array is as fast as a dictionary for small collections and uses much less memory.
-    private const int Threshold = 10;
+    internal const int Threshold = 10;
 
     private KeyValuePair<TKey, TValue>[] _array = Array.Empty<KeyValuePair<TKey, TValue>>();
     private ConcurrentDictionary<TKey, TValue>? _dictionary;
@@ -70,6 +70,8 @@ internal sealed class ThreadSafeLookup<TKey, TValue> where TKey : notnull
         }
         else
         {
+            // Add new value by creating a new array with old plus new value.
+            // This allows for lookups without locks and is more memory efficient than a dictionary.
             var newArray = new KeyValuePair<TKey, TValue>[snapshot.Length + 1];
             Array.Copy(snapshot, newArray, snapshot.Length);
             newArray[newArray.Length - 1] = new KeyValuePair<TKey, TValue>(key, newValue);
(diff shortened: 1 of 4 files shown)


src/Grpc.Net.Client/Internal/ThreadSafeLookup.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            if (EqualityComparer<TKey>.Default.Equals(kvp.Key, key))
+            {
+                return kvp.Value;
+            }
+        }
+
+        var newValue = valueFactory(key);
+
+        if (snapshot.Length + 1 > Threshold)
```
gfoidl, line 48:
```suggestion
        if (snapshot.Length > Threshold - 1)
```
To have the constants on one side only?

Well perf and potential overflow won't matter here, but good code style :wink:.

src/Grpc.Net.Client/Internal/ThreadSafeLookup.cs
```diff
@@ -0,0 +1,84 @@
...
+                    _dictionary = newDict;
+                    _array = Array.Empty<KeyValuePair<TKey, TValue>>();
+                }
+            }
+        }
+        else
+        {
+            // Add new value by creating a new array with old plus new value.
+            // This allows for lookups without locks and is more memory efficient than a dictionary.
+            var newArray = new KeyValuePair<TKey, TValue>[snapshot.Length + 1];
```
gfoidl, line 75: When `newArray` is created with `Thresold`-size is the memory usage too high if there is only 1 or two methods around?

But it would safe the "resizing" of the array, and thus the intermediate garbage.

src/Grpc.Net.Client/Internal/ThreadSafeLookup.cs
```diff
@@ -0,0 +1,84 @@
...
+            Array.Copy(snapshot, newArray, snapshot.Length);
```
BrennanConroy, line 76: This code may override or add duplicate entries when adding items in parallel.

We might want to be clear about the lack of guarantees here with regards to what specific `newValue` is returned, especially because it differs from the guarantees `ConcurrentDictionary` provides.

> The key is not in the dictionary. valueFactory generates a value. While valueFactory is generating the value, a different thread inserts a value for the key. After valueFactory executes and upon rechecking for the key, the key inserted by the other thread is found. | The value inserted by the other thread is returned.

https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-9.0#system-collections-concurrent-concurrentdictionary-2-getoradd(-0-system-func((-0-1)))

JamesNK:
@BrennanConroy Good feedback. I don't think the type would have broken and caused errors, but it could have gotten into a weird state (using the dictionary but still having some values in the array, duplicate values in the array). I've placed all updates in the lock. And added a comment about the value that's created.

@gfoidl The main goal here is to optimize idle memory usage. Making the array only as big as it needs to be, plus some array allocations while a channel is first used is an ok tradeoff.

BrennanConroy (approved):

Issue #2589, opened 2025-01-09
Potential deadlock in subchannel state update


Adarsh1994:
### What version of gRPC and what language are you using?
Version: 2.66.0
Language: C#

### What operating system (Linux, Windows,...) and version?
Linux

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.NET SDK 8.0

### What did you do?
We integrated gRPC .NET libraries to facilitate unary RPC calls between our services. While client-server communication generally works well with gRPC requests executing successfully, we occasionally encounter a potential deadlock. It looks like the issue arises when a subchannel state update occurs simultaneously after the completion of a request and a periodic DNS resolver attempting to update the same subchannel state.

### What did you expect to see?
No deadlocks during the gRPC communication

### What did you see instead?
Deadlock can be confirmed with the following stack traces of two stuck threads:

>Thread (0x8F25):  
``Grpc.Net.Client!Grpc.Net.Client.Balancer.Subchannel.UpdateAddresses(class System.Collections.Generic.IReadOnlyList`1<class Grpc.Net.Client.Balancer.BalancerAddress>)
  Grpc.Net.Client!Grpc.Net.Client.Balancer.PickFirstBalancer.UpdateChannelState(class Grpc.Net.Client.Balancer.ChannelState)
  Grpc.Net.Client!Grpc.Net.Client.Balancer.Internal.ChildHandlerLoadBalancer.UpdateChannelState(class Grpc.Net.Client.Balancer.ChannelState)
  Grpc.Net.Client!Grpc.Net.Client.Balancer.Internal.ConnectionManager.OnResolverResult(class Grpc.Net.Client.Balancer.ResolverResult)
  Grpc.Net.Client!Grpc.Net.Client.Balancer.DnsResolver+<ResolveAsync>d__11.MoveNext()
  System.Private.CoreLib!System.Threading.ExecutionContext.RunInternal(class System.Threading.ExecutionContext,class System.Threading.ContextCallback,class System.Object)
System.Private.CoreLib!System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1+AsyncStateMachineBox`1[System.Threading.Tasks.VoidTaskResult,Grpc.Net.Client.Balancer.DnsResolver+<ResolveAsync>d__11].MoveNext(class System.Threading.Thread)
[...]

PR #2601, opened 2025-02-24
Move updating connectivity state outside of subchannel lock

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2589

In connection manager and friends there are two important types of lock:
* A lock in the connection manager
* A lock for each subchannel

It's safe to take a subchannel lock inside the connection manager lock, but doing the opposite can lead to a deadlock.

There is one place locks were taken in the incorrect order which could cause a deadlock in rare circumstances: the channel is pick first, it has an idle subchannel, and a connection request occurs at the same time as the resolver updates.

Repoed in a unit test:

image

The fix is to move the connectivity state update outside of the subchannel lock.


src/Grpc.Net.Client/Balancer/Subchannel.cs
```diff
@@ -264,6 +266,11 @@ public void RequestConnection()
             }
         }
 
+        if (connectionRequested)
+        {
+            UpdateConnectivityState(ConnectivityState.Connecting, "Connection requested.");
```
BrennanConroy, line 271: Is this now going to cause races where someone else comes along and modifies the Subchannel state and we end up updating it to `Connecting` when it should stay as `TransientFailure` or `Ready`?

JamesNK: I believe this is fine. The logic in load balancing is designed that state can change at any point, and other parts of the system then react to it. All other places update the connectivity state outside the subchannel lock, this one is the outlier.

BrennanConroy (approved):

dannyheard7:
Would this be the cause of this reported issue? https://github.com/grpc/grpc-dotnet/issues/2612

Issue #2612, opened 2025-03-18
TaskCanceledException when reconnecting after disconnect


asleire:
### What version of gRPC and what language are you using?
Grpc.Net.Client 2.70.0, C#

### What operating system (Linux, Windows,...) and version?
Debian (Docker image mcr.microsoft.com/dotnet/aspnet:8.0)

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
Docker image mcr.microsoft.com/dotnet/sdk:8.0

### What did you do?
If possible, provide a recipe for reproducing the error. Try being specific and include code snippets if helpful.

### Application setup
Our GrpcClient is registered simply as such using `Grpc.Net.ClientFactory` version `2.66`:
```
 services.AddGrpcClient<Client>(o =>
        {
            o.Address = new Uri(<url>);
        });
```

### Issue description

Our application is having intermittent RPC errors with exception `System.Threading.Tasks.TaskCanceledException: A task was canceled.`. Investigating the logs it appears to happen when requests are picking a subchannel that has previously disconnected.

Full log example:
```
2025-03-17 17:01:24.684	Req X: Successfully picked subchannel id '"1-1"' with address "authorization-api.opa:8081". Transport status: ActiveStream
2025-03-17 17:01:26.777	 Subchannel id '"1-1"' state changed to Idle. Detail: '"Disconnected."'.
2025-03-17 17:01:26.777	 Channel state updated to Idle.
2025-03-17 17:01:26.777	 Channel picker updated.
2025-03-17 17:02:07.983	Req1: Picked subchannel id '"1-1"' doesn't have a current address.
2025-03-17 17:02:07.983	Req2: Subchannel id '"1-1"' state changed to Connecting. Detail: '"Connection requested."'.
2025-03-17 17:02:07.983	Req2: Channel state updated to Connecting.
2025-03-17 17:02:07.983	Req2: Channel picker updated.
2025-03-17 17:02:07.983	Req2: Picked subchannel id '"1-1"' doesn't have a current address.
2025-03-17 17:02:07.983	Req3: Picked queued.
2025-03-17 17:02:07.983	Req1: Picked queued.
2025-03-17 17:02:07.983	Req2: Picked queued.
2025-03-17 17:02:07.983	 Subchannel id '"1-1"' connecting to transport.
2025-03-17 17:02:07.984	 Subchannel id '"1-1"' canceling connect.
[...]

PR #2618, opened 2025-04-09
Fix race condition that caused inprogress connect to be canceled

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2612

A previous fix to stop deadlocks - https://github.com/grpc/grpc-dotnet/pull/2601 - caused transient request errors. If a subcannel's not connected, and multiple gRPC calls queue at the same time, it's possible for a connection to be rapidly started and canceled. Although the state of the subchannel is quickly resolved, the queued calls could get an error result.

The fix is to check if a connection request is in progress later in the connection pipeline. If there is already a request then interrupt the delay if it is waiting and exit.

### DIFF
--- a/src/Grpc.Net.Client/Balancer/Subchannel.cs
+++ b/src/Grpc.Net.Client/Balancer/Subchannel.cs
@@ -266,10 +266,10 @@ public sealed class Subchannel : IDisposable
             }
         }
 
-        if (connectionRequested)
-        {
-            UpdateConnectivityState(ConnectivityState.Connecting, "Connection requested.");
-        }
+        Debug.Assert(connectionRequested, "Ensure that only expected state made it to this point.");
+
+        SubchannelLog.StartingConnectionRequest(_logger, Id);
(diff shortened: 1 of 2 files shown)


src/Grpc.Net.Client/Balancer/Subchannel.cs
```diff
@@ -327,6 +327,15 @@ private async Task ConnectTransportAsync()
                 return;
             }
 
+            // There is already a connect in-progress on this transport.
+            // Don't cancel and start again as that causes queued requests waiting on the connect to fail.
+            if (_connectContext != null && !_connectContext.Disposed)
```
mgravell, line 332: I had to check the file to see if this had a race condition (i.e. can `_connectContext` *become* `null`; it can't) - it *isn't*, but may I suggest:

``` c#
if (_connectContext is { Disposed: false })
```

so that it is obviously not a race (i.e. single-read)?

PR #2586, opened 2024-12-16
Adds MapGrpcService overload for ServerServiceDefinition

aka-nse:
**This is a duplicate of #2535.**
**I made a serious operational error while tracking master, so I created a new pull request.**
**I apologize for the inconvenience.**

----

### Is your feature request related to a problem? Please describe.

This PR is implement for #2236.

`ServerServiceDefinition` is defined in *Grpc.Core.Api* library, and it was a prefer way to implement gRPC server with *Grpc.Core*.
In current grpc-dotnet, the server should be implemented on Asp.Net web application DI container via `GrpcEndpointRouteBuilderExtensions.MapGrpcService` generic method.
The APIs in this proposal bridges legacy way into grpc-dotnet to enable  to migrate easier.

### Describe the solution you'd like

2 overloads of `GrpcEndpointRouteBuilderExtensions.MapGrpcService` are added:

```csharp
public static class GrpcEndpointRouteBuilderExtensions
{
    public static GrpcServiceEndpointConventionBuilder MapGrpcService(
        this IEndpointRouteBuilder builder,
        ServerServiceDefinition serviceDefinition);

    public static GrpcServiceEndpointConventionBuilder MapGrpcService(
        this IEndpointRouteBuilder builder,
        Func<IServiceProvider, ServerServiceDefinition> getServiceDefinition);
}
```

This change enables to map gRPC services via `Grpc.Core.ServerServiceDefinition`,  therefore migration from Grpc.Core can be more easier.

#### Changes in this PR

- `Grpc.Core.Api`
  - makes `Grpc.Core.ServerServiceDefinition.BindService(ServiceBinderBase)` method as public[*](https://github.com/aka-nse/grpc-dotnet/blob/e3b571f77efadcbcf475634ed8dad4738ac72d4f/src/Grpc.Core.Api/ServerServiceDefinition.cs#L41)

- `Grpc.AspNetCore.Server`
  - adds new extension methods in `Microsoft.AspNetCore.Builder.GrpcEndpointRouteBuilderExtensions` class
    - `MapGrpcService(this IEndpointRouteBuilder, ServerServiceDefinition)`[*](https://github.com/aka-nse/grpc-dotnet/blob/e3b571f77efadcbcf475634ed8dad4738ac72d4f/src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs#L59-L68)
[...]

PR #2634, opened 2025-05-26
MapGrpcService with service definition refactor

JamesNK:
Overall changes in PR:

- Add example app `Definer`. Tempoarily references projects instead of nuget packages because nuget packages don't have APIs yet
- Add new `MapGrpcService` overloads to accept `ServerServiceDefinition` type. Binds methods added to the definition.
- Add support for not creating service instance when invoking endpoint that doesn't need it.
- Change internal `ServerServiceDefinition.BindService` method to public.

---

This PR is a refactor of https://github.com/grpc/grpc-dotnet/pull/2586. Adds same features but remove a lot of duplication.

- Reusing existing types which have a `TService` type parameter during binding. Possible by have a special internal service type that isn't created.
- Supports NativeAOT
- Update functional tests to build on top of this. Replaces functiona tests previous dynamic method source
- Rename example app to follow other app names


JamesNK:
cc @aka-nse

src/Grpc.Core.Api/ServerServiceDefinition.cs
```diff
@@ -38,7 +38,7 @@ internal ServerServiceDefinition(List<Action<ServiceBinderBase>> addMethodAction
     /// <summary>
     /// Forwards all the previously stored <c>AddMethod</c> calls to the service binder.
     /// </summary>
-    internal void BindService(ServiceBinderBase serviceBinder)
+    public void BindService(ServiceBinderBase serviceBinder)
```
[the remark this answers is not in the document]
JamesNK: Ok. I think it is fine to make this public. It seems useful, it doesn't leak other API, and eventually removes need for internalsvisibleto which I've found always causes issues eventually.

src/Grpc.AspNetCore.Server/GrpcEndpointRouteBuilderExtensions.cs
```diff
@@ -43,11 +45,44 @@ public static class GrpcEndpointRouteBuilderExtensions
...
+    /// </summary>
+    /// <param name="builder">The <see cref="IEndpointRouteBuilder"/> to add the route to.</param>
+    /// <param name="serviceDefinition">The instance of <see cref="ServerServiceDefinition"/>.</param>
+    /// <returns>A <see cref="GrpcServiceEndpointConventionBuilder"/> for endpoints associated with the service.</returns>
+    public static GrpcServiceEndpointConventionBuilder MapGrpcService(this IEndpointRouteBuilder builder, ServerServiceDefinition serviceDefinition)
+    {
+        ArgumentNullException.ThrowIfNull(builder, nameof(builder));
+        ArgumentNullException.ThrowIfNull(serviceDefinition, nameof(serviceDefinition));
+
+        var serviceRouteBuilder = builder.ServiceProvider.GetRequiredService<ServiceRouteBuilder<ServerServiceDefinitionMarker>>();
```
BrennanConroy, line 64: Any way to share method internals with the already existing one?

If not, should the new method also do `ValidateServicesRegistered(builder.ServiceProvider);`

src/Grpc.AspNetCore.Server/Model/Internal/ServiceDefinitionMethodProvider.cs
```diff
@@ -0,0 +1,91 @@
...
+            ArgumentNullThrowHelper.ThrowIfNull(handler, nameof(handler));
+            _context.AddClientStreamingMethod(method, CreateMetadata(), (service, request, context) => handler(request, context));
+        }
+
+        private static List<object> CreateMetadata()
+        {
+            // Accepting CORS preflight means gRPC will allow requests with OPTIONS + preflight headers.
+            // If CORS middleware hasn't been configured then the request will reach gRPC handler.
+            // gRPC will return 405 response and log that CORS has not been configured.
+            var metadata = new List<object>
```
BrennanConroy, line 84: nit: Can this list be a static so it's only allocated once?

PR #2720, opened 2026-04-17
Fix flaky DuplexStream_SendToUnimplementedMethod_ThrowError test

JamesNK:
Fix race condition in `DuplexStream_SendToUnimplementedMethod_ThrowError` where `GetStatus()` throws `InvalidOperationException` because `CallTask` hasn't completed yet.\n\nThe test was awaiting `ResponseHeadersAsync` before calling `GetStatus()`, but receiving response headers doesn't guarantee the internal call is fully complete. The fix drains the response stream (via `MoveNext` which throws `RpcException`) before checking status, ensuring the call is fully finished.


asheshvidyut (approved):
LGTM

Issue #2732, opened 2026-05-12
Client unhandled exception at Grpc.Net.Client.Internal.HttpContentClientStreamReader`2.MoveNextCore(CancellationToken cancellationToken)


osexpert:
### What version of gRPC and what language are you using?
Grpc.Net.Client 2.76.0

### What operating system (Linux, Windows,...) and version?
Windows 10

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
.net 10

### What did you do?
If possible, provide a recipe for reproducing the error. Try being specific and include code snippets if helpful.

Read things from server:
```
try
{
    while (await call.ResponseStream.MoveNext().ConfigureAwait(false))
    {
        // stuff
    }
}
catch (Exception e)
{
    // log
}
```

Some time later, I came back and I was met by unhandled excpetion.

### What did you expect to see?
No unhandled exception:-)

### What did you see instead?

 It is strange, as I am catching all. Also the unhandled exception stack trace is cut short and start out of nowhere.

```
.NETCoreApp,Version=v10.0
System.Net.Http.HttpProtocolException: The HTTP/2 server closed the connection. HTTP/2 error code 'STREAM_CLOSED' (0x5). (HttpProtocolError)
   at System.Net.Http.Http2Connection.ThrowRequestAborted(Exception innerException)
   at System.Net.Http.Http2Connection.Http2Stream.TryReadFromBuffer(Span`1 buffer, Boolean partOfSyncRead)
   at System.Net.Http.Http2Connection.Http2Stream.ReadDataAsync(Memory`1 buffer, HttpResponseMessage responseMessage, CancellationToken cancellationToken)
   at Grpc.Net.Client.Internal.StreamExtensions.ReadMessageAsync[TResponse](Stream responseStream, GrpcCall call, Func`2 deserializer, String grpcEncoding, Boolean singleMessage, CancellationToken cancellationToken)
   at Grpc.Net.Client.Internal.GrpcCall`2.ReadMessageAsync(Stream responseStream, String grpcEncoding, Boolean singleMessage, CancellationToken cancellationToken)
   at Grpc.Net.Client.Internal.HttpContentClientStreamReader`2.MoveNextCore(CancellationToken cancellationToken)
```

Looking at the code, I find it a bit supicious:

`E:\git\grpc-dotnet\src\Grpc.Net.Client\Internal\HttpContentClientStreamReader.cs`

Line 112:
```
                // Save move next task to track whether it is complete
[...]

PR #2738, opened 2026-05-26
_moveNextTask may be overwritten, after we release the lock, but befo…

osexpert:
_moveNextTask may be overwritten, after we release the lock, but before we return it

https://github.com/grpc/grpc-dotnet/issues/2732


JamesNK (approved):

JamesNK:
@osexpert Can you complete the CLA? That's required before merging

osexpert:
> @osexpert Can you complete the CLA? That's required before merging

I get "Merging is blocked", not sure if this is normal?

Issue #2739, opened 2026-06-03
Grpc.Net.* 2.80.0 DLLs ship with FileVersion 2.66.0.0 — regression from 2.76.0.0


virzak:
## Summary

The DLLs in the `2.80.0` packages (`Grpc.Net.Client`, `Grpc.Net.ClientFactory`, `Grpc.Net.Common`, `Grpc.Core.Api`) all ship with `FileVersion = 2.66.0.0`, which is **lower** than the same DLLs in `2.76.0` (`FileVersion = 2.76.0.0`). `ProductVersion` is correct in both releases — only the Win32 `FileVersion` resource regressed.

`2.80.0-pre1` is affected the same way; `2.76.0` is fine.

## Reproduction

Any platform with PowerShell can reproduce by inspecting the DLL inside the NuGet packages:

```powershell
$tmp = New-Item -ItemType Directory -Force -Path "$env:TEMP\grpc-check"
foreach ($pkg in 'grpc.net.client','grpc.net.clientfactory','grpc.net.common','grpc.core.api') {
  foreach ($v in '2.76.0','2.80.0') {
    $nupkg = Join-Path $tmp "$pkg-$v.nupkg"
    Invoke-WebRequest "https://api.nuget.org/v3-flatcontainer/$pkg/$v/$pkg.$v.nupkg" -OutFile $nupkg
    $zip = [System.IO.Compression.ZipFile]::OpenRead($nupkg)
    $entry = $zip.Entries | Where-Object { $_.FullName -match 'lib/net(8\.0|standard2\.[01])/[^/]+\.dll$' } | Select-Object -First 1
    $dll = Join-Path $tmp "$pkg-$v.dll"
    [System.IO.Compression.ZipFileExtensions]::ExtractToFile($entry, $dll, $true)
    $zip.Dispose()
    $info = [System.Diagnostics.FileVersionInfo]::GetVersionInfo($dll)
    '{0,-25} {1,-7} -> FileVersion={2,-10} ProductVersion={3}' -f $pkg, $v, $info.FileVersion, $info.ProductVersion
  }
}
```

Output:

```
grpc.core.api             2.76.0  -> FileVersion=2.76.0.0  ProductVersion=2.76.0+c52862d2…
grpc.core.api             2.80.0  -> FileVersion=2.66.0.0  ProductVersion=2.80.0+f22747c7…
[...]

PR #2747, opened 2026-07-22
Update version for v2.83.x

asheshvidyut:
# Description

Following [Step 3 for release doc](https://github.com/grpc/grpc-dotnet/blob/master/doc/release_process.md#releasing-a-new-version-of-grpc-dotnet-every-6-weeks) for `v2.83.0x`

Update version and remove `-dev` suffix with `-pre1` suffix

### DIFF
--- a/build/version.props
+++ b/build/version.props
@@ -8,7 +8,7 @@
     <GrpcDotnetAssemblyVersion>2.0.0.0</GrpcDotnetAssemblyVersion>
 
     <!-- file version of all grpc-dotnet -->
-    <GrpcDotnetAssemblyFileVersion>2.66.0.0</GrpcDotnetAssemblyFileVersion>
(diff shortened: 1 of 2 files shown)


build/version.props
JamesNK: This needs to be updated. See https://github.com/grpc/grpc-dotnet/issues/2739

Update the release notes to mention that this needs to be updated.

src/Grpc.Core.Api/VersionInfo.cs
JamesNK: And update this.

JamesNK (changes requested):

build/version.props
JamesNK: Update release_process.md on master branch.

Edit: https://github.com/grpc/grpc-dotnet/pull/2748

JamesNK (approved):

